PR #216, opened 2013-11-05
--depth for register

johannesgerer:
As discussed on IRC. 

Right now the depth is hard-coded in line 475. This needs to be taken from the command-line argument "--depth" (or 0, if "--collapse" is used).

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -7410,7 +7410,7 @@ The market value of a posting, or an account without its children.
 The net gain (market value minus cost basis), for a posting or an
 account without its children.  It is the same as @samp{v-b}.
 
-@item l
+@item depth
 The depth (``level'') of an account.  If an account has one parent,
 it's depth is one.
 
--- a/src/filters.cc
+++ b/src/filters.cc
@@ -473,6 +473,10 @@ void collapse_posts::report_subtotal()
 value_t& collapse_posts::find_totals(account_t* account)
 {
   unsigned short depth=3;
+
+	if(depth==0)
+		return totals[_("<Total>")];
+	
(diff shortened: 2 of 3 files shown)


jwiegley:
@johannesgerer Merged as is, thanks for your patience!

jwiegley:
@johannesgerer What does the "3" here mean:

``` c++
value_t& collapse_posts::find_totals(account_t* account)
{
  unsigned short depth=3;

	if(depth==0)
		return totals[_("<Total>")];

  if(account->depth==depth)
    return totals[account->fullname()];
```

jwiegley:
I'm reverting this for now, as it also has odd behavior: `--depth=1` only show postings where the account has amount in it *at depth 1*.  I.e., a posting into `Expenses:Food` doesn't show up at `depth=1`.

johannesgerer:
I am sorry for the confusion. I think my two previous posts in this thread should answer your questions.

jwiegley:
Ok, I think I'm just going to reject then, as I'm not sure it provides an intuitive enough experience to be worth the change to the code.

jwiegley:
For example, I would want to see unit tests demonstrating the capabilities, and an addition to the manual explaining what this feature does for the user.  My expectation was that `--depth=2` would aggregate subtotals in the level 2 accounts.  But maybe you're saying the implementation was just incomplete?  If so, then we can revisit.

johannesgerer:
Yes, "Right now the depth is hard-coded" and "I was just sharing my progress". The code only sets the stage for the actual functionality which is still missing.

jwiegley:
@johannesgerer My mistake.

PR #305, opened 2014-06-27
Add regression test file for bug #1057

thdox:


jwiegley:
This PR changes 114 files, yet the subject says it just adds a regression test.  Please resubmit just that test if you still want it to be included.

PR #312, opened 2014-07-30
Feature precision

tripun:
implemented precision commodity sub directive.

Use case: 
Monetary test cases only .

Performance: 
 It does not affect ledger performance significantly .Sorting is only done just when required and a flag can be added so that sorting is not done for those who do not require this feature. 

Comment:
Sorting by commodity is already being done for functions like printing.  fn: :map_sorted_amounts

Tests:
All tests which do not use this feature passed at the time of writing this feature. Tests using this feature passed but have been written to check specific requirements. More tests should be added with combination of options using this feature. 

Challenge faced: 
Account with single amount does not use balance_t object, The balance is maintained as amount_t object. Therefore, rounding implemented for balance valuation does not work and so rounding has been done after the amount is valued. 

Future work: 
Extend this feature to non-monetary use cases.

### DIFF
--- a/src/amount.cc
+++ b/src/amount.cc
@@ -447,6 +447,9 @@ amount_t& amount_t::operator+=(const amount_t& amt)
 
   mpq_add(MP(quantity), MP(quantity), MP(amt.quantity));
 
+ if (has_commodity() && commodity().has_flags(COMMODITY_SET_CUSTOM_PRECISION))
+    in_place_roundto(commodity().custom_precision());
+
   if (has_commodity() == amt.has_commodity())
     if (quantity->prec < amt.quantity->prec)
       quantity->prec = amt.quantity->prec;
--- a/src/amount.cc
+++ b/src/amount.cc
@@ -477,9 +477,6 @@ amount_t& amount_t::operator-=(const amount_t& amt)
 
   mpq_sub(MP(quantity), MP(quantity), MP(amt.quantity));
 
-  if (has_commodity() && commodity().has_flags(COMMODITY_SET_CUSTOM_PRECISION))
-     in_place_roundto(commodity().custom_precision());
-
   if (has_commodity() == amt.has_commodity())
     if (quantity->prec < amt.quantity->prec)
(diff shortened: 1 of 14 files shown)


jwiegley:
@tripun Can you define "significantly"?  Performance is a crucial feature of C++Ledger.

thdox:
I step into this discussion, because I find c++ ledger slow, and I have to set up workarounds to get it less slow (described in thread https://groups.google.com/d/msg/ledger-cli/2oYoEimj-r0/z0fO5hBIBg0J)

I have tried this pull request #312 on my system, and tested the performance of one command that I use a lot in my reporting (this is the same command that I talk about in the thread above). Here is result:

$ time ledger --exchange=€ -J register > /dev/null
real	3m28.857s
user	3m26.276s
sys	0m1.396s

Without this PR #312 , using next branch, the result is:

$ time ledger --exchange=€ -J register > /dev/null
real	1m58.277s
user	1m57.020s
sys	0m1.080s

Looking at the code of the commits, I have not found the "flag can be added so that sorting is not done for those who do not require this feature" that @tripun is talking about.

From my point of view the slowdown is big, and I would not accept this PR.

tripun:
Hi thdox, 
I shall check and add a fix. If precision option is used , then only it should be used. Please allow me time. Thanks for benchmarking :)

jwiegley:
@thdox Can you build with profiling enabled, and then send me the profiling report for your typical use case?  I'm guessing you either have a lot of automated transactions (which multiply the time required by the number of such transactions), or you have a huge number pricing entries in the pricing graph.  Seeing your profiling output will tell me how we could optimize your use case, or improve Ledger.

thdox:
@jwiegley, I emailed you privately the gmon.out file.

afh:
Any updates on this PR?

jwiegley:
No updates as of yet. I'm still waiting to learn what affect this proposed change will actually have, which hasn't been answered yet.

tbm:
Rounding is still an issue but this PR isn't ready and isn't getting updating, so closing.

PR #323, opened 2014-09-14
Manual install instruction

kain88-de:
The install instruction in the manual still used an old `configure` setup. I copied and adjusted the instructions from `README.md`. It would be nice if a native speaker could look over it. Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,10 +172,9 @@ footwork for you:
     # $HOME/local and build with 2 processes in parallel
     ./acprep update --boost-suffix=-mt --prefix=$HOME/local -j2
 
-Please read the contents of `config.log` if the configure step fails.  Also,
-see the `help` subcommand to `acprep`, which explains some of its many
-options.  It's pretty much the only command I run for configuring, building
-and testing Ledger.
+See the `help` subcommand to `acprep`, which explains some of its many options.
+It's pretty much the only command I run for configuring, building and testing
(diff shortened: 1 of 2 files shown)


kain88-de:
I fixed the broken test.

afh:
Hi @kain88-de,
I'd prefer if you could get rid of the following two commits by using `git rebase -i`:
* f42134d787fcda084aaeaa4c1f9df96ef415aa10
* 9afe61174d289df2ff8cd0804e42849f90c2d6fc

purcell:
> I'd prefer if you could get rid of the following two commits by using git rebase -i:

ie. squash them into the other commits, not remove them completely. :-)

afh:
@purcell As fas as I can see commit f42134d removes whitespace which is added again in 9afe611, so I'd prefer both commits to be removed form this pull request, since there is no value in them, or am I missing something?

purcell:
Oh, I see. If that's the case, then I completely agree.

kain88-de:
@afh done

afh:
@kain88-de Thanks :+1:  I have two more little things I'd like before I approve this pull-request:

1. Please change this pull-request so that it is against the `next` branch.
2. and add `[ci skip]` to each commit message, since the tests are irrelevant for the continuous integration.

kain88-de:
Done see #378

PR #367, opened 2015-01-10
[emacs] Fontification should work independentily on where the point currently is

racktear:
This is a subtle bug that happens only when org-src-fontify-natively is
enabled:

```emacs-lisp
(setq org-src-fontify-natively t)
```

If then you create a Babel block with ledger text, it won't be
fontified.

This happens because the 'ledger-fontify-buffer-part' starts
fontification from the current 'point', which in case of
org-src-font-lock-fontify-block happens to _not_ point to the beginning
of buffer. Instead it points to the original org-mode buffer.
This doesn't happen when one opens regular files, because the
'point' is almost always at the beginning of the file.

To reproduce the bug, you can do the following:
```emacs-lisp
(defun fontify-test()
  (let ((lang-mode 'ledger-mode) pos next)

  (with-current-buffer
      (get-buffer-create
       "ledger-fontification-bug")
    (unless (eq major-mode lang-mode) (funcall lang-mode))
    (erase-buffer)
    (insert "2015/01/09 asdf\n  foo\n  bar\n" " ")
    (font-lock-fontify-buffer)
    (switch-to-buffer "ledger-fontification-bug")
    )))
```
Call 'fontify-test' and see that the buffer is not fontified.

### DIFF
--- a/lisp/ledger-fontify.el
+++ b/lisp/ledger-fontify.el
@@ -42,6 +42,7 @@
 	(save-excursion
 		(unless beg (setq beg (point-min)))
 		(unless end (setq end (point-max)))
+		(goto-char beg)
 		(beginning-of-line)
 		(while (< (point) end)
 			(cond ((or (looking-at ledger-xact-start-regex)


afh:
Hi @racktear,
thanks for you pull-request please change the it to the `next` branch, we're moving to a [git-flow](http://nvie.com/posts/a-successful-git-branching-model/) based development model.

Additionally please add `[ci skip]` to the end of the commit message, so that we skip [continuous integration](https://travis-ci.org/ledger/ledger) for changes not relevant to it.

racktear:
@afh I'll open another pull request.

purcell:
> thanks for your pull-request please change the it to the next branch, we're moving to a git-flow based development model.

I've added a CONTRIBUTING.md so that this info is more obvious to would-be contributors -- github shows it when submitting pull requests. But if we find a lot of PRs come in against `master` anyway, might it not be better to make `next` the default branch?

afh:
Well, we want people using Ledger to use a current stable version from `master`, but we also want contributors to contribute to `next`. So the question is: Is the default github branch meant for users or developers?

purcell:
> So the question is: Is the default github branch meant for users or developers?

I'd guess the latter. Users are presumably going to want either the "Releases" section or the latest dev code. (That reminds me -- the [homebrew](https://github.com/Homebrew/homebrew) guys probably need to be told that the branch for `--head` builds is no longer `master`. I'll submit a PR for that.)

purcell:
(Note that if we changed the default branch, it would also mean the `homebrew` recipe would not need to be changed, so I'll hold off on submitting a change for now.)

afh:
I don't see the need to change the Homebrew formula, since @jwiegley wants frequent 'stable' pre-releases merged into `master` from next `next` (see [post to the mailing list](https://groups.google.com/d/msg/ledger-cli/s7cUVPobynI/hGF_u7w2zxwJ)).

purcell:
> I don't see the need to change the Homebrew formula, since @jwiegley wants frequent 'stable' pre-releases merged into master from next next (see post to the mailing list).

There's a separate "devel" build in homebrew for that, though. `HEAD` is explicitly supposed to be the development branch. But of course it won't make a huge difference to anyone.

afh:
Ah yes, I forgot about devel... I do agree it probably won't a huge difference to anyone. Let's see how it goes...

jwiegley:
If it will make your lives easier for us to set `next` as the default branch, then I'm fine with it.  I wanted `master` only because that should always pass tests, and it's what a drive-by GitHub user (or a packager) will track, but I don't mean for it to increase the inertia of contributing.

purcell:
I personally think that the default git branch should be `next`, so that that is what contributors will automatically end up committing onto. I can't judge what effect that will have on packagers or non-contributing github users.

jwiegley:
Ok, let's make it `next` then.  Packagers can learn easier than trying the educate the masses.

purcell:
:+1:

llcc:
When I updated the ledger-mode some time ago, the block source in org-mode can not show syntax any more. Can you reproduce this problem?

PR #369, opened 2015-01-12
Replace utfcpp submodule with partial subtree

afh:
Currently the files residing below `lib/utfcpp` are included via
a git submodule from https://github.com/ledger/utfcpp.
That repository has diverged from the upstream [utfcpp repository](http://sourceforge.net/p/utfcpp/code/HEAD/tree/)
and I'd like to use the last stable version of utfcpp (2.3.4) with Ledger.

Since using a complete mirror, e.g. https://github.com/svn2github/utfcpp
would include large test data (~20MB) and increase Ledger's release archive
size significantly I think a 'partial' git subtree is a good choice.

Using [git-subtree](http://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/) to manage the utfcpp files instead of a git submodule
has the advantage of those files residing in the source tree (as opposed
to a 'reference' to them with submodule) and they are therefore
available in archives, making it easier to build Ledger from a
downloaded source archive.  The package maintainers will thank us for it!

And with the [`FindUtfcpp.cmake` changes](https://github.com/ledger/ledger/commit/bedfe746f2e231f3c9c22394ae1fc1188ca40787) standard include locations are searched first and used should a distribution provide a `libutfcpp-dev` package.

The effort to keep utfcpp up-to-date is negligible since the project
has seen very few updates in the past (2.3.4 was released Februrary 17th, 2013).

Additional information can be found in my [post](https://groups.google.com/forum/#!topic/ledger-cli/-e6H-pZbDlE) to the list.

The following changes were made:
* Remove utfcpp submodule
* Squashed 'lib/utfcpp/' content from commit 63cd498
  git-subtree-dir: lib/utfcpp
  git-subtree-split: 63cd4984464e875546b4a91918be7355357a3a68
* Merge commit '82003f30ba8ba436cf7cef380155f50ed4242b37' as 'lib/utfcpp'
* Remove needless utfcpp subtree files
* Change include path to utfcpp
* Add FindUtfcpp.cmake


jwiegley:
+1

yaccz:
utfcpp is a dependency as any else. It has no business inside the ledger repository as git module, subtree or otherwise.

jwiegley:
It is a fairly tiny dependency which is not packaged on a lot of platforms, so expecting users to go out and find it and build it themselves is just too annoying to contemplate.  The license allows it to be included in the ledger sources, so I have no problem with that.

yaccz:
> not packaged on a lot of platforms

jeez I wonder why

> expecting users to go out and find it

s/users/packagers
s/find it/package unpackaged dependency
How unreasonable indeed.

>  and build it

That package does not need building.

> The license allows it to be included in the ledger sources, so I have no problem with that.

It's legally possible, therefore it must be the right thing to do, right?

jwiegley:
@yaccz Given your tone, I have no interest in discussing this further.  Thanks for stopping by.

yaccz:
Oh, the tone argument. Enjoy your not giving a fuck and have a nice day.

jwiegley:
The tone argument was not addressed to your points, but the fact that as a collaborator in an open source project, I choose to not to work with individuals who are openly hostile, as you seem to be.  The truth of your words is getting lost, due the manner in which you've expressed them.  You'd find that even a little bit of kindness would be rewarded with a hearing ear, almost always.

afh:
@yaccz I'm curious: What problems having the sources of utfcpp in `lib/` cause you as a packager or user?

With the proposed addition of the [`cmake/FindUtfcpp.cmake`](https://github.com/afh/ledger/blob/pull/utfcpp_subtree/cmake/FindUtfcpp.cmake) module a system provided utfcpp is used over the one in the source tree.

yaccz:
@afh As a user I expect all my programs to be managed through the distribution package manager. If you bundle the dependency and tell me to just use "./acprep make" or whatever which will install other dependency I might not not notice. I might decide I don't want to use your project because of that sneaky dependency. This may result into lazy packagers to just package the dependency with the ledger code. Now as I user I have no idea there another dependency sneaked in which makes my life harder if is a security issue in that dep.

As a packager I have to deal with your weird build system to figure out how to unbundle the stupid stuff. This may be attributed to my inexperience with cmake but still, your readme file says to build with acprep which seems the authoritative right way of doing things. It's not. The right thing to do is "Build just standard cmake way but hey, we are using this weird library nobody has bothered to package and neither did we. Instead of doing the technically right thing which would involve maybe some talking to people, we wrote this stupid wrapper to just pull this thing from the internet which we also patched but if you look closer it's just few cosmetical things. Just to make really sure the version we pull eventually gets outdated and nobody will even notice or bother to update."

So bundling is bad practice, mmmmkay. You might argue that utfcpp is just this little library and blah blah. Bad practice. It solves the problem at the wrong place in the wrong way with more code (more code is bad). That's dilettantism in my book.

jwiegley:
Just so you know, I'd prefer that everyone used just `cmake` also.  `acprep` was only ever written as an aid to my development; I did not intend for it to become as widely used as it is.  If we can take steps to avoid its use and promote proper deployment channels, as @yaccz recommends, then I'm all for it.

afh:
@yaccz: Thanks for you input.

`acprep make` did not install anything, it merely fetched a needed submodule.

`acprep dependencies` is provided for convenienence and it installs the
needed dependencies via the package manager for the operating system, but this
should be removed in my opinion.

As a packager building a binary Ledger package there is no need to unbundle
anything anymore since `cmake/FindUtfcpp.cmake` will find a system provided
utfcpp and the files utfcpp provides are only relevant for building the package
not the the package itself.

This is different if you are creating a ledger-dev package (which to my knowledge
does not exist (yet?) on any distribution), then that package should have a
dependecy on libutfcpp-dev and ignore files in lib/utfcpp.

If distributions provide the ledger headers, which need utfcpp, with the
regular binary ledger package, then a ledger-dev package should be introduced on that platform
from my point of view.

I think deprecating acprep is a good idea, how about `acprep` just prints out
the commands the user needs to run and suggests not to use `acprep` in the future?

Since `cmake` can be a bit of a pain to setup up some platform specific Makefiles
such as `Makefile.Darwin` or `Makefile.Homebrew` might be helpful.

What are you thoughts on this?

jwiegley:
>>>>> Alexis Hildebrandt <notifications@github.com> writes:

> I think deprecating acprep is a good idea, how about acprep just prints out
> the commands the user needs to run and suggests not to use acprep in the
> future?

Although I'm all for deprecating, I don't want to annoy everyone just because
of one person's view of how the world should be.  acprep could indeed print
out the commands you'd run instead of acprep, but I'd still like the script to
work for those of us who use it.

John

afh:
Fair point.

yaccz:
I think acprep takes a bad approach to helping users install the software since it 1. encourages running random stuff from the internet as root and 2. it duplicates the effort distribution package maintainers. I also saw some hardcoded things like if /Users/johnw/.. exists, do something else. This belongs to your development environment but not into the upstream project.

If you still want to do the extra effort I'd suggest at least making it clear in the readme the project should be buildable just by standard cmake and acprep is helper for less experienced users. I'd even suggest moving acprep into separate project like ledger-installer.git or something.

afh:
While there are many good points mentioned in this issue, the comments aren't about the initial pull-request anymore. I suggest creating issues in [Bugzilla](http://bugs.ledger-cli.org) or even better propose pull-requests to review for the mentioned issues, so this pull-request can be merged into `next` to move development forward `cmake/FindUtfcpp.cmake` IMO is a very useful addition to the project regardless how (submodule or subtree) the utfcpp files are stored in the repo for convenience.

jwiegley:
@afh Agreed.

PR #372, opened 2015-01-18
[emacs] fix copy

marcowahl:
ledger-xact.el:ledger-copy-transaction-at-point (C-c C-k) led to a confusing change in the ledger file for the case when the target date was smaller than last date in the ledger.


marcowahl:
Urgh, sorry, I should have read the Contributing.md before requesting the pull.

This is a [ci skip] pull request.

PR #374, opened 2015-01-19
Add those single letter directives to index.

thdox:


doc/ledger3.texi
```diff
@@ -2554,12 +2555,15 @@ alone, for backwards compatibility with older Ledger versions.
 @table @code
 
 @item A
+@findex A
 See @code{bucket}.
```
afh: Is it possible to use a `@pxref` to bucket here?

doc/ledger3.texi
```diff
@@ -2554,12 +2555,15 @@ alone, for backwards compatibility with older Ledger versions.
...
 
 @item Y
+@findex Y
 See @code{year}.
```
afh: Is it possible to use a `@pxref` to year here?

doc/ledger3.texi
thdox: Good, had to learn a new thing !  :-) I pushed a new pull request.

afh: Merged it. Thanks! :)

PR #379, opened 2015-01-20
[fix] Make trim function trim trailing whitespace

afh:
std::isspace(*e) returns false for the end of c-string null-byte.

Bugzilla: 1106


jwiegley:
I wish there was an efficiency way of linking from the Bugzilla over here to the pull request.

afh:
It can be done the other way around via the [GitHub Bugzilla Service Hook](https://github.com/ledger/ledger/settings/hooks/new?service=bugzilla).

PR #373, opened 2015-01-19
Make documented value expressions testable

afh:
I've changed the formatting of the value expressions using `@defun`, `@defvar`, etc. in the texinfo and to `.It Fn`, `.It Sy` in the manpage, which to me seem the more appropriate texinfo commands for documenting these.

Additionally I've added a `@smallexample`s to some of the value expressions in the texinfo manual, so they are tested and validated by `test/DocTests.py` now! :smile:

The undocumented value expression functions are also reported by `test/CheckManpage.py` and `test/CheckTexinfo.py`

What do you think?

PR #382, opened 2015-01-22
Add documentation for --no-pager option.

thdox:
Minor updates to --pager documentation.

I was surprised that this was not documented, I use this option very frequently.

@afh From my point of view this makes no sense to test this as a baseline options. Is there any trick in your python script to authorize to not test --no-pager as a baseline option?

[ci skip]

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -5620,6 +5620,9 @@ Report the last @var{INT} postings.
 @item --pager @var{FILE}
 Direct output to @var{FILE} pager program.
 
+@item --no-pager
+Direct output to stdout, avoiding pager program.
+
 @item --average
 @itemx -A
 Report the average posting value.
@@ -6430,6 +6433,9 @@ Aliases are completely ignored.
 @item --no-color
 Suppress any color TTY output.
 
+@item --no-pager
+Direct output to stdout, avoiding pager program.
+
 @item --no-rounding
 Don't output @samp{<Rounding>} postings.  Note that this will cause the
 running total to often not add up!  Its main use is for
@@ -6453,7 +6459,7 @@ been executed, such as periodic gathering.
 Redirect the output of ledger to the file defined in @file{FILE}.
 
 @item --pager @var{FILE}
-Specify the pager program to use.
+Direct output to @var{FILE} pager program.
 
 @item --payee @var{VEXPR}
 Sets a value expression for formatting the payee. In the
@@ -6960,9 +6966,13 @@ meaning of the flag (instead of the first five transactions being
 printed, for example, it would print all but the first five).
 
 @item --pager @var{FILE}
-Tell Ledger to pass its output to the given pager program; very useful
-when the output is especially long.  This behavior can be made the
-default by setting the @env{LEDGER_PAGER} environment variable.
+Tell Ledger to pass its output to the given @var{FILE} pager program;
+very useful when the output is especially long.  This behavior can be
+made the default by setting the @env{LEDGER_PAGER} environment variable.
+
+@item --no-pager
+Tell Ledger to @emph{not} pass its output to a pager program; useful
+when a pager is set by default.
 
 @item --average
 @itemx -A
--- a/doc/ledger.1
+++ b/doc/ledger.1
@@ -622,8 +622,8 @@ Shorthand for
 Aliases are completely ignored.
 .It Fl \-no-color
 Suppress any color TTY output.
-.\".It Fl \-no-pager
-Suppress any color TTY output.
+.It Fl \-no-pager
+Disables the pager on TTY output.
 .It Fl \-no-rounding
 Don't output
 .Nm <Rounding>


doc/ledger3.texi
```diff
@@ -6453,7 +6459,7 @@ been executed, such as periodic gathering.
 Redirect the output of ledger to the file defined in @file{FILE}.
 
 @item --pager @var{FILE}
-Specify the pager program to use.
+Direct output to @var{FILE} pager program.
```
afh: How about using `PROGRAM` or something similar instead of `FILE` to better describe the argument to the `--pager` option

doc/ledger3.texi
```diff
@@ -6960,9 +6966,13 @@ meaning of the flag (instead of the first five transactions being
...
-Tell Ledger to pass its output to the given pager program; very useful
-when the output is especially long.  This behavior can be made the
-default by setting the @env{LEDGER_PAGER} environment variable.
+Tell Ledger to pass its output to the given @var{FILE} pager program;
+very useful when the output is especially long.  This behavior can be
+made the default by setting the @env{LEDGER_PAGER} environment variable.
+
+@item --no-pager
+Tell Ledger to @emph{not} pass its output to a pager program; useful
+when a pager is set by default.
```
afh: Ou to curiosity I just tested `ledger --no-pager --pager less -f test/input/sample.dat reg` (displays report with `less`) and `ledger --pager less --no-pager -f test/input/sample.dat reg` (displays report directly) it might be worthwhile to document that the "the last given option wins".

doc/ledger3.texi
```diff
@@ -6430,6 +6433,9 @@ Aliases are completely ignored.
 @item --no-color
 Suppress any color TTY output.
 
+@item --no-pager
+Direct output to stdout, avoiding pager program.
+
```
afh: In the long run I'd like to get rid of the duplicate entries this makes maintenance cumbersome. Any idea on how to restructure the manual to avoid duplicates or using `[p][x]ref` in a meaningful way?

afh:
@thdox Thanks for your contribution. Yay, progress! :smile: 

Yes, there is a trick: please pull the latest change I pushed onto `next` and add `no-pager` to the `self.untested_options` array in `test/CheckBaselineTests.py:21`.

afh:
I took the liberty of merging this pull-request and add the reference in order to merge  #373.

doc/ledger3.texi
thdox: My first commit was with PROGRAM as you suggested, but I reset it, as a consistency matter. FILE is indeed used consistently in the documentation to tell that this must me a Unix file whatever it is a program, a file, or anything else.

doc/ledger3.texi
thdox: Good point! That is correct that this kind of precedence is not documented. What I propose is to add a section about precedence. I will also put in this section notes about precedence between options, ENV variables, and init file.

doc/ledger3.texi
thdox: Craig before me, and me sometimes ago, we add a thought on how to restructure the documentation. I have to admit, that me, discovering the size of the task and the difficulty, somewhat procrastinate to do it.
Anyway, I have some thoughts about restructuration. Yes, this is more than cumbersome to document in several places. But you can not list all options without a introduction. The balance is hard to maintain.
About the links, what you propose is HTML URL minded. What is currently implemented is based on another point of view : any item in documentation is added to index, and you refer to index to jump to relevant section. More than a book. At the moment, I have no prefered option. I am just used to use the second one. And this make the PDF more readable. But I agree that HTML linking is helpfull too. It is just ugly in the PDF outputed from texinfo files.

thdox:
@afh , you may have realized that I have inconsistent availability to dedicate to ledger. Anyway, I will try to retro-document the option where documentation is missing.

doc/ledger3.texi
afh: Very well.

doc/ledger3.texi
afh: Great thanks I look forward to the pull request :smiley:

doc/ledger3.texi
afh: Thanks for pointing out my limited view on the documentation, I really hadn't thought about the PDF output, and will change my mindset accordingly.

What would be possible is to use `@ifhtml . . . @end ifhtml` for example to link to the balance assertions in the documentation of the `--permissive` option when producing HTML output in addition your `@findex` change.

Another thing I just noticed while going through the [Texinfo reference card](http://svn.savannah.gnu.org/viewvc/*checkout*/trunk/doc/refcard/txirefcard-a4.pdf?root=texinfo&content-type=text%2Fplain): there are other index commands, e.g. `@vindex` for variables. Unfortunately I did not find one for options, but I think it would be nice to separate some of the index entries. What are your thoughts on this?

afh:
@thdox inconsistent availability is so much better than no availability at all. I'm quite thankful for your reviews, feedback, and valuable contributions.

I have some time at my hands at the moment and am really motivated in moving ledger forward, so please understand if I impatiently merge things to move ahead, and I appreciate your comments
on the commits. Keep it up (as your availability permits :)!

PR #384, opened 2015-01-28
[doc] Improve documentation about "balance assertions" and --permissive

thdox:
Remove back-link from option to section "balance assertions".
Add forward-link from section to option.
Add option to @findex
[ci skip]


afh:
Thanks for bringing this to my attention I wasn't aware if it.

PR #387, opened 2015-02-08
prepare-commit-msg fix

thdox:
prepare-commit-msg was failing on my computer, complaining about unexpected '('. This is fixed with commit #62977ea. Other commits are from advise by http://www.shellcheck.net/

[ci skip]

### DIFF
--- a/tools/prepare-commit-msg
+++ b/tools/prepare-commit-msg
@@ -5,7 +5,7 @@
 
 # Add [ci skip] to the commit message unless there are changes to files
 # that are relevant for testing such as src/*, test/*, ledger3.texi, ...
-function add_ci_skip()
+add_ci_skip ()
 {
   pattern="$1"; shift
   source="$1"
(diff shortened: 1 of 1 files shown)


tools/prepare-commit-msg
```diff
@@ -5,17 +5,17 @@
...
   source="$1"
 
   # Don't add [ci skip] if it's already in the commit message source
-  grep '\[ci skip\]' "$source" 2>&1 >/dev/null
+  grep '\[ci skip\]' "$source" >/dev/null 2>&1
   [ $? -eq 0 ] && return
 
   if [ $(git diff --cached --name-only | grep --count "$pattern") -eq 0 ]; then
-    tempfile=$(mktemp $0.XXXXXX)
+    tempfile=$(mktemp "$0".XXXXXX)
```
afh: I'd prefer `"${0}.XXXXXX"` here.

What's your opinion on this?

thdox:
@afh You're right! :-) Here is additional commit.

afh:
@thdox Thanks for improving.

PR #388, opened 2015-02-08
[emacs] Untabify emacs lisp

thdox:
Changes to emacs lisp code are hard to follow and to diff because of tabs. I cleaned the elisp files, but also I added emacs directory local variable to fix more permanently this issue.

[ci skip]

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,30 +1,36 @@
-((nil . ((tab-width . 2)
-         (sentence-end-double-space . t)
-         (bug-reference-url-format
-          . "http://bugs.ledger-cli.org/show_bug.cgi?id=%s")))
- (c-mode . ((c-file-style . "ledger")
-            (c-style-alist
-             ("ledger"
-              (indent-tabs-mode)
-              (c-basic-offset . 2)
-              (c-comment-only-line-offset 0 . 0)
-              (c-hanging-braces-alist
-               (substatement-open before after)
-               (arglist-cont-nonempty))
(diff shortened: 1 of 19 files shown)


enderw88:
Please add [emacs] to the commit message title to clarify that this is lisp code only.

thdox:
Of course I can do it and will do. AFAICT, this requirement is not in CONTRIBUTING.md.

afh:
It might be good to advise contributors to prefix the commit messages with `[component]` where `component` is preferably one of the [bugzilla components](http://bugs.ledger-cli.org/describecomponents.cgi).

thdox:
Back from business trip, so I tackle this tonight. I propose new pull request.

My point of view on prefixing:
- as I consider myself a member of the team, I will apply any best practice agreed by majority, and moreover when it is recognized as simplifying or adding clarity.
- but let's take the point of view of a newcomer to ledger. He proposes his first commit, and he starts contributing with something that has some added value. My point of view is that we should not discourages this kind of initiative from a newcomer. And imagine that this newcomer has carefully read CONTRIBUTING.md. His pull request is consistent with the rules. And then we stop him because of some bureaucracy rule that is written nowhere, and we are stopping us to benefit from his commit. And we should not stop him even if he did not follow this kind of rule. It is MY point of view, as "fight bureaucracy" is one of my motto. YMMV...

jwiegley:
I'm ok with the prefixing not happening, because I bet you I'd forget to do it myself.

PR #390, opened 2015-02-12
Improve documentation about --rich-data.

thdox:


doc/ledger.1
```diff
@@ -845,7 +851,10 @@ but show both sides of each transaction.
 .\".It Fl \-revalued
 .\".It Fl \-revalued-only
 .\".It Fl \-revalued-total Ar EXPR
-.\".It Fl \-rich-data
+.It Fl \-rich-data
+When generating ledger transaction from CSV file using the
+.Ic convert
+command, add CSV, Imported, and UUID meta-data.
```
afh: Please add an explanation for each metadata, e.g.
* CSV: the original csv 
* Imported: Date when the cvs was imported into ledger
* UUID: a uuid (I think based on  a checksum of the original csv) uniquely identifying this transaction

doc/ledger3.texi
```diff
@@ -6604,10 +6604,8 @@ show both @emph{sides} of each transaction.
 
 @item --rich-data
 @itemx --detail
-@c When generating ledger transaction from csv using the convert command
-@c add CSV, Imported, and UUID meta-data.
-@c see test/baeline/opt-rich-data.test
-@value{FIXME:UNDOCUMENTED}
+When generating ledger transaction from CSV file using the
+@command{convert} command, add CSV, Imported, and UUID meta-data.
```
afh: Please add an explanation for each metadata (see above comment).

doc/ledger.1
thdox: I have added an example for --rich-data in the manual, and also I also explained the associated tags. See additional commit.

But I will not add this to man page, that serve a different philosophy compared to texinfo file (see end of section Output_formats of http://en.wikipedia.org/wiki/Texinfo#Output_formats, and also http://www.gnu.org/software/texinfo/manual/texinfo/texinfo.html#Adding-Output-Formats). The man page serve as a reference card, and actually the current man page of ledger does not contain any example and explanation of each option is very short. I do not judge good or bad, but as a matter of fact, Ledger is more of less following same philosophy as Debian, with close to empty man page, referring to info if more is needed.

doc/ledger3.texi
thdox: Done! ;-)

doc/ledger.1
afh: Valid point, I agree.

doc/ledger3.texi
afh: Thanks

doc/ledger3.texi
```diff
@@ -4616,13 +4616,42 @@ from the file above.
...
-a UUID field. If an entry with the same UUID tag is already included in
-the normal ledger file (specified via @option{--file @var{FILE} (-f)} or
-via the environment variable @env{LEDGER_FILE}) this entry will not be
+@option{--invert} which inverts the amount field, @option{--account
+@var{STR}} which you can use to specify the account to balance against,
+and @option{--rich-data} which stores additional metadata as tags.
+
+Using following command on the two first lines of the above csv file,
+
+@smallexample
```
afh: Any chance we can add meta-data so that the example is validated and tested by `test/DocTests.py`?

thdox: Unfortunately no, as DocTests.py is not able to handle multi-lines ledger commands. I copied-pasted an already existing comment to track it as TODO.

afh: Not anymore :laughing: fde30f19774f9b933722f9bdf33cce59df129980

thdox:
@afh, thanks for improving the DocTests.py enabling multi-line command.

I then have added a commit to add UUID to the input, the command, and the result.

BUT, I think it is not yet time to accept the Pull Request...

I launched DocTests.py using the command:

./test/DocTests.py -vv --ledger ./ledger --file ./doc/ledger3.texi

There is two issues:
- first, in my input, I intentionally omit the the line ",date,payee,note,amount,,,code,", and the DocTests.py does not detect that this line is mandatory.
- second, in the *result*, I have changed the amount of first transaction from 8.8 to 8.9. And then run DocTests.py. DocTests.py does not detect that 8.9 is a faulty value.

Is there any chance to implement Unit Tests on DocTests.py?

doc/ledger3.texi
```diff
@@ -4616,13 +4616,49 @@ from the file above.
...
+and launching the below command,
+
+@smallexample @c command:DD85579
+$ ledger convert download.csv --input-date-format "%m/%d/%Y" \
+  --invert --account Assets:MyBank --rich-data
+@end smallexample
+
+you will get the result:
+
+@smallexample @c result:DD85579
```
afh: Please use the `output` keyword here instead of `result`.

doc/ledger3.texi
```diff
@@ -4616,13 +4616,49 @@ from the file above.
...
+@smallexample @c input:DD85579
+767718,12/13/2011,"Withdrawal","ACE HARDWARE 16335 S HOUGHTON RD",-8.80,,00001640.04,,
+767406,12/13/2011,"Withdrawal","ACE HARDWARE 16335 S HOUGHTON RD",-1.03,,00001648.84,,
+@end smallexample
+
+and launching the below command,
+
+@smallexample @c command:DD85579
+$ ledger convert download.csv --input-date-format "%m/%d/%Y" \
+  --invert --account Assets:MyBank --rich-data
```
afh: Add `--now 2015/02/13` to the options, otherwise the test will fail, since the convert command will generate the `Imported` tag with the date the test is run.

afh:
I've improved `test/DocTests.py` so it will print when tests are skipped, the test you added was silently skipped since the `result` keyword (which isn't used in `test/DocTests.py`) was used instead of the `output` keyword.

Additionally I've added the `file` and `with_file` keywords so that it is possible to specify additional input data (apart from the `--file` option) to ledger which is needed for the `convert` command.

doc/ledger3.texi
```diff
@@ -4616,13 +4616,49 @@ from the file above.
...
-a UUID field. If an entry with the same UUID tag is already included in
-the normal ledger file (specified via @option{--file @var{FILE} (-f)} or
-via the environment variable @env{LEDGER_FILE}) this entry will not be
+@option{--invert} which inverts the amount field, @option{--account
+@var{STR}} which you can use to specify the account to balance against,
+and @option{--rich-data} which stores additional metadata as tags.
+
+Using the two first lines of the above csv file,
+
+@smallexample @c input:DD85579
```
afh: change `input` here to `file`.

doc/ledger3.texi
```diff
@@ -4616,13 +4616,49 @@ from the file above.
...
+767718,12/13/2011,"Withdrawal","ACE HARDWARE 16335 S HOUGHTON RD",-8.80,,00001640.04,,
+767406,12/13/2011,"Withdrawal","ACE HARDWARE 16335 S HOUGHTON RD",-1.03,,00001648.84,,
+@end smallexample
+
+and launching the below command,
+
+@smallexample @c command:DD85579
```
afh: add `,with_file:download.cv` at the end of this line.

thdox:
Closed, the pull request #394 is the correct one

PR #393, opened 2015-02-21
[doc] Improve documentation about --rich-data

thdox:


thdox:
Pull request not properly done, not properly based on next

PR #398, opened 2015-02-22
Improve documentation of --base option

thdox:
and other minor fixes


afh:
Thanks for the changes1

PR #400, opened 2015-02-23
Fix indentation and position of Fedora test.

jaccarmac:
The blame for this section was a bit of a cluster, but, as someone who's not
using Fedora, this makes the most logical sense. (Fedora is a version of RH,
right?)


jaccarmac:
Ooops. My bad.

thdox:
Thanks @jaccarmac to fix this merge bizarrerie. 
@ merge powered users, this can be merge.

PR #401, opened 2015-02-24
Improve baseline tests

thdox:
Remove --cache from untested options in CheckBaselinetests.py
Remove --*help* from untested options in CheckBaselineTests.py
Add --no-pager to untested options of CheckBaselineTests.py
Add baseline test for --script option.
Add baseline test for --price-db option.
Add baseline test for --init-file option.


afh:
Nice work, thanks!

PR #405, opened 2015-02-26
Add ledger-mode to generation of documentation script.

thdox:
[ci skip]

### DIFF
--- a/tools/gendocs.sh
+++ b/tools/gendocs.sh
@@ -21,9 +21,12 @@ if [ $(uname -s) = 'Darwin' ]; then
     && export PATH="$(brew --prefix texinfo)/bin:$PATH"
 fi
 
-echo "===================================== Making Info..."
-makeinfo ledger3.texi
-echo "===================================== Making HTML..."
-makeinfo --html --no-split ledger3.texi
-echo "===================================== Making PDF..."
-texi2pdf --quiet --batch ${papersize} ledger3.texi
+for texi in ledger3.texi ledger-mode.texi; do
+    echo "==================== ${texi}"
+    echo "===================================== Making Info..."
+    makeinfo ${texi}
+    echo "===================================== Making HTML..."
+    makeinfo --html --no-split ${texi}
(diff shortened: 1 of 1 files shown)


tools/gendocs.sh
```diff
@@ -27,3 +27,10 @@ echo "===================================== Making HTML..."
 makeinfo --html --no-split ledger3.texi
 echo "===================================== Making PDF..."
 texi2pdf --quiet --batch ${papersize} ledger3.texi
+
+echo "===================================== Making Info..."
+makeinfo ledger-mode.texi
+echo "===================================== Making HTML..."
+makeinfo --html --no-split ledger-mode.texi
+echo "===================================== Making PDF..."
+texi2pdf --quiet --batch ${papersize} ledger-mode.texi
```
afh: How about a small shell for loop to remove the duplicate code?

        for info in ledger3.texi ledger-mode.texi; do
          echo "===================================== Making Info..."
          makeinfo ${info}
          echo "===================================== Making HTML..."
          makeinfo --html --no-split ${info}
          echo "===================================== Making PDF..."
          texi2pdf --quiet --batch ${papersize} ${info}
        done

thdox: Sure! See commit. 
But to be transparent, I did minimum effort, expecting that you will do better job in new cmakefile ! :-)

afh:
If that's the case the script (c|sh)ould be deleted since `ledger-mode.texi` is already being built when `ledger3.texi` is built (see [`doc/CMakeLists.txt`:45](https://github.com/ledger/ledger/blob/next/doc/CMakeLists.txt#L45)) and as far as I know it's not being used anywhere.

thdox:
@afh, you are fully correct, and I was not aware about this. Thank you.
Now doing the following, I get the result expected:

    cmake . -DBUILD_DOCS=1 -DBUILD_WEB_DOCS=1 -DBUILD_A4_PDF=1
    make doc

PR #407, opened 2015-03-12
Fix clang compile error (FreeBSD 10).

kk12l:


afh:
Hi @kk12l,
thanks for your contribution. As stated in the [contributing guidelines](https://github.com/ledger/ledger/blob/next/CONTRIBUTING.md) please issue pull requests to the `next` branch, since that's where current development is happening.

Additionally could you please add the compile error output, I'm not sure when I'll be able to setup a virtual FreeBSD system to investigate, and I'd like to have a look, thanks!

jwiegley:
@afh Hi Alexis, I think that we should move active development to the master branch, so that people don't mistakenly submit their pull requests to the wrong place.  Whatever decreases inertia relating to code contributions I think will be better for us than rigorously following the git-flow model, as beautiful it that can be.  What do you think?

PR #413, opened 2015-04-18
Add command directives to concepts and commands indexes.

thdox:


thdox:
up

This one, is very safe to merge. This is only documentation related, and what it does, is that "keywords" are added to index of concepts and index of commands.

thdox:
Thanks John

PR #414, opened 2015-04-27
Use ./.ledgerrc if ~/.ledgerrc doesn't exist

davidsklar:
read_init() was using ~/.ledgerrc instead of ./.ledgerrc if $HOME was set, whether or not ~/.ledgerrc exists. This patch changes it so that ./.ledgerrc will be used if ~/.ledgerrc doesn't exist (or, still, if $HOME is not set for some reason.)


jwiegley:
And thanks for the whitespace fixes :)

davidsklar:
Don't thank me, thank 
````
(add-hook 'before-save-hook 'delete-trailing-whitespace)
````
:)

PR #415, opened 2015-05-04
Don't explicitly link libpython on OS X

tdsmith:
Use -undefined dynamic_lookup to allow Python symbols to be discovered
when the ledger module is imported instead of at build time. Without
this change, the Python interpreter crashes when ledger is imported from
a python that is different from (but compatible with) the python against
which the module was built.


jwiegley:
I'd like to know if this works for others...

tdsmith:
FWIW, this is how distutils links extension modules on OS X. Here's where the flags are set: https://github.com/python/cpython/blob/9a77656e0d3c6a737da7c2574f81afa459f67b13/configure.ac#L2214-L2245

markupsafe is a quick-building example – running `pip install -U --force-reinstall -v --no-use-wheel markupsafe`, the speedups module is linked as:
> clang -bundle __-undefined dynamic_lookup__ -L/usr/local/lib -L/usr/local/opt/sqlite/lib build/temp.macosx-10.10-x86_64-2.7/markupsafe/_speedups.o -L/usr/local/lib -L/usr/local/opt/sqlite/lib -o build/lib.macosx-10.10-x86_64-2.7/markupsafe/\_speedups.so

and the resulting installed binary doesn't have an explicit Framework linkage:
```
darkside:Formula tim$ otool -L /usr/local/lib/python2.7/site-packages/markupsafe/_speedups.so 
/usr/local/lib/python2.7/site-packages/markupsafe/_speedups.so:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1213.0.0)
darkside:Formula tim$
```

tdsmith:
Thanks!

PR #416, opened 2015-05-10
link ledger executable to Python on OS X

tdsmith:
#415 was slightly incorrect; I didn't appreciate that the `ledger` executable embeds a Python interpreter (for `ledger python`, I guess).

Since the ledger executable embeds a Python interpreter, it does need an explicit link to a Python framework on OS X after all.

The dynamic library (which is simply copied into the Python site-packages as the importable module) does not need this link.

This change adds the library linkage back to the `ledger` executable but not the shared library.

It's a little ungainly that the Python stuff is spread out between a couple of different CMakeLists now, I think; I'd be happy to apply some refactoring if there are suggestions.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -273,6 +273,9 @@ if (BUILD_LIBRARY)
 
   add_executable(ledger main.cc global.cc)
   target_link_libraries(ledger libledger)
+  if (APPLE AND HAVE_BOOST_PYTHON)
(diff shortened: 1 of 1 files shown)


tdsmith:
The change in #415 would have worked anyway because boost_python.dylib is incorrectly built with an explicit link to a Python framework and so the symbols were available. When Boost.Python is rebuilt to avoid an explicit linkage, ledger breaks without this additional change.

afh:
Thank you for your contribution to ledger!

PR #417, opened 2015-05-11
build fixes for boost 1.58

kljohann:
Can anyone check that this still works for lower boost versions?

### DIFF
--- a/src/account.h
+++ b/src/account.h
@@ -261,11 +261,7 @@ public:
   mutable optional<xdata_t> xdata_;
 
   bool has_xdata() const {
-#if BOOST_VERSION >= 105600
-    return xdata_ != NULL;
-#else
-    return xdata_;
-#endif
+    return xdata_ != boost::none;
   }
   void clear_xdata();
   xdata_t& xdata() {
--- a/src/filters.cc
+++ b/src/filters.cc
@@ -707,7 +707,7 @@ namespace {
     insert_prices_in_map(price_map_t& _all_prices)
       : all_prices(_all_prices) {}
 
-    void operator()(datetime_t& date, const amount_t& price) {
+    void operator()(const datetime_t& date, const amount_t& price) {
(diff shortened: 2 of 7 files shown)


thdox:
I am on Debian stable 8 aka jessie, with libboost-1.55
With this PR #417 ledger does not compile anymore.
Here is the log http://pastebin.com/EmKFEwEj
This has also been reported today by dodekaeder on IRC

jwiegley:
@kljohann I have reverted these two commits, due to the problems encountered by thdox.  We need to test whether these changes can be made to work with older version of Boost first.

Ikke:
This fixes building on Archlinux which already ships boost 1.58. I currently added these commits as patches when I build the packages.

jwakely:
Hi - why was this fix reverted again?

jwiegley:
@jwakely See comment from @thdox above.

jwakely:
Is there a solution that works with both 1.55 and 1.58 though? The current Ledger master doesn't work with the current Boost release.

thdox:
Here is pastebin again, as first one expired.
http://pastebin.com/YHXVvhGe

I would be happy if someone find a fix that enables compilation with both boost 1.55 and 1.58. If you want me to test compilation, please send patch or PR.

jwakely:
This error suggests that making the `xdata_t` type EqualityComparable would work:

    /usr/include/boost/utility/compare_pointees.hpp:32:52: error: no match for ‘operator==’ (operand types are ‘const ledger::post_t::xdata_t’ and ‘const ledger::post_t::xdata_t’)
       return (!x) != (!y) ? false : ( !x ? true : (*x) == (*y) ) ;
                                                        ^

However, since it seems all the problematic lines are simply trying to test if the optional is empty, why not simply do that explicitly instead of comparing to `NULL` or `none`:

    return (bool)x;

PR #418, opened 2015-05-30
Do not use set-mark in ledger-post-align-postings

egh:
In emacs 25, save-excursion no longer saves the mark. Running
ledger-post-align-postings now sets the mark, which is annoying.
Moreover, this use of set-mark appears to do nothing, so I removed it.

### DIFF
--- a/lisp/ledger-post.el
+++ b/lisp/ledger-post.el
@@ -100,22 +100,14 @@ at beginning of account"
   (let ((bounds (ledger-navigate-find-xact-extents pos)))
     (ledger-post-align-postings (car bounds) (cadr bounds))))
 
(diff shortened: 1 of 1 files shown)


enderw88:
I need to make sure this doesn't break something in older emacsen.  I can't remember why I did this, but was there for a reason.

egh:
I'll test it out in emacs 24

egh:
Is `ledger-post-align-posting` supposed to work on the region if mark is set? It doesn't, maybe something is broken there.

egh:
Oops, it does work, I was calling `C-c C-q` (`ledger-post-align-xact`)

egh:
Simplified by using `(interactive "r")`

PR #420, opened 2015-07-06
Emacs: Only reconcile real balances

alanpearce:
Use only real transactions when calculating the cleared or pending balance.

With this, the cleared/pending amount will reach 0 once all real transactions are correctly reconciled,
whereas it would previously be off by the total of all virtual transactions.


enderw88:
This needs to be configurable. I actually balance virtual accounts.

PR #424, opened 2015-08-04
Add regression test file for bug #1057

thdox:
I never understood why PR #305 gathered so many commits not from me. Re-submitting the original one with only 3 commits.


jwiegley:
Thanks!

PR #427, opened 2015-09-17
doc for get_at function

paulbdavis:
Simple doc string for get_at


afh:
Thank you for your contribution

PR #428, opened 2015-09-18
fontify recurring transaction in ledger-mode

paulbdavis:
Adds a regexp and catches for allowing files containing recurring transactions to be fontified the same (in most cases) as a regular xact


afh:
Please add `[ci skip]` to the end of the commit message by amending it, so that continuous integration is being run for changes that do not affect the tests. Thanks

PR #431, opened 2015-09-19
Add dwim align

Fuco1:
This patch adds two changes:

1. Add a function to dwim (do what I mean) align: if no region is active, align the xact under the point, if a region is active, align that region.
2. Bind this function to `M-q`.  This might be controversial so I put it in a separate patch, but I think it is a very reasonable idea to replace `fill-paragraph` with something that actually makes sense in ledger mode.


jwiegley:
I'm ok with M-q meaning "reformat, by which I mean, align" in Ledger-mode.  I use M-q to mean exactly that in Lisp mode, when outside of a comment.

PR #432, opened 2015-09-21
Fix amount-regex to properly capture integer part.

Fuco1:
If the integer part was 2 digits, because the integer part was
non-greedy, it only captured as much as necessary, namely one digit.
The other digit was captured by the 5th group's [:word:] not 4th because
that one is also optional.  This results in incorrect reporting of
amount length and broken aligning.

Also move the decimal dot matching from 3rd group to 4th, and quote the
dot.


Fuco1:
Btw, there's a somewhat obscure feature in emacs called `re-builder`, you can start it with `M-x re-builder`, it allows you to visually inspect what is going on when you compose a RE.  Very useful with complex regexps like this one, I used it for great success :)

PR #434, opened 2015-09-28
Align amounts on the decimal separator.

Fuco1:
That is:

    A   $10.00
    B    $5
    C  -$15.00

Before the numbers aligned at the end of the last digit.

@thdox reported this issue to me.  My last align patch broke `,` as decimal separator, that is also fixed.  The bejaviour I "benchmarked" my patch against was master which aligned numbers on the last digit:

    A   $10.00
    B       $5
    C  -$15.00

We both found that undersirable.

If anyone has some edge-cases they feel like we should check, please share :)

### DIFF
--- a/lisp/ledger-post.el
+++ b/lisp/ledger-post.el
@@ -79,8 +79,7 @@ point at beginning of the commodity."
     (when (re-search-forward ledger-amount-regex end t)
       (goto-char (match-beginning 0))
       (skip-syntax-forward " ")
-      (- (or (match-end 4)
-             (match-end 3)) (point)))))
+      (- (match-end 3) (point)))))
 
 (defun ledger-next-account (&optional end)
   "Move to the beginning of the posting, or status marker, limit to END.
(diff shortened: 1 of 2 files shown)


thdox:
@Fuco1 I had a look, and it looks good!

Actually, I told you something wrong, that the amounts were aligning on the decimal separator. Actually there were aligning on the space between the figure and the commodity. But, you know what, I slightly tend to prefer the new way of working.

There is something that is less appealing to the eye, it is when I have several postings, and there are different numbers of figures after the decimal. But I cannot imagine a better solution.

I vote for merging.

jwiegley:
@Fuco1 I missed this pull request.  I'm quite happy to have decimal-aligned behavior, so long as we don't lose the ability to also have right-aligned behavior.  Some people want the latter, and not the former, and have come to expect the right-aligned behavior.

Can you please allow the option of having either, with a `defcustom` to switch between the two?  For now at least, the right-aligned variant should be the default, following the principle of least surprise.

Fuco1:
Right.  I'll see what I can do, hopefully this won't require some regexp black magic (any more that there already is anyway).

PR #436, opened 2015-10-09
Fix amount regexp in context computation.

Fuco1:
Before the fix the amount required decimal point even when the decimal
digits were optional.  Now both are optional.

With a transaction

```
2015/09/29 * Rent
    Expenses:Utilities:Rent                     5000 Kc
    Assets:Checking:CS
```

and point at `Expenses` the context returned by `(ledger-context-at-point)` was

```elisp
(acct-transaction account
                  ((indent "    " 3104)
                   (status nil nil)
                   (account "Expenses:Utilities:Rent                     5000 Kc" 3108)))
```

Notice how the acount name includes the amount and the commodity.  After this patch the result is

```elisp
(acct-transaction account
                  ((indent "    " 3104)
                   (status nil nil)
                   (account "Expenses:Utilities:Rent" 3108)
                   (separator "5000" 3152)
                   (amount nil nil)
                   (commodity "Kc" 3157)))
```


Fuco1:
Hah... and immediately after posting this I see that the `amount` is now a `separator`, because I used capturing group.  Let me fix it right away *blushes*.

Fuco1:
Seems like the bug was there all along, as `ledger-extract-context-info` expects to extract a field out of every "name" but `separator` doesn't capture anything... so all shifts back one slot.  Was that intentional?

PR #437, opened 2015-10-14
Add an option to change amount-alignment mode.

Fuco1:
Make the default alignment behaviour the same as before 767ab3e, that
is, align on the last digit by default.  A new option
`ledger-post-amount-alignment-at' now exists to adjust alignment
preference.

This patch addresses @jwiegley's request from #434.

### DIFF
--- a/lisp/ledger-post.el
+++ b/lisp/ledger-post.el
@@ -41,6 +41,16 @@
   :type 'integer
   :group 'ledger-post)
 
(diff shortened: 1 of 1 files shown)


Fuco1:
What do you think about the terminology? `:end` and `:decimal` seem logical to me but I'd like to hear another opinion.

jwiegley:
I'm pro-customize, so long as you made it clear there, I'm cool with whatever keywords you like.

PR #439, opened 2015-11-11
Changed drewr.dat to drewr3.dat in docs

cerk:
I had a problem when going through the tutorial portion of the ledger manual. It was easier for me to cut and paste from the manual to generate the tutorial journal file (drewr3.dat) instead of using the file from the repo itself. However, when I did so, I encountered a run-time error during the 2.2.3 Cleared Report section. After trying to track down what was going on, I realized that the example journal in the manual was actually drewr.dat and not drewr3.dat. So this pull request is to fix the docs so that the example journal is actually drewr3.dat. However, I actually think you should disregard this pull request and fix this problem in a more elegant fashion. Because drewr3.dat has commented out sections and tags that may or may not aid comprehension for the new ledger user (and I just cut and paste drewr3.dat into the docs).

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -10152,7 +10152,7 @@ is now @env{LEDGER_PRICE_EXP}.
 @appendix Example Journal File
 
 The following journal file is included with the source distribution of
-ledger.  It is called @file{drewr.dat} and exhibits many ledger
+ledger.  It is called @file{drewr3.dat} and exhibits many ledger
 features, include automatic and virtual transactions,
 
 @smallexample @c input:validate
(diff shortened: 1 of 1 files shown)


afh:
I'm sure you are well aware that there are two files in `test/input`:
* [drewr.dat](https://github.com/ledger/ledger/blob/next/test/input/drewr.dat)
* [drewr3.dat](https://github.com/ledger/ledger/blob/next/test/input/drewr3.dat)

Can you please provide the exact commands you ran when you encountered the run-time error.

cerk:
In the docs, in Section 2.1 Star a Journal File, it references drewr3.dat and then links to Appendix A. But Appendix A shows the file drewr.dat, not drewr3.dat. So if you were to use the journal file in Appendix A while going through the Section 2 Ledger Tutorial, you would get incorrect results. Also, when going through Section 2.2.3 Cleared Report, if I run

```
ledger -f drewr.dat cleared
```

then I get:

```
    (latest_cleared ? (format_date(latest_cleared) : "               "))
Error: Assertion failed in "../ledger/src/ledger-3.1/src/post.cc", line 109:virtual ledger::date_t ledger::post_t:primary date() const: xact
```

What I'm saying is that if a beginner uses the Appendix A journal file, they will be using the wrong file (they will be using drewr.dat instead of drewr3.dat). But the docs sort of imply that it is drewr3.dat. So to avoid confusion, that Appendix A should change to drewr3.dat. I could be completely wrong here because I am a ledger beginner, but this is the problem I encountered when going through the tutorial in the docs.

Note that

```
ledger -f drewr3.dat cleared
```

runs fine without error. It is just that I *thought* I was using drewr3.dat (that I had cut and paste from Appendix A), but I was actually using drewr.dat.

afh:
Thanks for clarifying it and thanks for your contribution.

PR #442, opened 2016-01-08
Fix highlight of the transaction under point

daniele-nicolodi:
Extend the highlighting of the last line in the transaction to the
right margin of the buffer. Do not highlight empty lines.


Fuco1:
Plusone from me, this is a neat feature and consistent with the Emacs UI (new rectangles, hl-line-mode)

enderw88:
Very nice, I have been meaning to fix this one since I wrote the new overlay code. Thank you!

PR #443, opened 2016-01-11
[python] Remove double quotes from unicode values

afh:
When converting a `ledger.Value` to unicode the Python API added
double quotes around it.

It looks like a bug that was introduced in d5e95720, @jwiegley would you mind having a look please.

To me it does not make sense to have a second `.def("__unicode__", py_dump_relaxed)` in `src/py_value.cc:323` that overwrites the previous `.def("__unicode__", py_value_unicode)` in `src/py_value.cc:317` (see [`src/py_value.cc:317ff`](https://github.com/ledger/ledger/commit/d5e95720#diff-20451a5ff4ce15ef3adc1ad28121d16aR317) or `git show d5e95720:src/py_value.cc | nl -ba | sed -ne '315,325p'`).

### DIFF
--- a/src/py_value.cc
+++ b/src/py_value.cc
@@ -325,7 +325,6 @@ void export_value()
     .def("to_mask", &value_t::to_mask)
     .def("to_sequence", &value_t::to_sequence)
 
-    .def("__unicode__", py_dump_relaxed)
     .def("__repr__", py_dump)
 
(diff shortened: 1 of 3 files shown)


afh:
Thanks for merging!

jwiegley:
Thanks for the code! :)

PR #445, opened 2016-01-25
[doc] Add "Dependency" anchor link to quick start section

nchelluri:
The quick start section is already called "For the Impatient", and this
minor doc tweak should decrease the time and effort demanded of a
new user before their first run by a very tiny amount, and it makes the
README a touch more cohesive.

[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,8 +23,8 @@ and easy to use.
 
 ## For the Impatient
 
-I know, you just want to build and play.  If you have all the dependencies
-installed (see below), then simply do this:
+I know, you just want to build and play.  If you have all the [dependencies](#dependencies)
+installed, then simply do this:
 
     $ git clone git://github.com/ledger/ledger.git
     $ cd ledger && ./acprep update  # Update to the latest, configure, make


README.md
```diff
@@ -23,8 +23,8 @@ and easy to use.
 
 ## For the Impatient
 
-I know, you just want to build and play.  If you have all the dependencies
-installed (see below), then simply do this:
+I know, you just want to build and play.  If you have all the [dependencies
+installed](#dependencies) (see below), then simply do this:
```
afh: In my opinion `installed` should not be part of the link.
If `dependencies` is a link then I find `(see below)` obsolete.
Another possibility would be to make `see below` a link.

Just my 2¢

nchelluri: Have updated the PR. Thoughts?

afh:
Please squash the three commits into one keeping the `[ci skip]` line.

nchelluri:
Should be good now. I wish I could rename the branch but I think that'll close the PR and I'll have to create a new one. Happy to do that if you'd prefer. Thinking of a name like `minor-doc-tweak`.

afh:
It's fine like this and I merged it. Thank you for your contribution to the project! :+1:

PR #449, opened 2016-03-05
Use interval start date (period from/since) to initialize first period.

ppickfor:
This allows the offset of a period start to be specified

-p 'every 12 months from 2000-04-01'

will have periods

yyyy-04-01 to yyyy-03-31
....

If no from/since is specified for the period the first transaction date reported is used to initialize the period as before.

### DIFF
--- a/src/filters.cc
+++ b/src/filters.cc
@@ -982,10 +982,12 @@ void interval_posts::flush()
   std::stable_sort(all_posts.begin(), all_posts.end(),
                    sort_posts_by_date());
 
-  // Determine the beginning interval by using the earliest post
-  if (all_posts.size() > 0 && all_posts.front() &&
-      ! interval.find_period(all_posts.front()->date()))
-    throw_(std::logic_error, _("Failed to find period for interval report"));
+  // only if the interval has no start use the earliest post
+  if (!(interval.begin() && interval.find_period(*interval.begin())))
+    // Determine the beginning interval by using the earliest post
+    if (all_posts.size() > 0 && all_posts.front()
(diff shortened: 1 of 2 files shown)


tbm:
Can you also add some test cases for that?

tbm:
Does that fix bug 370?  if so, might be worth mentioning in the changelog (plus test case)

ppickfor:
Thanks for the feedback.

What is the file naming convention for tests and which directory should they go in?
I'm thinking test/regress and some fort of bug number seems to be used for some and hex for others.

It does not fix bug 370
That is due to date stabilization using a start date of date-400 days for days and weeks; the period start date is then offset by the remainder for 400/duration but the explicit start date is used for the first period - very weird.

I believe that if an from/since date is passed that should the the exact start of the period then there is no need for stabilization (which establishes start date based on the previous date quanta (year, quarter, month) - I believe days have quanta of 1day so there always at the start of a period.

I made some changes to times.cc but its only a partial fix.
I don't understand why there is complication around establishing the period start for a quanta date_duration_t::find_nearest looks sufficient.

'1 weeks from 2011/10/01' behaves differently to '2 weeks from 2011/10/01' - not sure why yet.

Do you agree this is the way -p should work - if a from/since date is used -b can still be used to limit reported transactions via start date?

Also I'm not sure how this relates to periodic transactions I haven't used them yet.
Periodic transactions could work the same way but if you want a start date it would have to correspond to the nearest quanta start date if you want the transactions on a default quanta start (jan 1st, 1st of month, week start etc)

ppickfor:
raise bug 1159 for this

PR #450, opened 2016-03-14
Don't stabilize days periods, reduce number of iterations to find start of period for weeks

ppickfor:
My attempt at making day periods have the specified start date.
May resolve bug 370 by replacing biweekly with 14 days in the periodic transaction.


jwiegley:
Thank you, @ppickfor!

PR #456, opened 2016-06-17
Fix buildability on MSYS2/MinGW32.

jaccarmac:
The build process on MSYS2 is a lot easier than manually installing
Boost and using Visual Studio, in my opinion. However, the source was
not set up to build properly in this environment. When running `make` in
a standard MSYS2/MinGW32 environment, `WIN32` is not set, but `_WIN32`
and (maybe) `__WIN32__` are. Checking these along with the unqualified
version catches all the varieties of Windows builds.

`t_value`'s header was not set up to properly import the `strptime`
replacement on Windows, so it is added.

The process for building on MSYS is similar to the method for Visual
Studio documented in the Wiki. All the work is done from inside
the *32-bit* MinGW shell. The basic development pipeline, CMake, and
Boost all need to be installed (from inside MSYS and for the MinGW32
environment).

Run `cmake -G "MSYS Makefiles"` to create build files.

Edit `system.hh`, replacing all the blank `#define`s in the application
configuration section with `#define VARNAME 0`.

Set `HAVE_ISATTY` in `system.hh` to 0 as well; It's wrong.

Run `make`.

After building, `ledger.exe` should work fine from inside the MSYS
shell, but outside it needs dynamic access to some DLLs. Tracking down
exactly which ones is nasty; I used Dependency Walker for it. Note that
if you have Git for Windows installed on your PATH, libstdc++ will seem
like it exists, but as the wrong architecture. Put the 32 bit libstdc++
in the EXE's directory.

In my case, the necessary DLLs were:

libboost_filesystem-mt.dll
libboost_regex-mt.dll
libboost_system-mt.dll
libgcc_s_dw2-1.dll
libgmp-10.dll
libicudt57.dll
libicuuc57.dll
libledger.dll
libstdc++-6.dll
libwinpthread-1.dll

### DIFF
--- a/src/global.cc
+++ b/src/global.cc
@@ -451,7 +451,7 @@ expr_t::func_t global_scope_t::look_for_command(scope_t&      scope,
 
 void global_scope_t::visit_man_page() const
 {
-#ifndef WIN32
+#if !defined(_WIN32) && !defined(__WIN32__) && !defined(WIN32)
   int pid = fork();
   if (pid < 0) {
     throw std::logic_error(_("Failed to fork child process"));
--- a/src/main.cc
+++ b/src/main.cc
@@ -74,7 +74,7 @@ int main(int argc, char * argv[], char * envp[])
 #endif
 
   std::signal(SIGINT, sigint_handler);
-#ifndef WIN32
+#if !defined(_WIN32) && !defined(__WIN32__) && !defined(WIN32)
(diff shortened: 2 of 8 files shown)


schmave:
fyi I incorporated these changes into my PR #462. Thanks for sharing them!

jaccarmac:
Sounds good, thanks! I'll be watching for the close on that pull.

PR #458, opened 2016-06-28
improve explanation about period expressions

arademaker:
as suggested in the mailing list.


jwiegley:
@arademaker The only strange thing about saying "these are instances in time" is that you cannot specify the time.  They really *are* full days, it's just exclusive, not inclusive.

arademaker:
Yes, it makes sense! ;-)

PR #460, opened 2016-07-16
Makes links in ledger-report "register" commands optional.

atheriel:
This commit adds a new custom `ledger-report-links-in-register` flag that controls whether `ledger-report` will attempt to link transactions in the report buffer to the entries in the original file. Essentially, it controls whether `ledger-report` will silently pass the `--prepend-format` flag to the binary.

The primary motivation for this is that not all implementations of ledger support the `--prepend-format` flag, and thus all "register" commands run by `ledger-report` (and usually *only* "register" commands)
will always fail, because they are treated as a special case in `ledger-do-report`.

At present, the user has no ability to prevent the `--prepend-format` flag from being silently passed to the binary. This PR provides that feature.

### DIFF
--- a/lisp/ledger-report.el
+++ b/lisp/ledger-report.el
@@ -82,6 +82,12 @@ text that should replace the format specifier."
   :type 'boolean
   :group 'ledger-report)
 
+(defcustom ledger-report-links-in-register t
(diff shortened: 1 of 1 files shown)


enderw88:
Thank you.

PR #463, opened 2016-09-05
Fix strict/pedantic breakage due to non-uncleared transactions

### LINKED ISSUE #838: account directive not giving warning with --strict (BZ#838)

Bugzilla bug ID: BZ#838
From: John Wiegley (@jwiegley)
CC:   @thdox

CandyAngel:
Removing these checks will stop ledger from regarding accounts and commodities which are part of pending/cleared transactions as being defined/known.

This fixes the unexpected (and unhelpful?) behaviour of strict/pedantic not working as expected (warn/error if an account has no 'account' directive defining it).

I believe this resolves #838, #844 and #846 on Bugzilla.

Please note that I haven't run the test suite with these changes as following the instructions yielded no output.

### DIFF
--- a/src/journal.cc
+++ b/src/journal.cc
@@ -152,9 +152,6 @@ account_t * journal_t::register_account(const string& name, post_t * post,
           fixed_accounts = true;
         result->add_flags(ACCOUNT_KNOWN);
       }
-      else if (! fixed_accounts && post->_state != item_t::UNCLEARED) {
-        result->add_flags(ACCOUNT_KNOWN);
-      }
       else if (checking_style == CHECK_WARNING) {
         current_context->warning(_f("Unknown account '%1%'") % result->fullname());
       }
@@ -267,13 +264,6 @@ void journal_t::register_commodity(commodity_t& comm,
       if (context.which() == 0) {
         if (force_checking)
           fixed_commodities = true;
-        comm.add_flags(COMMODITY_KNOWN);
-      }
-      else if (! fixed_commodities &&
-               ((context.which() == 1 &&
-                 boost::get<xact_t *>(context)->_state != item_t::UNCLEARED) ||
-                (context.which() == 2 &&
-                 boost::get<post_t *>(context)->_state != item_t::UNCLEARED))) {
         comm.add_flags(COMMODITY_KNOWN);
       }
       else if (checking_style == CHECK_WARNING) {


afh:
@CandyAngel Thank you for this pull-request:pray:

Please have a look at the [failed tests](https://travis-ci.org/ledger/ledger/jobs/157719073):boom:and update this PR accordingly (either by changing the tests, if they tested the "wrong" behaviour or by fixing the issues this PR introduced).

The [failed tests](https://travis-ci.org/ledger/ledger/jobs/157719073) output also shows how ledger is compiled and the tests are run; if you would like to run them locally it's basically:

```sh
% cmake . -DUSE_PYTHON=ON  -DBUILD_DEBUG=ON
% make
% ctest --output-on-failure
```

CandyAngel:
Sorry for the delay, when I ran the tests on my local machine, I got a lot more than the 2 failures Travis shows, then got sidetracked with other projects. I'll address this shortly!

thdox:
@CandyAngel , perhaps you did not move away your ~/.ledgerrc file ? I come into this frequently... I use something like:

`mv ~/.ledgerrc{,.bak} ; make check ; mv ~/.ledgerrc{.bak,}`

jwiegley:
@CandyAngel Are we ready to merge?

tbm:
@CandyAngel are you still there? Can you update this PR? Can you add a test case? Is there a bug number to reference?

twrightsman:
Bumping because this is unfortunately causing problems with my ledger workflow. I sadly don't have the required C++ expertise to help with pushing this fix through, but it seems the tests are failing as designed and need to be rewritten to test for the new desired behavior (pending/cleared transactions now error out with `--pedantic` the same as uncleared transactions).

CandyAngel:
Clearly, I forgot the ™ after "shortly"!

I will look into fixing this properly tomorrow.

CandyAngel:
As this issue was also affecting other elements (payees and tags), I've created a new PR (#1819).

CandyAngel:
> desired behavior (pending/cleared transactions now error out with `--pedantic` the same as uncleared transactions).

Just a quick note that adding the `--explicit` option will cause pedantic/strict to be applied regardless of posting state. The current discussion in the new PR is whether this should be the default behaviour (and `--explicit` be dropped as an option).

PR #462, opened 2016-08-13
Continuous integration for windows

schmave:

PR #466, opened 2016-09-17
Fix build under msys32; add Appveyor continuous build

schmave:
* Appveyor is a continuous integration platform for Windows that is free for open source projects. See latest test build here: https://ci.appveyor.com/project/Evan/ledger/build/artifacts

* Changed WIN32 to _WIN32 because this article [http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system] claims that it is defined by the greatest number of compilers.

* Modified HAVE_ISATTY and other system defines so that system.hh compiles when cmake fails to find the relevant function.

* Add missing include in test/unit/t_value.cc

Almost all tests fail when you run them on msys32. I will address that in a future PR.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,28 @@
+version: build-{build}
+
+image: Visual Studio 2013
+
+shallow_clone: true  # don't download repo history
+
+install:
+- time /t
+- tzutil /s "Eastern Standard Time_dstoff"
+- time /t
+
+build_script:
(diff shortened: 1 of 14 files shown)


schmave:
@jaccarmac @tbm Here's the first PR that I broke off of #462. Let me know what you think.

jwiegley:
@schmave Thank you!

jaccarmac:
Beautiful. Thanks @schmave + @jwiegley!

PR #468, opened 2016-10-02
Fix build under Cygwin

Vanav:
Cygwin doesn't define _WIN32, and if you define it, build will fail.


thdox:
There is a regression with this Pull Request. The build does not pass on osx/clang. I believe this is related to changes in src/strptime.cc

jwiegley:
@Vanav Can you please provide a fix for this?

Vanav:
I expected my changes are safe. Can you show error from log?

jwiegley:
@thdox Do you have a log?  It actually built for me with 10.11/clang.

PR #469, opened 2016-10-03
Fix regress test 383 on Windows

schmave:
%g is not available in Windows strftime. See documentation at https://msdn.microsoft.com/en-us/library/fe06s4ak.aspx

If this test is meant specifically to test %g as opposed to %y, I will have to come up with another way of fixing this.


thdox:
Thanks @schmave . This fix regression test. Most probably %g is an error, I cannot imagine what could be the usefulness of it... Can be merged from my point of view.

PR #472, opened 2016-10-05
Fix bug 1187: Allow balance assertions with multiple posts to the same account

mk12:
Fixes [Bug 1187](http://bugs.ledger-cli.org/show_bug.cgi?id=1187)

@thdox also discussed this in #451

Before, balance assertions/assignments just subtracted the amount in the current post. With these changes, they also subtract all amounts posted to the same account so far in the xact. I think anyone who wrote such a transaction would expect this behaviour.

I added two regression tests as well.


thdox:
Hi @mk12

Thank you very much for this first patch. It **corrects** the use case I described in PR #451 :-)

However, my ledger file is still not passing. Here is another use case, when there are several commodities on the same account.

```
2001/01/01 * Balance
    Actif:Courant:Cc                  10000,00 F = 10000,00 F
    Equity

2002/01/11 * Passage à l'euro
    Actif:Courant:Cc                 -10000,00 F = 0,00 F
    Actif:Courant:Cc                   1524,49 € = -1524,49 €
    Revenu:Devise                    -10000,00 F
    Revenu:Devise                      1524,49 €
```

Is there something you can do? Maybe checking the commodity while computing the balance ?
Thank you so much for your help.

mk12:
@thdox, is there a mistake in your example? That second transaction looks me like it doesn't balance.

I see what you mean though. Currently I am subtracting the amounts from previous postings to the same account in the same transaction without regard to commodities. I should change it to only subtract amounts of the same commodity as the current posting. I'll try to do this tonight or on the weekend.

thdox:
@mk12, yes, that looks to 
```
2001/01/01 * Balance
    Actif:Courant:Cc                  10000,00 F = 10000,00 F
    Equity

2002/01/11 * Passage à l'euro
    Actif:Courant:Cc                 -10000,00 F = 0,00 F
    Actif:Courant:Cc                   1524,49 € = 1524,49 €
    Revenu:Devise                     10000,00 F
    Revenu:Devise                     -1524,49 €
```
I tried myself, based on your pull request, but this is much harder than I thought, I realized that I am not a c++ guy, and that I lack knowledge of the ledger's internals...
Thank you to look at it
Cheers

mk12:
@thdox, I have updated the PR with your two examples as 1187_1.test and 1187_2.test, respectively. Both are passing with my new changes.

@jwiegley, What's up with the clang build? It seems to be failing for your `next` branch as well. In any case, let me know if this looks good to merge :)

thdox:
# THANK YOU :+1: 

@mk12, thank you so much, it does work on my ledger file now. FWIW, I did not tell that I tried to fix myself, and spend something like 6 hours with a lot more lines of codes, and time to compile, but without success. Sometimes Internet is beautiful. You made my day.

tbm:
@mk12, I want to echo Thierry's comments. Your work is very welcome!

mk12:
Glad to be of help :)

@tbm, are you able to merge this? CI tests passed for gcc, and the clang failure seems be happening on the `next` branch already.

thdox:
Thanks for merging :+1:

PR #474, opened 2016-10-27
Ledger mode finalize split

thdox:
As ledger-mode has now its own github repository, these changes are updating files which are still referencing old situation.


jwiegley:
@thdox Thanks!

PR #477, opened 2016-11-01
Add regress test for bugs 550 and 584

thdox:
The file spaces.dat was unused in ledger source code base.
I moved it to test/regress/ so that it is used as regress test for bugs 550 and 584.


thdox:
No comment on this PR has been done since November 1st. I estimate this PR is safe enough to merge it myself.

PR #478, opened 2016-11-01
Fix indent-tabs-mode to nil

thdox:
Probably a bug in Emacs:
- This add 't' in .dir-locals.el
`M-x add-dir-local-variable RET emacs-lisp-mode RET indent-tabs-mode RET t   RET`
- This does not add 'nil' in .dir-locals.el
`M-x add-dir-local-variable RET emacs-lisp-mode RET indent-tabs-mode RET nil RET`

Fixing this eases the diff between files (spaces instead of tabs). Even if ledger-mode has been moved to its own repository, there are still Emacs lisp files in this repository.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -32,5 +32,5 @@
     (c-special-indent-hook . c-gnu-impose-minimum)
     (c-block-comment-prefix . ""))))
  (emacs-lisp-mode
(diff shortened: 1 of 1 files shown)


thdox:
No comment on this PR has been done since November 1st. I estimate this PR is safe enough to merge it myself.

PR #482, opened 2016-11-17
Fix MSVC build

AndrewSav:
This commit: https://github.com/ledger/ledger/commit/c2785590834dfe91574367f4256918c37b6ddfcc fixed the cygwin build but broke MSVC build. This pull request is to fix this.

### DIFF
--- a/src/strptime.cc
+++ b/src/strptime.cc
@@ -21,6 +21,7 @@
 #include <ctype.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


AndrewSav:
Ledger MSVC build procedure I'm using is described here: https://github.com/AndrewSav/ledger-windows-build please feel free to let me know if there a more proper one

PR #484, opened 2016-12-20
Fix Bug 1182: Error message for parse failure after '='

mk12:
Fixes [Bug 1182](http://bugs.ledger-cli.org/show_bug.cgi?id=1182)
- Handled case where parsing a predicate in an automated transaction fails.
- Before it crashed with SIGABRT on `operator->` in Boost's intrusive_ptr.
- Now it throws an error "Expected predicate after '='".
- Added two regression tests.


mk12:
CI succeeded for gcc and failed for clang. It seems like that's been happening for a while now though.

PR #485, opened 2017-01-10
Added support for `ledgerfmt`. Similar to `gofmt` for Ledger files.

abourget:
I propose adding this to .emacs files:

```lisp
(add-hook 'ledger-mode-hook (lambda ()
                          (local-set-key (kbd "C-c C-f") 'ledgerfmt)))
```

See https://github.com/abourget/ledger for `ledgerfmt`.

### DIFF
--- a/lisp/ledger-format.el
+++ b/lisp/ledger-format.el
@@ -1,0 +1,99 @@
+;;; ledger-format.el --- Format your ledger similar to `gofmt`
+
+;; Copyright (C) 2003-2016 Alexandre Bourget (alex AT bourget DOT cc)
+
+;; This file is not part of GNU Emacs.
+
+;; This is free software; you can redistribute it and/or modify it under
+;; the terms of the GNU General Public License as published by the Free
+;; Software Foundation; either version 2, or (at your option) any later
+;; version.
+;;
+;; This is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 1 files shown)


enderw88:
I had never heard of ledgerfmt before this pull request.  I think this belongs with ledgerfmt, not in the ledger base code.

abourget:
Ledgerfmt got created yesterday night, that's why :-)

I know it could be its own package, but it seems more practical to have a
single one... Even if it calls third party tools. That's how it's
integrated in the Go tools for example. gofmt is in a different package,
yet there's one go-mode.

Also, since it doesn't affect the rest of the package, does it hurt to be
there?  It only allows (slightly) more discoverability.

enderw88:
I don't have a strong feeling either way, but would it be more appropriate over in ledger-mode then?  Since this is very specific to emacs.

abourget:
Oh is there another ledger-mode repo?  I've been using this one in Emacs.

enderw88:
We pulled ledger-mode out a few months ago.  There isn't any elisp code left here anymore.  See https://github.com/ledger/ledger-mode

abourget:
Ok let's close this then. I'll reopen over there.

tbm:
Closing since ledger-mode moved to another repo.

PR #486, opened 2017-01-27
typo in docs

RusAlex:


tbm:
I am not entirely sure how this works but I believe you should also update test/baseline/feat-value-expr.test to keep things in sync.

PR #487, opened 2017-02-17
Improve wording, remove connection between financial worth and personal worth.

craig9:


enderw88:
This seems a bit pedantic, but sure. Maybe we should encourage people to track their karma with ledger as well.

PR #488, opened 2017-02-23
Add mention of third-party tutorials to documentation

bkuhn:
I was updating the tutorial that Conservancy published regarding how it uses Ledger CLI for non-profit accounting and realized that mentioning the tutorial's availability in Ledger's own documentation might be useful as well.

I hereby assert that the copyrights on this contribution are my own and I license them under the license found at the top of file to which this patch applies.


enderw88:
"Once such" --> "One such"

PR #494, opened 2017-10-26
Fix bug where .total used in value exprs breaks totals

jjlee:
Sorry my C++ is probably terrible...

* Re-initialize (to VOID) totals for the account and its ancestors on adding
  postings.  Otherwise the cache intended for use by recursive calls of C++
  function total() in computing family (i.e. account hierarchy) totals is
  incorrectly retained from one top-level call to the next, causing
  inconsistent and broken behaviour.

* Re-initialize (to false) calculated and gathered.  Otherwise we won't
  e.g. recalculate stale totals for ancestor accounts (e.g. won't recalculate
  Assets:Savings total if Assets:Savings changes via a posting).

Although the value expression total function is used by ledger itself in
computing totals, this bug would only appear on use of .total in user-supplied
value expressions computed *during parsing* of ledger files, rather than after
parsing (I believe ledger only ever calls it for internal purposes after
parsing is complete).

It is possible this bug also affected other functions than total (perhaps even
in circumstances other than analagous to that described in the preceding
paragraph).  I have not checked that.

### DIFF
--- a/src/account.cc
+++ b/src/account.cc
@@ -136,6 +136,19 @@ void account_t::add_post(post_t * post)
     xdata_->self_details.calculated   = false;
     xdata_->family_details.gathered   = false;
     xdata_->family_details.calculated = false;
+    if (! xdata_->family_details.total.is_null()) {
+      xdata_->family_details.total = ledger::value_t();
+    }
+    account_t *ancestor = this;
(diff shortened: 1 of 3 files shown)


jwiegley:
Nice work!

PR #495, opened 2017-10-26
Add value expression function debug

jjlee:
Add new value expression function `debug`: prints the value of a value
expression to stderr.

Probably this should be called `print`?  Or possibly should instead add a
second argument to assert which is printed when the assertion fails (similar
meaning to Python)?

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -2400,6 +2400,23 @@ should never be auto-downloaded.
 
 The @code{default} sub-directive marks this as the ``default'' commodity.
 
+@item debug
+@findex debug
+@c instance_t::debug_directive in textual.cc
+Prints to stderr the value of a value expression.  For example:
+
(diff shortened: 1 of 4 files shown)


jwiegley:
If the intention is to just give information messages to the console during a run, and you don't want to bother with Python, then yes, I'd call it print.

cpitclaudel:
Is this better than just `eval print(…)`?

jjlee:
I just got around to trying that :-) -- yes, I think eval print is indeed just as good as this debug command was.

Here's an example for anybody hunting for how to print ledger expression values:

    eval print("Foo", account("Assets:Foo").total)

For example you can then run

    ledger my_ledger.ledger b Assets:Foo

and see output like this:

    Foo $123.00

For some reason I don't understand, with my ledger file at least, the following does NOT print anything:

    ledger my_ledger.ledger b Assets

PR #498, opened 2017-11-02
Allow printing the depth of the first displayed parent

JDemler:
For balance reports ledger uses the `depth_spacer` function.
Alas, when creating reports for LaTeX spaces as indentation don't work.

This commit introduces the `depth_parent` command that returns the indentation level equal to the one used in native balance reports.

The command `depth` returns the depth of the current account, which results in unexpected indentation when some accounts collapse.

### DIFF
--- a/src/account.cc
+++ b/src/account.cc
@@ -309,6 +309,20 @@ namespace {
     return long(&account);
   }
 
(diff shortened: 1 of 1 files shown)


jwiegley:
Nice!

PR #503, opened 2017-12-29
Use an std::set instead of an std::list to store the the journal file paths

ghost91-:
As it does not make sense to use a certain journal file several times we should use a set instead of a list to store the journal file paths in order to ensure that the paths added are actually unique. Unfortunately, this change does not account for the fact that links and relative paths might be used to get different paths that actually point to the same file, but it is a first step in the right direction.

Please don't pull this yet, as its is a work in progress

### DIFF
--- a/src/session.h
+++ b/src/session.h
@@ -52,6 +52,16 @@ namespace ledger {
 
 class xact_t;
 
+struct ComparePaths
+{
+  bool operator()(const path& p1, const path& p2) const
+  {
+    return p1 < p2 && !boost::filesystem::equivalent(p1, p2);
+  }
+};
+
+#define COMMA ,
+
 class session_t : public symbol_scope_t
 {
   friend void set_session_context(session_t * session);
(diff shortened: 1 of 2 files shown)


src/session.h
```diff
@@ -52,6 +52,16 @@ namespace ledger {
...
 
+struct ComparePaths
+{
+  bool operator()(const path& p1, const path& p2) const
+  {
+    return p1 < p2 && !boost::filesystem::equivalent(p1, p2);
+  }
+};
+
+#define COMMA ,
```
jwiegley: What is the purpose of this macro definition?

ghost91-: The definition of `data_files` is done via the macro `OPTION__`. In order to provide an own comparison fuction for `std::set<path>`, we need to provide it as a template parameter which would mean writing `std::set<path, ComparePaths>`. But because this is all part of a macro parameter, the "," gets interpreted by the precompiler and we get an error (something along the lines of "macro `OPTION__` only allows 3 parameters, not 4"). There are several ways to work around this, one is defining a macro which inserts the "," (which I did). Other solutions involve modifying the macro being used (`OPTION__`) and in order to keep things simple, I wanted to avoid that. I just realized that boost already defines such a comma macro in `boost/preprocessor/punctuation/comma.hpp` (`BOOST_PP_COMMA()`) so I guess using that one would be the appropriate way.

PR #508, opened 2018-01-04
Fix segfault when using --market with --group-by

mbudde:
`changed_value_posts::create_accounts()` reuses the `<Revalued>` account
from `display_filter`, but when clearing `changed_value_posts`
`create_accounts()` would be called before the account had been
recreated by `display_filter_posts`. This results in a segfault when
using the --group-by option.

I'm not sure if `display_filter_posts` has the same problem but I
reordered the calls there too for good measure.

### DIFF
--- a/src/filters.h
+++ b/src/filters.h
@@ -565,9 +565,9 @@ public:
     last_display_total = value_t();
 
     temps.clear();
-    create_accounts();
-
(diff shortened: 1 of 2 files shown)


jwiegley:
Good catch!

PR #519, opened 2018-01-25
Add support for --prepend-format in accounts command

mbudde:
Add support for the `--prepend-format` and `--prepend-width` options in
the `accounts` command.

### DIFF
--- a/src/output.cc
+++ b/src/output.cc
@@ -282,8 +282,23 @@ void format_accounts::operator()(account_t& account)
 void report_accounts::flush()
 {
   std::ostream& out(report.output_stream);
+  format_t      prepend_format;
+  std::size_t   prepend_width;
+
+  if (report.HANDLED(prepend_format_)) {
+    prepend_format.parse_format(report.HANDLER(prepend_format_).str());
+    prepend_width = report.HANDLED(prepend_width_)
+      ? lexical_cast<std::size_t>(report.HANDLER(prepend_width_).str())
+      : 0;
+  }
 
   foreach (accounts_pair& entry, accounts) {
+    if (prepend_format) {
+      bind_scope_t bound_scope(report, *entry.first);
+      out.width(static_cast<std::streamsize>(prepend_width));
+      out << prepend_format(bound_scope);
+    }
+
     if (report.HANDLED(count))
       out << entry.second << ' ';
     out << *entry.first << '\n';
--- a/test/baseline/opt-prepend-format.test
+++ b/test/baseline/opt-prepend-format.test
@@ -15,3 +15,8 @@ VMMXX07-Feb-02 RD VMMXX              As:Inves:Vanguar:VMMXX  0.350 VMMXX  0.350 
 VMMXX                                In:Divid:Vanguar:VMMXX       $-0.35       $-0.35
                                                                      0.350 VMMXX
 end test
+
+test accounts --prepend-format "%(account_base) "
+VMMXX Assets:Investments:Vanguard:VMMXX
+VMMXX Income:Dividends:Vanguard:VMMXX
+end test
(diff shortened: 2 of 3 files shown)


afh:
Thank you for your contribution @mbudde. Out of curiosity, what is a potential use case for this?

mbudde:
It was requested by a user on the [mailing list](
https://groups.google.com/forum/m/#!topic/ledger-cli/KZtVlaeMRXo), but he
didn't explain his exact use case. My motivation for making the patch was
to get more familiar with the code base :-)

afh:
Thank you for providing more context @mbudde and thanks so much for the contribution. So happy to see people take interest in ledger and providing patches 😃 

In general I'm uncertain whether we really need `--prepend-width=20` since it appears that the same result could be achieved with `--prepend-format='%-20(" ")'`, but I'll leave that for some other time. Especially since the latter is much less intuitive than the first

afh:
Haha, I was just about to merge this PR, @jwiegley 😆

jwiegley:
@afh Great idea! Since you're completed right that the format can have a width specifier, we should deprecate the other option now.

mbudde:
FWIW, I believe that `--prepend-width` is still needed in the balance and register reports because posts with multiple lines only get the string prepended on the first line while `--prepend-width` is used to indent the remaining lines. E.g.
```
2015-01-01 Foo
    AA         5 USD
    AA         5 EUR
    BB
```
```
$ ./ledger --args-only -f test.dat reg --prepend-format '%-5("xxx")' --columns 40
xxx  15-Jan-01 Foo      AA       5 USD  5 USD
xxx                     AA       5 EUR  5 EUR
                                   5 USD
xxx                     BB      -5 EUR  5 USD
xxx                     BB      -5 USD      0
```
```
$ ./ledger --args-only -f test.dat reg --prepend-format '%-5("xxx")' --prepend-width 5 --columns 40
xxx  15-Jan-01 Foo      AA       5 USD  5 USD
xxx                     AA       5 EUR  5 EUR
                                        5 USD
xxx                     BB      -5 EUR  5 USD
xxx                     BB      -5 USD      0
```

afh:
Thank you @mbudde for the excellent example regarding `--prepend-width`. What this reveals is potentially an area for improvement where balance and register reports with multiple lines should also prepend the given format string for all lines and not just for the first one. I might be missing something, but it seems worthwhile to investigate further.

@mbudde interested in having a l:eyes:k?

mbudde:
I'm not sure if "prepend to all lines" is the desired behavior? I guess it depends on what you're trying to do. It also looks like it will be tricky to implement – currently the prepend logic is split between the code and the report formats. The code decides whether to output the `prepend-format` before the post (and uses `prepend_width` to set a minimum width) and `prepend_width` is then used in the balance and register formats to justify subsequent lines. This split means that if you specify `--prepend-width` without `--prepend-format` then first line of a post will not be indented but the subsequent lines will.

afh:
Thank you @mbudde for your comment, it shows a valid use case for `--prepend-width`, yet I think it would be worthwhile to elaborate on the functionality and difference in the docs.

jungle-boogie:
Thank you for this commit. I'm using it to print account numbers for various utilities.

```
$  ledger -f account.ledger accounts --prepend-format '%(note)'
account#: 0000000Expenses:Home:Bill:Vendor
account#: 1111111Expenses:Home:Bill:Vendor2
account#: 2222222Expenses:Home:Bill:Vendor3
account#: 3333333Expenses:Home:Bill:Vendor4
```

Is there a way to have it look like this:
```
account#: 0000000 Expenses:Home:Bill:Vendor
account#: 1111111 Expenses:Home:Bill:Vendor2
account#: 2222222 Expenses:Home:Bill:Vendor3
account#: 3333333 Expenses:Home:Bill:Vendor4
```

jwiegley:
You can always add a ' ' at the end of your format string.

jungle-boogie:
hi @jwiegley, Yep, I figured that out seconds after making my post. Sorry about that.

PR #522, opened 2018-01-26
Add --check-cleared option

pfsmorigo:
Currently if the transaction is cleared, ledger skips the pedantic and
strict checking for it. This flag force the transaction to be checked
even if that's the case.


cpitclaudel:
Is this different from --explicit?

pfsmorigo:
I wasn't aware of --explicit. For my use case it works so I will cancel this request. Thanks for the info!

PR #540, opened 2018-02-19
Fix warnings for -Wimplicit-fallthrough

scfc:
I have not tested these for `clang`, but I think ``[[ noreturn ]]` is standard.

### DIFF
--- a/src/error.h
+++ b/src/error.h
@@ -47,7 +47,7 @@ namespace ledger {
 extern std::ostringstream _desc_buffer;
 
 template <typename T>
-inline void throw_func(const string& message) {
+[[ noreturn ]] inline void throw_func(const string& message) {
   _desc_buffer.clear();
   _desc_buffer.str("");
   throw T(message);
--- a/src/mask.cc
+++ b/src/mask.cc
@@ -75,9 +75,8 @@ mask_t& mask_t::assign_glob(const string& pat)
       if (i + 1 < len) {
         re_pat += pat[++i];
         break;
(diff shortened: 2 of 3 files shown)


jwiegley:
@scfc Is this gcc or clang specific?

scfc:
@jwiegley The warnings are emitted by g++ (GCC) 7.3.1 20180130 (Red Hat 7.3.1-2).

jwiegley:
@scfc Will your change still compile with clang? Should these be #if guarded for g++?

scfc:
@jwiegley I did not test the change with `clang`, but my understanding of http://en.cppreference.com/w/cpp/language/attributes is that all C++ compilers should accept `[[ noreturn ]]`.

nemequ:
`[[noreturn]]` in C++11+. Many compilers support compiler-specific attributes before then; see [`HEDLEY_NO_RETURN`](https://github.com/nemequ/hedley/blob/4b07f1f98b55b3d6624af00fb44b34140b2a29bd/hedley.h#L685).

The fall through comments are a bit more complicated.  Several compilers (and static analysis tools) work with the comments, but the standard (as of C++17+) is `[[fallthrough]]`.  And, of course, lots of compilers have compiler-specific ways of doing it; see [`HEDLEY_FALL_THROUGH`](https://github.com/nemequ/hedley/blob/4b07f1f98b55b3d6624af00fb44b34140b2a29bd/hedley.h#L981)

PR #557, opened 2018-04-05
budget_posts: Keep pending items until the last day they apply

smr894:
@mthurman as myself found out that posts that belong to a budget are missing from the register report on the last period of a budget.
This should fix that and may replace pull request https://github.com/ledger/ledger/pull/537.
@mthurman thanks for spotting the location this saved me a lot of time.


jwiegley:
Thank you!

PR #1651, opened 2018-06-10
Remove TOK_A_YEAR token

### LINKED ISSUE #1626: Error: Unexpected date period token '1000' (BZ#554)

Bugzilla bug ID: BZ#554
From: seanh
CC:   @tbm

Christoph-D:
This fixes #1626.

The tokenizer eagerly classifies 4-digit integers as TOK_A_YEAR
tokens.  In some contexts such as "every 1000 years", this causes
errors.

I think the tokenizer does not have enough information available to
distinguish between integers and years.

After this patch, the tokenizer will always classify integers as
TOK_INT tokens.  The "has 4 digits" heuristic to determine if an
integer is a year is moved to the place where it's actually
needed (and it can be slightly more generic there, too).

### DIFF
--- a/src/times.cc
+++ b/src/times.cc
@@ -420,7 +420,6 @@ class date_parser_t
         TOK_DASH,
         TOK_DOT,
 
-        TOK_A_YEAR,
         TOK_A_MONTH,
         TOK_A_WDAY,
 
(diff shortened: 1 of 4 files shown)


jwiegley:
Nice work!

Issue #1187, opened 2016-10-04
Can not balance when multiple posting with same account is one transaction (BZ#1187)


tbm:
Note: the issue was created automatically with bugzilla2github

Bugzilla bug ID: BZ#1187
From: Thierry (@thdox)

tbm:
Fixed in d541cceaf7d4e77c239849679e3ae7a449ae8c8b

PR #1667, opened 2018-07-17
Improve fix of issue #1187

pascalfleury:
I coded a bit on a better fix for issue 1187. What I did:
- the computation of the diff between what is in the account and the assertion is now a `balance_t` instead of `amount_t`, so it handles multiple currencies.
- it then checks only the the commodity specified in the assertion amount when verifying the assertion
- if the assertion value is 0 without a commodity, it will ensure that the diff is zero for all commodities.

Added a test, updated documentation. Feel free to comment.


jwiegley:
Great work, thanks!

PR #1672, opened 2018-08-02
Add notes to 5.7.1.1 Payee metadata tag doc

scarpent:
Based on my observations. The first behavior makes sense to me. For the second, it would be nice if ledger would enforce `--check-payees` on the tag. In either case I thought it would be helpful to call these out.


scarpent:
I also notice that the payee tag appears to be case insensitive. Is that also worth noting?

scarpent:
Thank you -- good point. I've made an update.

PR #1691, opened 2018-11-18
explain round brackets in section 3.8 on multiple funds

GeorgLink:
Section "Working with multiple funds and accounts" introduces square brackets and explains them but does not explain round brackets in the following example. This commit adds the explanation what the round brackets do.


jwiegley:
Thank you!

PR #1710, opened 2019-01-13
Replace HAVE_UNIX_PIPES with test for _WIN32 and __CYGWIN__

scfc:


tbm:
@scfc do you think you could update the patch as @jwiegley requested?

tbm:
@scfc do you think you could update the patch as @jwiegley requested?

PR #1716, opened 2019-01-13
Ubuntu fixes

pascalfleury:
This contains a few fixes for Ubuntu-based systems:
 - acprep decodes and compares strings (returned from the commands are bytes), which also works with Python3
 - updated the deps for bionic (18.04) which is the current LTS (Long Term Support) version
 - adjusted README.md to reflect these newer systems. Also removed instruction for 12.04 (which has not gotten any update since April 28, 2017)

### DIFF
--- a/acprep
+++ b/acprep
@@ -510,7 +510,7 @@ class PrepareBuild(CommandLineApp):
 
         self.log.info("Installing Ledger's build dependencies ...")
 
-        system = self.get_stdout('uname', '-s')
+        system = str(self.get_stdout('uname', '-s'))
 
         if system == 'Darwin':
             if exists('/opt/local/bin/port'):
--- a/acprep
+++ b/acprep
@@ -62,17 +62,7 @@ class BoostInfo(object):
         if system == 'centos':
             return [ 'boost-devel' ]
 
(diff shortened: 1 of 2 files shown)


acprep
```diff
@@ -510,7 +500,7 @@ class PrepareBuild(CommandLineApp):
 
         self.log.info("Installing Ledger's build dependencies ...")
 
-        system = str(self.get_stdout('uname', '-s'))
+        system = self.get_stdout('uname', '-s').decode('utf8')
```
tbm: It seems you're undoing the change from the other commit there?

pascalfleury: yes that was a mistake, and I did not clean the change history.

pascalfleury:
I am a bit lost as to how I can a) bring my fork up-to-date, and b) that the pull from ledger/ledger does not show up as a commit in my pull request for the deps changes in acprep.

It seems that syncing both github repos leavves too many traces...

pascalfleury:
Yes, had no time to act on it yet.

I did not find a way to not have some of the commits show up in the PR, so
I created a new one (#1733).
Hopefully that one is clean and working.

Pascal

PR #1717, opened 2019-01-13
Suppress fallthrough warnings

pascalfleury:
My compiler has new warnings switched on by default, so that now it is visible that the compiler does not understand that the macro+template combo about throw_ is actually unconditionally throwing an exception (and exiting the function).

This change is about adding 
 - 'break' statements where needed to help the compiler 
 - add "// fall through" comments where fall through is wanted, also to make the compiler not warn about it.

### DIFF
--- a/src/mask.cc
+++ b/src/mask.cc
@@ -75,9 +75,8 @@ mask_t& mask_t::assign_glob(const string& pat)
       if (i + 1 < len) {
         re_pat += pat[++i];
         break;
-      } else {
-        // fallthrough...
       }
+      // fall through...
     default:
       re_pat += pat[i];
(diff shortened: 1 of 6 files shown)


tbm:
Thanks, I noticed these as well the other day. I think instead of all the `break` statements after `throw`, it would be better to annotate the `throw` macro so the compiler knows it doesn't return.  It looks like this is possible when I looked at the documentation briefly, but I didn't investigate much. Maybe you know how to or can check.

btw, your PR contains a merge commit  that shouldn't be there.

tbm:
I just noticed there's already a pull request open about this. See #540

PR #1726, opened 2019-01-16
Compile strptime.cc only on Windows

scfc:


scfc:
(I have not tested this on Windows,)

PR #1730, opened 2019-01-16
Remove unnecessary include for boost/filesystem/path.hpp

scfc:


tbm:
Many of your pull requests should really be one (some with several commits, some with just one).

The last 4 prs should definitely be one PR, maybe even one commit.

scfc:
No problem; I just test these as single units to ensure that they work on their own (and can be discussed if necessary) and so publish them this way so that a simple change is not blocked by something controversial.

I'll compose a combined pull request once I'm done with my current batch and then publish that.

PR #1733, opened 2019-01-17
Ubuntu deps update

pascalfleury:
This contains a few fixes for Ubuntu-based systems:

acprep decodes and compares strings (returned from the commands are bytes), which also works with Python3
updated the deps for bionic (18.04) which is the current LTS (Long Term Support) version
adjusted README.md to reflect these newer systems. Also removed instruction for 12.04 (which has not gotten any update since April 28, 2017)

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,21 +114,13 @@ run:
 ### Ubuntu
 
 If you're going to build on Ubuntu, `sudo apt-get install ...` the
-following packages (current as of Ubuntu 14.04):
+following packages (current as of Ubuntu 18.04):
 
     $ sudo apt-get install build-essential cmake doxygen \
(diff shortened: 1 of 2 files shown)


pascalfleury:
It could, if the returned bytes from the command are non-UTF8. And currently, UTF-8 is the most used encoding on systems, if it is non-ASCII. And ASCII is a subset of UTF-8.
Or is your suggestion to use plain ASCII, so that an error message output at that stage would simply fail the conversion, which will display it on the screen as an error message itself ? It would possibly work for some errors, not for others, and wold make things more confusing.

Note that this is also a detail, as this script is used by people who know how to handle a console output.

PR #1737, opened 2019-01-19
Use standard GCC in Travis CI

scfc:
Commit 4c4367fe6b7f184605c900735fc5b646f45311c1 added some logic to
compile Ledger with GCC 4.8 as Travis CI's Ubuntu Precise environments
only offered 4.6 at that time.  Since then, the default image has
changed to Ubuntu Trusty which provides GCC 4.8.


tbm:
Thank you!

PR #1750, opened 2019-01-26
Fix possible stack overflow in date parsing routine

### LINKED ISSUE #1224: stack-buffer-overflow in ledger::parse_date_mask_routine (BZ#1224) (CVE-2017-12482)

Bugzilla bug ID: BZ#1224
From: @gy741

mbudde:
It is possible to create a stack overflow by giving a date that is
longer than the buffer that is used during date parsing because the
length of the input string is not checked. The `VERIFY` macro is only
enabled when debug-mode is enabled and the `--verify-memory` argument is
used.

Prevent the issue by always checking the input string length and
discarding dates that does not fit in the buffer as invalid.

This issue has been assigned CVE-2017-12482.

Fixes #1224


tbm:
Thank you, @mbudde

PR #1756, opened 2019-01-30
Ensure the org report always lists commodities with the same ordering

### LINKED ISSUE #1706: baseline/cmd-org.test fails
I see:

```
/usr/bin/python2.7 "/home/tbm/ledger/test/RegressTests.py" "/home/tbm/ledger/ledger" "/home/tbm/ledger" "/home/tbm/ledger/test/baseline/cmd-org.test"

FAILURE in output from /home/tbm/ledger/test/baseline/cmd-org.test:
--
$ledger -f "/home/tbm/ledger/test/baseline/cmd-org.test" org
--
  @@ -5,8 +5,6 @@
   |||Test 3|*|D|-20.00 EUR|0.00 EUR| Payee: Test 3
   |12-Mar-25|99|Test 4|*|E|30.00 GBP|30.00 GBP|
   ||||*|F|-30.00 GBP|0.00 GBP|
  -|12-Mar-26|test|Test 5|*|G|1 AAA|0.00 GBP|
  -|||||||1 AAA|
  -||||*|H|$-10|0.00 GBP|
  +|12-Mar-26|test|Test 5|*|G|1 AAA|1 AAA|
  +||||*|H|$-10|1 AAA|
   |||||||$-10|
  -|||||||1 AAA|
E[cmd-org.test]STDERR:
```

jwiegley:
Fixes #1706

### DIFF
--- a/src/org.cc
+++ b/src/org.cc
@@ -137,49 +137,42 @@ void posts_to_org_table::operator()(post_t& post)
 
     if (amt.type() == value_t::BALANCE || tot.type() == value_t::BALANCE) {
       balance_t amt_bal(amt.to_balance());
+      balance_t::amounts_array amt_sorted;
+      amt_bal.sorted_amounts(amt_sorted);
+
       balance_t tot_bal(tot.to_balance());
-      balance_t::amounts_map::const_iterator i = amt_bal.amounts.begin();
-      balance_t::amounts_map::const_iterator j = tot_bal.amounts.begin();
+      balance_t::amounts_array tot_sorted;
+      tot_bal.sorted_amounts(tot_sorted);
+
+      auto i = amt_sorted.begin();
+      auto j = tot_sorted.begin();
       bool first = true;
-      while (i != amt_bal.amounts.end() || j != tot_bal.amounts.end()) {
+
+      while (i != amt_sorted.end() || j != tot_sorted.end()) {
         if (first) {
           first = false;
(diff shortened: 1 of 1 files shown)


jwiegley:
This may not be entirely correct yet. We now stabilize the ordering based on the results of calling `name()` for each commodity, but this may only be the base name, not the entire name as it might appear when using `--lots`. In that case, perhaps we need to fully enable lot reporting locally before asking for the name.

tbm:
@mbudde can you review this change?

jwiegley:
@tbm @mbudde Reading this code more carefully, the report is a hack. We're using the main logic to emit the "first line of a posting" (which works if it's a single amount), but then we're trying to take the balance, sort it the way the report would have sorted it, skip the first line, and then emit what comes after.

The code is extremely fragile! The sorting in the main logic is not the same as the sorting being done here.

I don't know the best way to fix this right now, and so can only recommend we disable this report until a better implementation appears. The formatter provided by `format.cc` needs to accept a template for what to do in the case of subsequent lines in multi-commodity balances.

That way that it's done for regular register reports right now is also hack: we use this `justify` function to tell the amount printer how much whitespace it should create to the left of each subsequent line in the balance report. This is terrible! It should be handled by a formatting line specifically for these other lines.

tbm:
@jwiegley removed the `org` command in f0dacdb22f2888 so I'm closing this PR.

PR #1766, opened 2019-02-16
Fix tag value parsing

### LINKED ISSUE #1702: Tag value parsing includes extra characters (": ")
Anywhere that I try to match the value of a tag, if I try matching e.g. `Value`, it doesn't work, but if I instead match `: Value` (that's a colon and a space prepended to the value), it does.  An [MCVE](https://stackoverflow.com/help/mcve):

For ledgerfile:

```
tag Foo
    assert value =~ /^: Bar$/

2019/01/01 * Payee
    ;; Foo: Bar
    Income:Foo    $-1
    Assets:Cash    $1
2019/01/01 * Another Payee
    Assets:Cash   $-1
    Expenses:Baz   $1
```

running `ledger bal` works.  If line 2 is changed to `assert value =~ /^Bar$/`, it gives the error:

```
While parsing file "test.ledger", line 7:
Error: Metadata assertion failed for (Foo: : Bar): (value =~ /^Bar$/)
```

The behavior is the same for `--limit` args, as well as when using direct comparison instead of a regex.  I.e.:

```
ledger bal --limit 'tag("Foo") == ": Bar"'
```

shows the correct results, while:

```
ledger bal --limit 'tag("Foo") == "Bar"'
```

returns nothing.

I tested this on 863cb3b, which is the latest `next` at present.

mbudde:
If a tag is more than 2 characters from the beginning of the comment the
tag value offset will be wrong. #1702 gives an example where the tag
line starts with `;;` and the tag value thus becomes `: Bar` because of
this bug.

The use `index` in the offset calulation seems to be a lucky coincidence
that works in the common case: "; tag: value"

Fixes #1702

### DIFF
--- a/doc/NEWS
+++ b/doc/NEWS
@@ -4,6 +4,8 @@
 
 - Properly reject postings with a comment right after the flag (bug #1753)
 
+- Fix bug in tag value parsing (bug #1702)
+
 * 3.1.2 (2019-02-05)
 
 - Increase maximum length for regex from 255 to 4095 (bug #981)
--- a/src/item.cc
+++ b/src/item.cc
@@ -198,7 +198,7 @@ void item_t::parse_tags(const char * p,
       tag = string(q, len - index);
 
       string_map::iterator i;
-      string field(p + len + index);
+      string field(p + (q - buf.get()) + len);
       trim(field);
       if (by_value) {
         bind_scope_t bound_scope(scope, *this);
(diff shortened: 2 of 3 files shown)


tbm:
This doesn't apply anymore. Can you update it to the current master branch.

mbudde:
Rebased on top of `next`.

PR #1769, opened 2019-03-05
Provide Docker information in README

alberto56:


crazyh:
Not sure that this image is usable for all: it contains hard coded timezone.

alberto56:
Hi @crazyh -- I'd be interested in obtaining more information about how ledger uses timezone information; I have also opened https://github.com/dcycle/docker-ledger/issues/2 to track this within the Docker project itself.

crazyh:
I do not know how ledger uses timezone for calculations but it uses `strftime` so it is  probably that some operations like pricedb-related should be affected by TZ.

alberto56:
Have just merged in a pull request which no longer uses a hard-coded timezone, see https://github.com/dcycle/docker-ledger/issues/2

PR #1790, opened 2019-04-21
alpine docker image

### LINKED ISSUE #1788: Alpine Distribution / Docker Image
Hi!

Reducing the [Docker image](https://github.com/dcycle/docker-ledger/issues/4) size would be a fantastic way to run the same version of Ledger on any platform that supports Docker (i.e. Linux and Windows).

I see that @rgdoliveira has created some [Alpine packages](https://pkgs.alpinelinux.org/packages?name=ledger&branch=edge), which is great. Are or will these be considered official packages, listed on the Download page? What is the current status of these images? Are they reliable? Will they be built regularly? (ok, not that there has been many releases recently but still).

Building a microcontainer based on Alpine Linux (3MB) with Ledger binary package (3.73 MB) should result in an incredibly small portable distribution. This would be my personal preference compared to Windows and OpenSuse distributions, for example. OpenSuse package is still at 3.1.1, for example so I'm using a custom-built one.

I've tried building a microimage locally with 
```
# Docker microimage based on Alpine Linux.
FROM alpine
RUN apk update && apk upgrade
RUN apk add ledger
```

but it is failing with:

image

MisterY:
Link to a Docker image based on Alpine linux and binary distribution.
Resolves #1788

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ If you have Docker installed on your computer or server, you can use a [Docker v
 
     $ docker run --rm -v "$PWD"/test/input:/data dcycle/ledger:1 -f /data/sample.dat reg
 
+A smaller image, based on Alpine Linux and weighing only 23MB, is available [here](https://hub.docker.com/r/cick0/ledger-docker-alpine). This is based on the binary packages available for the Alpine Linux distribution, which are currently in testing phase.
+
 ## Dependencies
 
 If you wish to proceed in this venture, you'll need a few dependencies.  The


jwiegley:
Has anyone else given this a try and is willing to vouch for it?

crazyh:
Isn't it a bad practice to suggest to the end user to use a unofficial Docker image which employs an unofficial non-HTTPS repository (nl.alpinelinux.org is not listed on https://mirrors.alpinelinux.org/)?

There are a lot of other Ledger[-cli] images on the Docker Hub: https://hub.docker.com/search?q=ledger. It may be better just to mention https://hub.docker.com in README.md...

MisterY:
No worries. If you don't find it useful, I'll remove it tomorrow morning.

MisterY:
@crazyh , if you bothered to check the reference docs, you would have found the URLs on this page: https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Packages_and_Repositories

jwiegley:
I'm not sure if you meant it, @MisterY, but your message could come across as dismissive of this user's efforts, whether or not they researched the documentation beforehand. I'm grateful for the effort, even if a solution already exists.

MisterY:
If by "solution" in the previous sentence you mean the image which is linked on the ReadMe page, or others that 10 minutes of googling provides, I would disagree. All the available ones are >500MB downloads. To me that was unacceptable. Then again, it might not be a problem for others. Thanks for the feedback.

jwiegley:
Heh, I got mixed up. I meant that I was grateful *your* efforts, @MisterY!

MisterY:
Thank you, both. The thing is that I used up my spare time over the long weekend and won't be able to contribute significantly in the foreseeable future. Hence, I cancelled the pull request to skip the need to explain what was already described in detail in the related issue (#1788) or to make modifications to the image, etc.
I'm in contact with the author of the currently linked Docker image and the author of the Ledger binaries in the Alpine Linux distribution (he's also a member of the core team), to get this image included in that repository.
As for the ReadMe, anyone can add any link at any time, as they see fit.
Naturally, I also want to thank you for an excellent piece of software, that is Ledger.

jwiegley:
Offering your weekend time is a precious gift indeed.

crazyh:
> @crazyh , if you bothered to check the reference docs, you would have found the URLs on this page: https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Packages_and_Repositories

Yeah, I saw examples in wiki but I am pretty sure that https://mirrors.alpinelinux.org/ must be considered at the first place. A regular user will trust README.md blindly so the reference to the untrusted Dockerfile which employs an unofficial repo is a potential attack vector.

MisterY:
@crazyh,

I understand your concern (I'm also a CISSP) but it may be an overreaction in this case. Calling a repository, hosted on alpinelinux.org domain and listed in the official guide, "unofficial and untrusted" might not be the best choice of words. The ledger package is created by a member of the Alpine core team. 
I'm almost certain that replacing the "nl." domain with any other "official" one would work just the same, as they are all probably mirrored, but unfortunately do not have time or the environment to play around with that and package another image for at least another week or so.

The million dollar question is, even if there was one, what would such an attack vector target? Running in an isolated container with the only entry point being the ledger binary, it could lie to you about your net worth but hardly much else.
The "ordinary users", who build from the source, normally do not go through ledger's source code either.

crazyh:
@MisterY I believe that it is a wrong place to discuss Docker security. I have just pointed that your container does not follow the best practices and it is strange that you are still insisting that it is safe.

> The million dollar question is, even if there was one, what would such an attack vector target?

Spoof DNS and replace APK.

PR #1805, opened 2019-07-06
Look up annotated amounts within balances by their displayed name

jwiegley:


jwiegley:
Note that this change slows down Ledger by about 30% on files that make heavy use of annotations, but it also means that instead of seeing this:

```
Assets:Brokerage       AAPL -50.000 {$190.709} [18-Nov-20]  $-15,799.11
      AAPL -0.366 {$108.98907103825} [14-Nov-13]
       AAPL 0.366 {$108.98907103825} [14-Nov-13]
         AAPL 35.000 {$144.88171429} [19-Jan-07]
         AAPL 35.000 {$159.73371429} [18-Dec-24]
         AAPL 60.000 {$174.32916667} [18-Nov-28]
        AAPL -31.000 {$174.32916667} [18-Nov-28]
       AAPL 0.741 {$187.84075573549} [19-May-16]
```

You see this:

```
Assets:Brokerage       AAPL -50.000 {$190.709} [18-Nov-20]  $-15,799.11
         AAPL 35.000 {$144.88171429} [19-Jan-07]
         AAPL 35.000 {$159.73371429} [18-Dec-24]
         AAPL 29.000 {$174.32916667} [18-Nov-28]
       AAPL 0.741 {$187.84075573549} [19-May-16]
```

jwiegley:
I don't quite understand your question @drewr, can you elaborate?

tbm:
Is there a bug report about this issue?

Is there a test case you can add?

I don't think I encountered this issue before, but the new `--lots` output you posted definitely looks better than the old one.

jwiegley:
No bug report, just something I noticed while entering all my brokerage data into Ledger and looking at the report. It's important that when I ask for the final balance, I only see securities that are currently in my account. This is how I know I've dealt with the lots and their adjusted basis costs correctly.

tbm:
@jwiegley do you have a testcase?

Anyway, I can't comment on the code change but the behaviour certainly looks right to me but I'd be curious to see a testcase since I never saw this.

jwiegley:
Sure, that's a completely valid request. I won't merge without a test case.

The basic problem that this addresses is that we compare commodities by pointer, not by "appearance", so two commodity may have the same printed appearance, but not be on-the-nose equal under the hood. Such commodities should be merged in the final output.

This requires, however, that we render annotated commodities into strings before comparing them, which adds a lot of CPU and memory cost, compared to a simple pointer comparison.

Another solution would be to hash commodities upon creation, so that we can still be comparing by a simple hash value. I may examine this possibility too.

tbm:
@jwiegley I'd be curious to see a test case for this to understand better when this is a problem.

tbm:
It looks like this has been committed.

@jwiegley do you have a testcase that can be added?

jwiegley:
No, I forgot about your request for a test case here! I'll leave this open.

tbm:
Closing this.  It would still be nice to have a test case but John is busy.

Issue #1822, opened 2019-09-04
Undefined reference to `clock_gettime` and `boost::unit_test::unit_test_main(bool (*)(), int, char**)`


twrightsman:
I'm attempting to package ledger (https://github.com/conda-forge/staged-recipes/pull/9410) with [conda](https://conda.io) into [conda-forge](https://conda-forge.org) to make it easier for users to install ledger without needing to build from source.

Unfortunately I'm running into some build issues with the current head of master in the [CI pipeline](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=69006&view=logs&jobId=10bda42a-d420-5c28-a43f-bc4e66405e5c&taskId=bef775d0-a1a2-55a3-9721-17147029baf0&lineStart=2893&lineEnd=2897&colStart=1&colEnd=43). The MacOS build is also failing for different reasons but if you're not interested in supporting MacOS then I can disable that build and just keep the Linux one.

Here is the relevant error from the build pipeline:
```
cc1plus: warning: unrecognized command line option '-Wno-unused-comparison'
[ 97%] Linking CXX executable ../../MathTests
/home/conda/feedstock_root/build_artifacts/ledger_1567553150239/_build_env/bin/../lib/gcc/x86_64-conda_cos6-linux-gnu/7.3.0/../../../../x86_64-conda_cos6-linux-gnu/bin/ld: CMakeFiles/MathTests.dir/t_amount.cc.o: in function `main':
t_amount.cc:(.text.startup.main+0xe): undefined reference to `boost::unit_test::unit_test_main(bool (*)(), int, char**)'
/home/conda/feedstock_root/build_artifacts/ledger_1567553150239/_build_env/bin/../lib/gcc/x86_64-conda_cos6-linux-gnu/7.3.0/../../../../x86_64-conda_cos6-linux-gnu/bin/ld: ../../libledger.so.3: undefined reference to `clock_gettime'
collect2: error: ld returned 1 exit status
make[2]: *** [test/unit/CMakeFiles/MathTests.dir/build.make:156: MathTests] Error 1
make[1]: *** [CMakeFiles/Makefile2:304: test/unit/CMakeFiles/MathTests.dir/all] Error 2
make: *** [Makefile:163: all] Error 2
```

twrightsman:
Building locally with the dependencies gives similar errors:
```
[ 90%] Linking CXX executable ../../UtilTests
[...]

PR #1823, opened 2019-09-04
Fix build issues with boost 1.70

twrightsman:
Fixes https://github.com/ledger/ledger/issues/1822 when building with boost 1.70.

Credit: https://stackoverflow.com/questions/14715837/can-i-link-multiple-boost-unit-tests-into-a-single-test-binary

Could we possibly create a release or pre-release to include these changes? conda-forge prefers to build only on full releases instead of master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,17 +21,15 @@ env:
   matrix:
     # Boost version to build against; an empty string means the
     # distribution's default.
-    - BOOST_VERSION=""
-    - BOOST_VERSION="1.61.0"
+    - BOOST_VERSION="" CMAKE_VERSION=""
+    - BOOST_VERSION="1.70.0" CMAKE_VERSION="3.15.3"
 
 # The configuration for macOS only works with Boost installed by
 # homebrew, so exclude the other combinations.
 matrix:
   exclude:
-    - os: linux
-      env: BOOST_VERSION=""
     - os: osx
-      env: BOOST_VERSION="1.61.0"
+      env: BOOST_VERSION="1.70.0" CMAKE_VERSION="3.15.3"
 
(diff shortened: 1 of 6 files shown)


twrightsman:
It appears the Linux build has succeeded. @jwiegley Do you want to try building and testing with Boost 1.70 on Travis instead of 1.61?

twrightsman:
Well that's weird, it's [not linking](https://travis-ci.org/ledger/ledger/jobs/580941744#L956) `ledger` with Boost.

jwiegley:
@twrightsman Travis is saying the Linux build failed...

twrightsman:
@jwiegley I still need to look into why Boost isn't being included when building ledger on Travis but is on the conda-forge CI and my local machine.

twrightsman:
@tbm I'm still not sure this is the proper fix, just want to quickly test it with Travis

twrightsman:
@jwiegley Sorry, I didn't see your review before pushing a new commit, could you repost it?

twrightsman:
I have a suspicion that an outdated Travis CMake is the reason we are unable to build with Boost 1.70+ on both Mac and Linux. Other ideas? Could we try and update the Travis CI images to newer OS versions?

twrightsman:
Personally, I'd say we only merge in the first commit, which doesn't break the current Travis CI Linux builds but allows newer CMakes (e.g. on conda-forge) to build ledger with Boost 1.70.

xnox:
Current master builds against 1.71.0 on Mac OS X and passes all tests (sans python demo, which is fixed in a recent PR from me)

So this is obsolete / not needed anymore, and can be closed?

PR #1825, opened 2019-09-06
Add conda installation instructions to README

twrightsman:
This will add the instructions to install the latest full release (including the build fix patches for Boost 1.70) with conda to the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,13 @@ run:
          boost +st+python26+icu texlive doxygen graphviz \
          texinfo lcov sloccount
 
+### Conda
+
+Ledger is also available through [Conda](https://conda.io) from the
+[conda-forge](https://conda-forge.org) channel:
+
+    $ conda install -c conda-forge ledger
+
 ### Ubuntu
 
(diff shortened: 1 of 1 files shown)


jwiegley:
Why would I include instructions for `conda`? I haven't even heard of it yet.

twrightsman:
Conda is a package manager that is typically used to install packages into a user's home directory, so that root access is unneeded and the system environment doesn't need to be modified or filled with development libraries.
It's quite similar to homebrew, but for Linux, so I figured instructions for it would be useful to more casual Linux users, like how the homebrew instructions are in the README. Would you prefer the instructions further down near homebrew? Or would you rather not include this?

jwiegley:
I'm happy to include it, based on your description; can you move it down near homebrew? That way we have the Mac and Linux "quick start" options near each other. Thanks!

twrightsman:
That's great, I hope it will help adoption by more casual Linux users!
How does the current positioning look?

jwiegley:
Great, thanks!

Issue #545, opened 2018-02-22
Unknown account warning with --strict when using $account


rjurga:
I reported this some months ago on Google Groups, but I believe I should put it here if Github is used to track issues.

I get a warning that I think shouldn't be there when I use the --strict option.

> Warning: "Budget.dat", line 16: Unknown account 'Budget:$account'

Here are the relevant lines in Budget.dat:

> = /^Expenses:(?=Groceries|Rent|Utilities)/
>     (Budget:$account)    -1.0

When the referenced $account is parsed during execution, Budget:$account always yields accounts that are already known. So in my opinion there should be no warning and I would consider this a bug.

A workaround is to add the following at the top of the Budget.dat file:

> account Budget:$account

tbm:
96eb5fe19fcbe46ca should be reverted when this is fixed.

PR #1834, opened 2019-11-15
Add note regarding $account under --strict or --pedantic modes

elamperti:
This change adds a warning in the documentation for users using automated transactions with `--strict` or `--pedantic`, as ledger will throw an error when `$account` is not defined explicitly (see #545).


jwiegley:
Thank you!

Issue #1693, opened 2018-12-11
--pivot doesn't respect --date-format


tbm:
```
2018-12-11 * Test for date-format
    Assets:Card:Tesco       10.00 GBP
        ; Card: 123
        ; Expiration:: [2019-12-11]
    Assets:Cash            -10.00 GBP
```

```
ledger -f d --date-format %Y-%m-%d bal card --pivot Expiration 
           10.00 GBP  Expiration:2019/12/11:Assets:Card:Tesco
```

tbm:
Related to #1202

tbm:
@jwiegley I need your input on this. ledger has `FMT_WRITTEN` and `FMT_PRINTED`. I think the idea was that if a ledger file is produced that is likely written down, ledger would always use a consistent date format (rather then the user-defined one).  Whereas `FMT_PRINTED` leads to the date format specified by the user.

However, users complained about the output not respecting their choice and e.g. in 0c9e188 we changed it for lot dates.

This bug and #1202 can be fixed by changing some `FMT_WRITTEN` to `FMT_PRINTED` in `src/value.cc`.  But basically if we go ahead with this change, I think `FMT_WRITTEN` becomes mostly obsolete (unless I'm missing some other case where this is being used).

Any opinion on this @jwiegley?

jwiegley:
@tbm I'm fine with getting rid of `FMT_WRITTEN` based on your analysis.

jwiegley:
Fixed by PR #2773. The --pivot option correctly respects --date-format in the current codebase. A regression test was added in test/regress/1693.test to confirm this behavior.

PR #1836, opened 2019-11-25
Format annotations using format that can be parsed

dmage:
I expect an output of `ledger print` to be consumable by ledger.

But on the next journal

```
2019/11/25 * test
    Foo    1 AAPL {1.00 EUR} [2019/11/24]
    Bar
```

it prints [19-Nov-24], which it does not understand with default options.

With this patch it prints [2019/11/24].


jwiegley:
Thank you!

tbm:
for the record, this relates to issue #1693 which I never got around to.

Also, this was merged on next which is no longer used. I pushed it to master.

PR #1840, opened 2019-12-05
travis: use the same python executable, as found by cmake

xnox:


xnox:
@jwiegley please merge this, this is an improvement as MacOSX PY2 build now passes again.

/me ponders if there is a way to mark the PY3 builds as "expected failure" in travis.

jwiegley:
Yes, there's a way. :)

PR #1848, opened 2019-12-26
Remove reference to lisp from default.nix

purcell:
There is no longer a submodule which provides elisp, so the `cp` portion of this expression fails.

Also remove obsolete comment about presence of gitmodules.

Background to this is that we're using nix to obtain `ledger` for the `ledger-mode` Travis builds, and building from this expression [fails in CI](https://travis-ci.org/ledger/ledger-mode/jobs/594317052).

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -9,20 +9,11 @@ in
 stdenv.mkDerivation {
   name = "ledger-${version}-${rev}";
 
-  # NOTE: fetchgit because ledger has submodules not included in the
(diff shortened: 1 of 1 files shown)


purcell:
Thanks Martin. I'm a committer on this repo but preferred that someone else glance over the commit because I'm not really sure how (or if) `default.nix` is being used by anyone. It seemed pretty broken due to this issue afaict.

PR #1851, opened 2020-01-08
fix #1850: segfault when using ledger print

### LINKED ISSUE #1850: Segfault when using ledger print: std::__cxx11::basic_string / boost::u32regex_search
ledger print -f test.txt gives a segfault. test.txt:
```
= expr payee =~ /Max/ & account =~ /Expenses:Unknown/
	Expenses:Rent  1
	Expenses:Unknown  -1

account Expenses:Car
	payee SB Tank


2019/12/30 * Max Mustermann  ; Rent
    Expenses:Unknown                        650,00
    Assets:Checking                             -650,00 €
```
Note that there is no segfault if:
  * the regex does not match the payee
  * Expenses:Unknown is not used in the automatic transaction
  * The account declaration for Expenses:Car is removed

Ledger 3.1.3-20190331, the command-line accounting tool

Panke:
This solution does not break any tests, but seems wrong to me: what are the actual invariants w.r.t posts and their associated accounts? What is an xact_t anyway? 

However this is good enough as a workaround for me. Maybe someone else finds it useful as well.

### DIFF
--- a/src/journal.cc
+++ b/src/journal.cc
@@ -136,7 +136,7 @@ account_t * journal_t::register_account(const string& name, post_t * post,
   // the payee indicates an account that should be used.
   if (result->name == _("Unknown")) {
     foreach (account_mapping_t& value, payees_for_unknown_accounts) {
-      if (post && value.first.match(post->xact->payee)) {
+      if (post && post->xact && value.first.match(post->xact->payee)) {
         result = value.second;
         break;
       }


jwiegley:
An `xact_t` is a transaction, which must have zero or more balancing postings (unless some are of the non-balancing virtual kind). Every valid posting must have a corresponding transaction, so if `post->xact` is NULL, it is an orphaned posting, probably created as a temporary for some reason.

PR #1856, opened 2020-01-21
Allow install dev-dependencies on ubuntu-eoan

daraul:
I wasn't able to install dependencies on my pop-os laptop easily, until I made these changes. I am not sure if these changes involve any pipelines or not, so I didn't do `ci.skip`.


jwiegley:
Thank you!

PR #1861, opened 2020-02-10
Fix Buffer Overflow when Evaluating Pricing Date

s-rah:
There is the potential for a buffer overflow in times.cc when parsing the date portion of pricing history entires. glibc fortifications prevent this from being a major security concern (but without them this would cause stack corruption).

(Apologies, the sample test case is a little ridiculous because it is a fuzzing output)


tbm:
> (Apologies, the sample test case is a little ridiculous because it is a fuzzing output)

Sorry for replying to late.  I wanted to see if I could find an ASCII test case instead but never found time.

Now that I tried the patch, I actually can't reproduce the issue.  I get:

```
Error: Invalid date/time: 6
```

followed by some binary.

I wonder if this got fixed by:

commit 7c0ae5b02571e21f97d45f5d091cb78af9885713
Author: Michael Budde <mbudde@gmail.com>
Date:   Sat Jan 26 09:30:35 2019 +0100

    Fix possible stack overflow in date parsing routine

which adds the same check in a different time function.

But in any case, I agree the check should be added.  I'll go ahead and commit your patch but maybe without the test case.  Thanks!

tbm:
Merged as commit 466ed19374903299a3ad21213d3a91e18559f9db

Thank you!

PR #1875, opened 2020-03-26
py3: ensure python output is not lost

q3aiml:
Unbuffer python's stdio to avoid output getting stuck in buffer when
stdout is not a TTY. Normally buffers are flushed by `Py_Finalize` but
Boost has a long-standing issue preventing the proper shutdown of the
interpreter with `Py_Finalize` when embedded [1].

This applies the same fix as 139beba but to any ledger usage rather than
only the test suite. I removed `PYTHONUNBUFFERED=1` from tests as there
is no expectation that users should need to have this set for ledger to
function.

For example without this fix piping ledger into cat usually loses any
output (unless the output is large enough to cause the buffer to flush):

    $ ./ledger -f "test/baseline/feat-value_py3.test" reg
    <class 'bool'> True
    [...]
    $ ./ledger -f "test/baseline/feat-value_py3.test" reg | cat
    $

Interestingly `--verify` causes `python_interpreter_t` to be destroyed
-- it doesn't appear to be otherwise -- which does call `Py_Finalize`.
As expected this fixes the issue but can also crash due to the boost
issue mentioned above:

    $ ./ledger -f "test/baseline/feat-value_py3.test" --verify reg
    <class 'bool'> True
    [...]
    Segmentation fault (core dumped)
    $ ./ledger -f "test/baseline/feat-value_py3.test" --verify reg | cat
    <class 'bool'> True
    [...]
    $

1. https://www.boost.org/doc/libs/1_62_0/libs/python/doc/html/tutorial/tutorial/embedding.html

### DIFF
--- a/src/pyinterp.cc
+++ b/src/pyinterp.cc
@@ -151,6 +151,11 @@ void python_interpreter_t::initialize()
     DEBUG("python.interp", "Initializing Python");
 
 #if PY_MAJOR_VERSION >= 3
+    // Unbuffer stdio to avoid python output getting stuck in buffer when
+    // stdout is not a TTY. Normally buffers are flushed by Py_Finalize but
+    // Boost has a long-standing issue preventing proper shutdown of the
+    // interpreter with Py_Finalize when embedded.
+    Py_UnbufferedStdioFlag = 1;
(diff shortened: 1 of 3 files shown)


q3aiml:
@xnox thanks for taking a look and all your work on python3 support!

PR #1878, opened 2020-03-31
Make explicit option a no-op

### LINKED ISSUE #1877: --explicit now a no-op?
The change from PR #1819 needs to be documented.

If I understand correctly. `--explicit` is now a no-op and should be removed from the documentation.

@CandyAngel?

CandyAngel:
This continues the work from #1819, resolves #1877 and also removes all the "dead" code that I can find from `explicit` becoming a no-op.

### DIFF
--- a/doc/NEWS.md
+++ b/doc/NEWS.md
@@ -4,6 +4,10 @@
 
 - Add support for '%F' date format specifier (bug #1775)
 
+- Entities are no longer regarded as defined due to being part of a cleared
+  transaction. `--explicit` is effectively enabled by default and is now a
+  no-op.
+
 ## 3.1.3 (2019-03-31)
 
 - Properly reject postings with a comment right after the flag (bug #1753)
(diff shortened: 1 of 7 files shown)


tbm:
Thanks for this.

Do you know if `fixed_payees` (et al) were actually used anywhere? I can't see them being used.

If they are not being used, you could also remove them from `src/journal.h` or does that break something?

CandyAngel:
Sorry, please bear with me. This breaks the test suite. I will update once I have sorted that out.

CandyAngel:
Seems I had removed a bit too much as the `explicit` option was no longer recognised at all, making the tests that use it fail.

As the intent is only to no-op it, I've left the opt-explicit test in, but removed the option from the other test that used it.

Should be ready now. Sorry again!

tbm:
Merged manually

tbm:
Thank you for completing your work @CandyAngel!

CandyAngel:
Can you tell I don't code in that style? I much prefer explicit braces everywhere but was trying to keep the existing style! Hope it wasn't too much hassle and thanks for merging this in.

PR #1880, opened 2020-04-01
fix macports dependencies on 10.14+

q3aiml:
10.14+ does not support universal builds:

	Error: python27 cannot be installed for the configured universal_archs 'x86_64 i386' because the arch(s) 'i386' are not supported.

From https://trac.macports.org/ticket/57369#comment:6

> You just can't install ports universal on Mojave, so don't try to do so. Don't specify +universal on the command line

### DIFF
--- a/acprep
+++ b/acprep
@@ -57,7 +57,7 @@ class BoostInfo(object):
             return [ 'boost' ]
 
         if system in ['darwin-macports']:
-            return [ 'boost-jam', 'boost', '+python27+universal' ]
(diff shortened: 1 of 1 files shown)


jwiegley:
I have no way of testing it, so if it worked for you, it looks good to me.

q3aiml:
Ahh I somehow got it in my mind that you used macports. But looking now I see zero evidence of that. Sorry, my mistake.

Issue #1894, opened 2020-05-06
Issue with expression evaluation


blatte:
I have automated transactions formated like so:

```
= /Expenses:Transportation/ and not expr "has_tag(/NOTAX/)"
         Expenses:Tax                     0.2
         $account                        -0.2
```

The idea is for the transaction to apply unless the account has the tag NOTAX.

This used to work but it now does not apply for any transcation at all.

tbm:
Confirmed to work with 3.1.2. Does not work with 3.2.0.

tbm:
This was introduced by:

commit 49b07a1c19489547b859d61fbc5c240aff224dda (refs/bisect/bad)
Author: John Wiegley <john@dfinity.org>
Date:   Tue Sep 17 16:54:58 2019 -0700

    Correction to the way parens are parsed in query expressions

@jwiegley can you take a look?

tbm:
Complete test case:

```
= /Expenses:Transportation/ and not expr "has_tag(/NOTAX/)"
         Expenses:Tax                     0.2
         $account                        -0.2

2020-05-06 * Test
    Expenses:Transportation            200.00 EUR
    Assets:Bank                       -200.00 EUR
```

Correct output:

```
         -200.00 EUR  Assets:Bank
          200.00 EUR  Expenses
           40.00 EUR    Tax
          160.00 EUR    Transportation
--------------------
                   0
```

jwiegley:
The query parser is I'm really not very happy with.

I confirm the bug, and have reverted my change.

Rahix:
I think I have also hit this bug with the following automatic transaction:

```text
= ^Expenses and %split-half and expr "any(account =~ /^A:Assets|^A:Expenses:Cash/ and R)"
    [A:Assets:Split]              (amount / 2)
    [B:Assets:Split]              (-amount / 2)
```

Using the latest version from git (869302ae9ce3) fixes the issue.

jwiegley:
It's good to have more examples, thank you. Before I make any future change to this code (to fix the issue I had run into), I'll build up a battery of regression tests to ensure we don't break people's Ledger files.

tbm:
@jwiegley I'm not sure if this is *strictly* related since I don't know what you were trying to fix, but when you look into this please also consider issue #872 and issue #1185.  They are examples of things not properly parsed today.

Rahix:
[...]

PR #1897, opened 2020-05-08
Add regression tests for issue #1894

Rahix:
Add tests for wrong behavior caused by commit 49b07a1c1948 ("Correction to the way parens are parsed in query expressions") which was reverted with commit 869302ae9ce3.

I made sure the tests fail prior to the fix and pass with the revert applied.


tbm:
Thank you, I was going to add those as well.

tbm:
@Rahix what does "and R" do?

Rahix:
@tbm: `R` is 1 if the posting is real and 0 if it is virtual.  The idea of `and R` is to filter out the virtual postings created by the automatic transaction here.

Issue #987, opened 2013-08-10
collapse and accumulate sub accounts to hide unwanted detail (BZ#987)


tbm:
Note: the issue was created automatically with bugzilla2github

Bugzilla bug ID: BZ#987
From: myles (@mylese)

Rahix:
This would be a really useful feature!  I have found the `--collapse`/`-n` flag but that only allows collapsing _everything_ right now.  Maybe it would be useful to add a number argument to `--collapse` for this instead of changing the behaviour of `--depth`?

Rahix:
If I am not mistaken, PR #216 is related?

beetleb:
I too need this:

https://groups.google.com/forum/#!topic/ledger-cli/XIh43fRvaRs

Rahix:
This issue can be closed now, I think?

atsaloli:
> This issue can be closed now, I think?

No, https://github.com/ledger/ledger/pull/216 was reverted

Rahix:
#216 was superseded by #1901

PR #1901, opened 2020-05-13
Make --depth in register report collapse sub-accounts

Rahix:
This is my attempt to revive PR #216 which was merged prematurely and then reverted again.  I've taken @johannesgerer's original work and built a (hopefully) working implementation of this feature (which was also reqested in #987).

This PR changes the behavior of `--depth` in the register report so that accounts are collapsed to the requested level instead of cut off.  This is essentially the `--collapse` behavior, but the given depth level specifies how far accounts should be collapsed (`--collapse`/`-n` behaves like `--depth 0`).

### Example
The following register report

```text
$ ledger reg ^Expenses
20-May-13 Test                Expenses:AA:BB          3.00 EUR    3.00 EUR
                              Expenses:AA             3.00 EUR    6.00 EUR
20-May-14 Test                Expenses:AA:BB:CC       2.00 EUR    8.00 EUR
                              Expenses:DD             4.00 EUR   12.00 EUR
```

can be collapsed to only the second level like this:

```text
$ ledger reg ^Expenses --depth 2
20-May-13 Test                Expenses:AA             6.00 EUR    6.00 EUR
20-May-14 Test                Expenses:AA             2.00 EUR    8.00 EUR
                              Expenses:DD             4.00 EUR   12.00 EUR
```

---

First of all: I have little experience with C++ so for the most part I just tried to mimic the rest of the codebase.  Please let me know if I fell into some pit-falls while doing that ...

I split the work into small commits to clearly separate the changes I needed to do.  The commit-messages contain the details and reasons for each change.

There are a few points that I'd like to bring up:

- [ ] Right now, I changed the behavior of `--depth` directly and unconditionally.  It would also be possible to retain the old `--depth` behavior by default and only change it if combined with `--collapse`.  I feel like this would be very unintuitive but if keeping the old `--depth` behavior is important, this would be an option (although `--display 'depth <= X'` will always archieve the same if needed).
- [ ] The interaction with `--collapse-if-zero` is weird because both are implemented by the same filter.  I think it should be possible to make the two work together more nicely but I have not tried that yet.  I think the best route would be to treat it as two separate instances of the same filter.  Is this something I should work on?
- [ ] Right now I am using an `account_t*` as the key of the totals map.  Will this always work or could it lead to issues when there are multiple `account_t` instances for the same account?
- [x] When virtual postings are involved, somehow them being virtual is lost.  I couldn't easily find out why this is the case ... What should be the behavior here?  Is it ok as is or should collapsed postings which are entirely virtual be marked as such?

### DIFF
--- a/doc/ledger.1
+++ b/doc/ledger.1
@@ -513,7 +513,10 @@ will print balances only for accounts with two levels, i.e.
 but not
 .Sy Expenses:Entertainment:Dining .
 This is a display predicate, which means it only affects display,
-not the total calculations.
+not the total calculations.  In register reports,
+.Fl \-depth
+will collapse postings of child accounts beyond the given level and add
+their balances to the parent account.
 .It Fl \-detail
 Related to
 .Ic convert
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -8012,7 +8012,7 @@ The market value of a posting or an account, without its children.
 The net gain (market value minus cost basis), for a posting or an
 account, without its children.  It is the same as @samp{v-b}.
 
-@item l
+@item depth
 The depth (``level'') of an account.  If an account has one parent,
 its depth is one.
 
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -6471,7 +6471,9 @@ Limit the depth of the account tree.  In a balance report, for example,
 a @samp{--depth 2} statement will print balances only for accounts with
 two levels, i.e. @samp{Expenses:Entertainment} but not
 @samp{Expenses:Entertainment:Dining}.  This is a display predicate, which
-means it only affects display, not the total calculations.
+means it only affects display, not the total calculations.  In register
+reports, @samp{--depth} will collapse postings of child accounts beyond
+the given level and add their balances to the parent account.
 
 @item --deviation
 Report each posting’s deviation from the average.  It is only meaningful
--- a/src/chain.cc
+++ b/src/chain.cc
@@ -193,10 +193,16 @@ post_handler_ptr chain_post_handlers(post_handler_ptr base_handler,
 
     // collapse_posts causes xacts with multiple posts to appear as xacts
     // with a subtotaled post for each commodity used.
-    if (report.HANDLED(collapse))
+    if (report.HANDLED(collapse)) {
+      unsigned short collapse_depth = 0;
+      if (report.HANDLED(depth_))
+        collapse_depth = lexical_cast<int>(report.HANDLER(depth_).str());
+
       handler.reset(new collapse_posts(handler, report, expr,
                                        display_predicate, only_predicate,
-                                       report.HANDLED(collapse_if_zero)));
+                                       report.HANDLED(collapse_if_zero),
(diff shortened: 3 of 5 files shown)


tbm:
> Right now, I changed the behavior of --depth directly and unconditionally. It would also be possible to retain the old --depth behavior by default and only change it if combined with --collapse. I feel like this would be very unintuitive but if keeping the old --depth behavior is important, this would be an option (although --display 'depth <= X' will always archieve the same if needed).

Can you post to the mailing list to discuss this change and seek feedback from users.

tbm:
This needs to be reviewed by John; the question about virtual accounts needs to be addressed too.

beetleb:
I agree that the virtual behavior must not be lost. I rely quite a bit on it.

Rahix:
@tbm, I am not too familiar with the workflow for ledger.  Do you mean I should post the entire patchset to the list?  Or just a mail explaining the change and linking here?

Rahix:
@beetleb, can you elaborate what exactly you need?  Right now (unless I missed anything), virtual postings which have not been filtered out earlier in the chain (e.g. by using `--real`) will be added to the account just like real postings.  This is the same behavior which `-n`/`--collapse` excerted before my changes.

The issue which I saw is only that for an account which consists entirely of virtual postings, it will show in the final register without the square brackets.

beetleb:
@Rahix 

Ah OK. This will work for me. I was worried it would not handle virtual postings numerically.

tbm:
>  Or just a mail explaining the change and linking here?

Since you're proposing a change to the current behaviour, I think it should be discussed on the list.  Not the actual patches, just the change in behaviour you're proposing.

(At least that's my view; John my have different opinions on this)

jwiegley:
I'm in favor of this change, once the matter of virtual accounts is handled. Otherwise, let's do it!

Rahix:
@jwiegley, can you give me some comments on my code + the open points in the original post?

> once the matter of virtual accounts is handled.

Can you elaborate what exactly you mean?  The point I raised in the OP or something different?

jwiegley:
The point raised above.

Rahix:
Sorry, but this doesn't really help me much ... To reiterate, the current behavior is this:

*Any virtual posting that was not filtered out (e.g. by `--real`) is added to the account just like a real posting.  That means real and virtual postings are added together, which is necessary to correctly collapse accounts containing both real and virtual postings.  However when an account contains only virtual postings, the resulting output in the register report will not enclose the account in `[...]` because during addition the postings were 'converted' into real postings.*

I could probably change that last part and check whether all postings which were added to an account are virtual and in that case mark the resulting account as virtual as well.  Is this what you meant?

tbm:
Isn't what you describe exactly the behaviour of ledger even without --depth?

tbm:
Originally you wrote "When virtual postings are involved, somehow them being virtual is lost", but if `--real` filters them out I think adding them up like other postings is the expected behaviour.  ledger doesn't show the brackets for virtual postings.

Rahix:
Okay, in that case everything is working already.

Issue #1783, opened 2019-04-06
Output of pricedb not deterministic / BaselineTest_cmd-pricedb fails


tbm:
@tko mentioned in issue #1747:

> I can't make BaselineTest_opt-lots trip anymore, but with (very many) repeated test runs BaselineTest_cmd-pricedb eventually fails.

I just saw this on Travis after applying the patch from @mbudde (of course it's not related to his patch): https://travis-ci.org/ledger/ledger/jobs/514501257

```
$ledger -f "/Users/travis/build/ledger/ledger/test/baseline/cmd-pricedb.test" pricedb
--
  @@ -1,3 +1,3 @@
   P 2012/01/02 00:00:00 GBP 1.20 EUR
  +P 2012/01/04 00:00:00 USD 0.63 GBP
   P 2012/01/03 00:00:00 EUR 1.3166666667 USD
  -P 2012/01/04 00:00:00 USD 0.63 GBP
```

tbm:
Paging @q3aiml :)

q3aiml:
@tbm sure, can you give me enough access so I can trigger easily trigger rebuilds without needing to push phony commits?

tbm:
@q3aiml do you know what permissions you need exactly to trigger rebuilds?

tbm:
@afh fixed a similar issue in commit c5366fea2f0cd9f48f0cdb013cf6d17b9ffcabcc by adding `--sort d`. I'm not sure if this is a fix or a workaround but maybe it should be done for BaselineTest_cmd-pricedb too.

afh:
Although adding `--sort date` seems to make the test pass more often, I have also seen test failures when running the following command multiple times:
`ctest -j8 --test-dir build --repeat-until-fail 5000 -R BaselineTest_cmd-pricedb`

tbm:
@afh do you have time to investigate?

jwiegley:
OK, I've run into this and have been able to reliably reproduce it locally. I'm pretty sure it's due to a pointer comparison in the sort of the output, instead of a string sort. Digging in now.

jwiegley:
Claude verified my assumption, and came up with a solution that avoid relying on mere pointer comparisons when enumerating the set of commodity symbols. Membership should be name-based rather than pointer-value-based, and output for the pricedb and prices commands should be explicitly date-ordered.

PR #1904, opened 2020-05-13
Update CMake FindPython module to 3.17.2 to fix #1893

### LINKED ISSUE #1893: Specifying Python version when running cmake
Hi!

I'm trying to build 3.2.0 with Python support. All dependencies are installed, however I have issue with adding Python support.

When running `cmake -DUSE_PYTHON=yes .`, it does detect Python, but only the latest version installed which is 3.8.2, but the main version I have is python 3.7 and Boost is also built only with python 3.7.

I tired passing `-DPYTHON_EXECUTABLE=/usr/bin/python3.7` but that doesn't work:
[ledger-1.log](https://github.com/ledger/ledger/files/4573848/ledger-1.log)

If I remove Python 3.8, then `cmake -DUSE_PYTHON=yes .` finishes without errors:
[ledger-2.log](https://github.com/ledger/ledger/files/4573847/ledger-2.log)

I'm on Gentoo, the following deps are installed:
- Python 3.6, Python 3.7 (this is the main version that is used by the system and is supported by Boost), Python 3.8,
- GCC 9.3.0
- Cmake 3.16.5
- ledger 3.2.0

Is there another option that could help?

ngleb:
Attempt # 2. Update CMake FindPython module to 3.17.2 to allow building with specified Python version and fix issue with include paths in 3.16.x. ledger tree uses 5 files from CMake tree, only 3 of them were changed:

    cmake/FindPackageHandleStandardArgs.cmake
    cmake/FindPython.cmake
    cmake/FindPython/Support.cmake

Fix for issue #1893.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ if (USE_PYTHON)
   if (PYTHON_FOUND)
     set(BOOST_PYTHON "python${Python_VERSION_MAJOR}${Python_VERSION_MINOR}")
     set(HAVE_BOOST_PYTHON 1)
-    include_directories(SYSTEM ${Python_INCLUDE_DIR})
+    include_directories(SYSTEM ${Python_INCLUDE_DIRS})
   else()
     set(HAVE_BOOST_PYTHON 0)
     message("Could not find a Python library to use with Boost.Python")
@@ -163,7 +163,7 @@ if (HAVE_BOOST_PYTHON)
 cmake_push_check_state()
 
 set(CMAKE_REQUIRED_INCLUDES
-    ${CMAKE_INCLUDE_PATH} ${Boost_INCLUDE_DIRS} ${Python_INCLUDE_DIR})
+    ${CMAKE_INCLUDE_PATH} ${Boost_INCLUDE_DIRS} ${Python_INCLUDE_DIRS})
 set(CMAKE_REQUIRED_LIBRARIES
     ${Boost_LIBRARIES} ${Python_LIBRARIES} ${PROFILE_LIBS})
 
--- a/cmake/FindPackageHandleStandardArgs.cmake
+++ b/cmake/FindPackageHandleStandardArgs.cmake
@@ -27,6 +27,8 @@ valid filepaths.
       [VERSION_VAR <version-var>]
       [HANDLE_COMPONENTS]
       [CONFIG_MODE]
+      [NAME_MISMATCHED]
+      [REASON_FAILURE_MESSAGE <reason-failure-message>]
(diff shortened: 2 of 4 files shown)


ngleb:
According to FindPython module docs, variable for include dirs for Python that is set by the module is:
`Python_INCLUDE_DIRS`, not `Python_INCLUDE_DIR` currently used in `CMakeLists.txt` and it was so from the very first release of this module:
https://github.com/ledger/ledger/blob/master/cmake/FindPython.cmake#L105-L106
https://github.com/Kitware/CMake/blob/v3.12.0/Modules/FindPython.cmake#L90-L91
https://github.com/Kitware/CMake/blob/v3.15.0/Modules/FindPython.cmake#L105-L106
https://github.com/Kitware/CMake/blob/v3.17.2/Modules/FindPython.cmake#L112-L113

But command line argument uses a different name which is `Python_INCLUDE_DIR` and the description is "The path to the directory of the ``Python`` headers. It will be used to compute the variable `Python_INCLUDE_DIRS`.":
https://github.com/Kitware/CMake/blob/v3.17.2/Modules/FindPython.cmake#L268-L270

ngleb:
Looks like some tests are failing because of the issue #1783.

I'm not sure if this command is correct for new FindPython module:
```
PYTHONPATH=. `cmake -LA . | sed -n 's/Python_EXECUTABLE:FILEPATH=//p'` python/demo.py
```
`Python_EXECUTABLE` variable is no longer public (is this a correct word for this?) but internal since 3.16 when they introduced command line arguments to manually specifying Python versions:
https://github.com/Kitware/CMake/commit/06d9e67fbd2b2dfc9cba12327866b2f73eab8a18

tbm:
@q3aiml can you take a look at this? Sorry to bother you but it seems you understand cmake.

tbm:
@ngleb thanks for doing the work. Sorry, I thought it would be an easy change to take the new version.  I don't know much about cmake so I can't really comment.

q3aiml:
@ngleb thanks for looking at this. It agree that `Python_EXECUTABLE` either became internal or cmake behavior changed around printing internal vars.

I opened #1909 because I think it may make sense to use cmake's built-in FindPython when cmake is newer rather than maintaining a copy here (updating for new python versions, for example), but I'm also fine with cherry-picking https://github.com/ledger/ledger/pull/1909/commits/aa82d43f19a59277f64e8bedbd8f16674477710f to this PR instead as that should work too.

tbm:
@ngleb thanks for your work. I decided to go with @q3aiml's approach in the other PR. Let us know if that doesn't work.

PR #1914, opened 2020-05-18
acprep: opt flavor should build optimized binary

kcwu:


kcwu:
(just in case it was configured to debug)

PR #1919, opened 2020-05-27
Remove the-kenny from list of maintainers in default.nix

purcell:
Without this fix, the Nix build is broken, which causes `ledger-mode` CI builds to fail.

See https://github.com/NixOS/nixpkgs/commit/887295fd2d8c4da06acdaa185cbb3cc214d83285

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -48,6 +48,6 @@ pkgs.stdenv.mkDerivation {
     '';
 
     platforms = pkgs.stdenv.lib.platforms.all;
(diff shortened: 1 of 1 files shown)


purcell:
Thanks @tbm

purcell:
Should this fix also end up in `next`? I'm a bit fuzzy on the branch setup here. We run the `ledger-mode` CI against both stable releases of `ledger` and the code currently on `next`, which is how we found this issue. Should we perhaps be targetting `master`?

tbm:
@purcell the next branch is no longer used; you should use master.

purcell:
Great, thanks.

PR #1924, opened 2020-06-10
Add support for split debit/credit columns to convert

### LINKED ISSUE #1120: Support debit and credit amounts in CSV convert command (BZ#1120)

Bugzilla bug ID: BZ#1120
From: Rodlogic

kwohlfahrt:
This adds support for a "debit" column to the `ledger convert` command. If both an "amount"/"credit" and "debit" are supplied for the same row, an error is reported. Includes a slight refactoring of masks/field types from a big if/else if ladder into an array and loop.

Please let me know if I chose good semantics. My C++ is a bit rusty, so let me know if I'm going against good style anywhere too.

Closes #1120


Panke:
This looks good to me. Can you leave one example in the documentation that still shows the old amount behaviour?

tbm:
> Can you leave one example in the documentation that still shows the old amount behaviour?

That sounds like a good idea. @kwohlfahrt?

kwohlfahrt:
Ah, missed the first comment, thanks for the reminder. I added a small example, but not a tested one since I wanted to keep the documentation succinct.

tbm:
I was hoping @jwiegley would review but I think he's too busy.

This generally looks good to me.

One question: why are amount and credit the same?  Why not amount and debit?

kwohlfahrt:
I think it only makes sense this way - a positive value in the "credit" column increases the amount in the account (at least that's how my bank's statements work). This is the same behaviour as ledger's "amount" column, so I made those two aliases for each other. The "debit" column has the opposite behaviour.

tbm (approved):
Thanks for your work, @kwohlfahrt!

Someone asked about just this feature on IRC a few days ago, so I'm sure this will be useful to some users.

PR #1945, opened 2020-07-17
Improve the example of displaying amount in another commodity

hieu-van:
The current one doesn't work, also Ledger has got a much neater command-line option to achieve the goal.
I replaced `-T` option with `-X` option.

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -1105,15 +1105,10 @@ current price of gold:
 $ ledger -X AU balance checking
 @end smallexample
 
-Although the total expression appears complex, it is simply saying
-that the reported total should be in multiples of AU units, where the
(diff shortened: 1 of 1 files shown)


tbm:
Adding @enderw88 who might have input.

hieu-van:
> Thanks.
> 
> Please look at the next sentence "Although the total expression appears complex..."
> 
> That needs to be updated or deleted. I think it can just be deleted.
> 
> I'm also not sure about the following example. It says 14.01 AU. But the latest pricedb shows:
> 
> P 2004/06/21 02:18:02 AU $400.00
> 
> 5000/400 = 12.5
> 
> So should the example be 12.50 AU? Or where does 14.01 come from?

Ah right, I was a little silly. I will look into that.

tbm:
Merged manually.

Issue #1721, opened 2019-01-15
CMake warning about deprecated policy related to RPATH


scfc:
CMake 3.11.2 warns:

    CMake Deprecation Warning at CMakeLists.txt:4 (cmake_policy):
      The OLD behavior for policy CMP0042 will be removed from a future version
      of CMake.
    
      The cmake-policies(7) manual explains that the OLD behaviors of all
      policies are deprecated and that a policy should be set to OLD only under
      specific short-term circumstances.  Projects should be ported to the NEW
      behavior and not rely on setting a policy to OLD.

The triggering code was added in commit ac5fd3afd3826a9b0fadcb270900cf42c5ee6e39 with the intention:

    Set CMake policy CMP0042
    
    to remove the warning below when running cmake.
    In the long run changing the CMake rules to set MACOSX_RPATH
    might be the better option.
    
      CMake Warning (dev):
        Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
        --help-policy CMP0042" for policy details.  Use the cmake_policy
        command to set the policy and suppress this warning.
        MACOSX_RPATH is not specified for the following targets:
    
         libledger
    
      This warning is for project developers.  Use -Wno-dev to suppress it.

[Policy CMP0042](https://cmake.org/cmake/help/latest/policy/CMP0042.html) is (AFAIUI) irrelevant to Linux where CMake 3.11.2 seems to do Almost The Right Thing™ by default (set `RPATH` for `ledger` to source directory when building and testing and to (for example) `/usr/lib64` when installing; some/most distributions will want to clear that field when installing by setting `CMAKE_SKIP_INSTALL_RPATH`).  So my instinct would be to remove the four lines from `CMakeLists.txt` and the one line from `src/CMakeLists.txt` that relate to `RPATH`.

But I do not know how this plays out under macOS.  I assume the default there is reasonable as well, but I have no way of testing this.

tbm:
@afh @xnox @scfc any comments on this?

xnox:
https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling

tbm:
@scfc ... ?

tbm:
@xnox any help?

tbm:
@q3aiml any chance you could look at this issue?

tbm:
@scfc @q3aiml can someone submit a patch for this?  It sounds like this needs to be fixed.

Rahix:
[...]

PR #1949, opened 2020-07-28
WIP: Implement cryptographic access to files using GPGME

spacefrogg:
###  Goal
Access encrypted files (given to ledger as filenames) directly from ledger the GPGME++ wrapper for GPGME.

### Current State
- Implemented file access as a `std::istream` class (same as files are currently used) named `decrypted_stream_t`
  - Maintain the same interface to files as before
- Encapsulate communication with GPG in `data_streambuffer_t` a `std::streambuf` class:
  - Initialize GPGME library (i.e., works with zero knowledge from the view of `decrypted_stream_t` users)
  - Open file with `fopen()` (GPGME requires a `FILE*` object)
  - Determine encryption method and establish encryption context (works transparently with unencrypted files)
  - Decrypt en bloc at creation time (GPG does not employ stream ciphers, so the whole content has to be decrypted upfront)
- Implement readout of decrypted content as `std::streambuf` backing store `data_streambuffer_t`
  - Read out data from GPGME while the stream is processed
- Complete decrypted content is held in memory (by GPGME)
- `FILE` object is immediately disposed after decryption
- Supports PGP (symmetric, asymmetric) and CMS (S/MIME) encrypted files
- No signature verification
- Rough performance evaluation:
  - On an Intel i5 7xxx, a 46 MB file takes about 3 seconds to decrypt (without disk I/O)

### Current possible deficiencies
- Even when a file is assumed to be unencrypted, it is still handled by GPGME. This might incur a performance penalty.

### Missing
- Tests (not working on NixOS due to linking errors in `MathTests` and `UtilTests`)
- Proper performance evaluation


spacefrogg:
This one is now good to go from my side.

tbm:
@spacefrogg thanks for submitting this.  It needs to be reviewed by @jwiegley

tbm:
I think the "default.nix: Fix testing" commit should be removed since it is unrelated.

If default.nix is broken, can you open a separate PR (the commit should describe what you're fixing).

tbm:
is the `export LD_LIBRARY_PATH=$PWD` still needed? I recall a fix for that (or maybe it was only a conversation without a fix)

tbm:
> is the export LD_LIBRARY_PATH=$PWD still needed? I recall a fix for that (or maybe it was only a conversation without a fix)
I think it might be related to issue #1721

spacefrogg:
> is the `export LD_LIBRARY_PATH=$PWD` still needed? I recall a fix for that (or maybe it was only a conversation without a fix)

Yes, it is needed in this circumstance and is only partially related #1721. The RPATH is set during build time to point to the final location of the library. During testing, though, the library is not there, yet. So, `ledger` won't find its own `libledger.so` during testing.

As some modifications to `default.nix` are related to this PR (adding dependencies and build flags). Should I keep the fixes in this PR, as well, or move them to a separate PR?

spacefrogg:
I see, you've already answered my question. I will make a separate PR.

tbm:
> As some modifications to `default.nix` are related to this PR (adding dependencies and build flags). Should I keep the fixes in this PR, as well, or move them to a separate PR?

Well, I was talking about the second commit, which looks unrelated.  But if it's related, leave it here.

tbm:
(I know you also modified default.nix as part of the GPGME change, but that was obviously related and should be part of the first commit)

Rahix:
I cannot find any documentation or examples on how this feature is meant to be used.  Am I correct to assume that ledger will now just automatically call into GPGME when it detects an encrypted file and decrypt it transparently?  I.e. there is no additional configuration necessary, it _just works_ (TM)?

spacefrogg:
Yes, it will just work. You could name your ledger `ledger.gpg` or so, for your own reference, but it won't matter for en-/decryption.

Issue #1942, opened 2020-07-08
No exception is thrown for balance assignments that should be unbalanced


daraul:
This is unusual, but please consider the two test cases below: 

<details>
<summary>This one should throw an error, but doesn't</summary>

```ledger
2018/01/01 * Opening Balance
    Assets:Checking                             $1000.00
    Equity:Opening Balances                     -$1000.00

2018/01/01 * Budget
    [Assets:Budget:Rent]                        $100.00
    [Assets:Budget:Emergency]                   $200.00
    [Assets:Checking]                           -$300.00

2018/01/01 * Budget
    [Assets:Budget:Rent]                        = $100.00
    [Assets:Budget:Emergency]                   = $400.00
    [Assets:Bank]                               -$500.00

test bal Assets
While parsing file "$FILE", line 13:
While balancing transaction from "$FILE", lines 10-13:
> 2018/01/01 * Budget
>     [Assets:Budget:Rent]                        = $100.00
>     [Assets:Budget:Emergency]                   = $400.00
>     [Assets:Bank]                               -$500.00
Unbalanced remainder is:
            $-300.00
Amount to balance against:
             $200.00
Error: Transaction does not balance
end test
```

</details>

<details>
<summary>This one throws the error as it should</summary>

```ledger
2018/01/01 * Opening Balance
    Assets:Checking                             $1000.00
    Equity:Opening Balances                     -$1000.00

2018/01/01 * Budget
    [Assets:Budget:Rent]                        $100.00
    [Assets:Budget:Emergency]                   $200.00
    [Assets:Checking]                           -$300.00

2018/01/01 * Budget
    [Assets:Budget:Rent]                        $0 = $100.00
    [Assets:Budget:Emergency]                   = $400.00
    [Assets:Bank]                               -$500.00

test bal Assets
While parsing file "$FILE", line 13:
While balancing transaction from "$FILE", lines 10-13:
> 2018/01/01 * Budget
>     [Assets:Budget:Rent]                        $0 = $100.00
>     [Assets:Budget:Emergency]                   = $400.00
>     [Assets:Bank]                               -$500.00
Unbalanced remainder is:
            $-300.00
Amount to balance against:
             $200.00
Error: Transaction does not balance
end test

```

</details>

The two problematic transactions should work out to an identical set of postings. I hope I'm not mistaken. 

The same thing seems to happen with real postings, not just virtual ones.

tbm:
so basically the last transaction should be equivalent to:

```
[...]

PR #1958, opened 2020-08-19
WIP: Do not treat balance assignments with 0 diff as a null posting

daraul:
I think I found the solution to #1942: 
Ledger is treating balance assignments that have a 0 diff as having a null posting, leading to the posting being auto-balanced and therefore causing incorrect values to be returned for the transaction.

I fixed this by just making the posting equal to `amt - amt` ($0.00). This works, though I don't think it's the right solution, nor am I certain that the unbalanced remainder and amount to balance against are correct in my testcases. 

Therefore I'm submitting this WIP PR so I can ask for some feedback.


daraul:
Upon further inspection, I don't think there's any corrections that are needed here. Could it be merged? Maybe I could merge the latest upstream changes?

tbm (approved):
I'm not familiar with the code but I looked through it and your fix seems correct to me.

Since John is pretty busy, I think I'll merge the change.  If there are problems, we can always fix later.

tbm:
Applied manually.  Thanks @daraul

PR #1967, opened 2020-10-12
Fix typo in ledger3.texi

kcwu:


tbm (approved):
Thanks for spotting this!

PR #1970, opened 2020-10-30
Fix regression in sorted_amounts

tbm:
Commit 557ab32 ("Expose a new utility function for balances:
sorted_amounts") split out some code into a separate function.
Unfortunately, in the process an "if" statement was dropped,
leading to amounts being shown that shouldn't be shown because
they are smaller than the display precision.


tbm:
@jwiegley did you drop that `if` for some reason or was that an oversight?

PR #1971, opened 2020-11-13
Fix build for x86_64-w64-mingw32

### LINKED ISSUE #1905: Ledger failed to build under Msys2 MinGW 64-bit runtime
I tried building ledger on Windows using Msys2 MinGW 64 bit runtime, and it failed with the following error:

```
[ 40%] Building CXX object src/CMakeFiles/libledger.dir/account.cc.obj
C:/ledger/ledger/src/account.cc: In function 'ledger::value_t ledger::{anonymous}::get_addr(ledger::account_t&)':
C:/ledger/ledger/src/account.cc:309:12: error: cast from 'ledger::account_t*' to 'long int' loses precision [-fpermissive]
  309 |     return long(&account);
      |            ^~~~~~~~~~~~~~
C:/ledger/ledger/src/account.cc: In function 'void ledger::put_account(boost::property_tree::ptree&, const ledger::account_t&, boost::function<bool(const ledger::account_t&)>)':
C:/ledger/ledger/src/account.cc:773:24: error: cast from 'const ledger::account_t*' to 'long unsigned int' loses precision [-fpermissive]
  773 |     buf << std::hex << reinterpret_cast<unsigned long>(&acct);
      |                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
make[2]: *** [src/CMakeFiles/libledger.dir/build.make:395: src/CMakeFiles/libledger.dir/account.cc.obj] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:270: src/CMakeFiles/libledger.dir/all] Error 2
make: *** [Makefile:183: all] Error 2
```

Build with Msys MinGW 32 bit works fine. Is this actually supported or is it expected behaviour?

coonce:
I have zero C++ experience, but after some Googling and trial and error I managed to get the 64-bit build to finish in MSYS2. :) I may be way off, but it works for me!

Fixes #1905


coonce:
Using MINGW64:
```
99% tests passed, 1 tests failed out of 385

Total Test time (real) =  57.93 sec

The following tests FAILED:
        382 - RegressTest_fix-missing-trans-in-last-budget-period (Failed)
```
For some reason, `--date-format "%F"` doesn't print any date at all, so test `382` fails...

tbm:
> For some reason, --date-format "%F" doesn't print any date at all, so test 382 fails...

I believe `%F` is a non-standard format string that might not be available on all platforms. You can change the test case to `%Y-%m-%d`.

tbm:
> Also, backslashes are confusing to regex parsers and
> since Windows can deal with regular slashes in file paths, just convert
> all path separators to '/'.

That's a bit ugly. What would it take to solve it properly (i.e. make it work with `\`)?

The `re.sub()` can just become a `.replace()` if that's there the problem is.

tbm (changes requested):
This broke the build:

/home/travis/build/ledger/ledger/src/main.cc:38:10: fatal error: boost/nowide/args.hpp: No such file or directory

coonce:
> 
> 
> This broke the build:
> 
> /home/travis/build/ledger/ledger/src/main.cc:38:10: fatal error: boost/nowide/args.hpp: No such file or directory

Hmm, I didn't realize that this was new in Boost 1.73 and the standalone dependency requires Boost >1.56. This may not be as easy as I thought. :)

https://github.com/boostorg/nowide

tbm:
This broke the build:

```
TypeError: 'encoding' is an invalid keyword argument for this function
```

tbm:
@q3aiml can you review these changes?

q3aiml:
> This broke the build:
> 
> ```
> TypeError: 'encoding' is an invalid keyword argument for this function
> ```

@coonce can you try fixing the above Python 2 error with `from io import open`?

It probably makes sense to drop Python 2 support at some point but I don't have a good gut feel for what older, still supported distro releases that would impact.

coonce:
I haven't really been able to work on this much to get all the encoding issues worked out. The differences between python2 and python3 plus the Windows command prompt issues make it tough to track down anything. Hopefully I can work on this more, but for now if I just stick to ascii characters in my ledger files and on the console, it works well (other than failing tests...)

I'm configuring like this:
`cmake -DBoost_USE_STATIC_LIBS=1 -DCMAKE_CXX_FLAGS="-static-libgcc -static-libstdc++" -DBUILD_LIBRARY=0`

That gives me a portable `.exe` that I can use in cmd, powershell, etc.

tbm:
Regarding "Use Boost library to read argv as UTF-8 in Windows ", which broke the build: can you check for the feature in CMakeLists.txt. Maybe @q3aiml can help.

coonce:
That's a great idea - looks like it fixes #1986.

And looking good on my machine:
`100% tests passed, 0 tests failed out of 389`

coonce:
I created a [GitHub Action](https://github.com/coonce/ledger/blob/ghaction/.github/workflows/run-tests.yml) to run the tests in Windows, but it's showing some failing tests. Unfortunately I can't reproduce any of them on my machine... 
```
 27 - BaselineTest_cmd-pricedb (Failed)
 34 - BaselineTest_cmd-source (Failed)
228 - RegressTest_1057 (Failed)
240 - RegressTest_1182_2 (Failed)
264 - RegressTest_1942_a (Failed)
265 - RegressTest_1942_b (Failed)
266 - RegressTest_1942_c (Failed)
267 - RegressTest_1942_d (Failed)
324 - RegressTest_786A3DD0 (Failed)
```
I have the [LastTest.log](https://github.com/ledger/ledger/files/5901428/LastTest.log) from the test run, but it doesn't look like the failures are related to text encoding.

tbm:
>  27 - BaselineTest_cmd-pricedb (Failed)

This happens on other platforms too.  See #1783

> 228 - RegressTest_1057 (Failed)

I've seen this one before.  This happens if the timezone doesn't get set properly.  See commit 8d61945ce7641b756de2ad253905f9fbb8828088   Maybe this doesn't work on Windows

the other ones are strange.  As if the output by ledger is wrong (the amount printed before the posting line).

tbm:
@coonce thanks for all your fixes and for your patience and persistence!

I took the liberty to clean up the Git commit history a bit and merged your follow-up fixes into the original patches.  I hope I didn't mess anything up.

Please test if things are still working for you.

tbm:
Closing - applied. Thanks!

coonce:
Looks good to me. Thanks!

PR #1989, opened 2021-01-12
Support for building on older cmake versions

sanel:
Backported python cmake module expects at least cmake 3.7 (requested
by policy), which will break builds on older cmake versions, like 3.5.

This fix will omit python module inclusion for any cmake version less
than 3.7 and the the build will continue without problems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,8 +13,10 @@ set(Ledger_TEST_TIMEZONE "America/Chicago")
 # Point CMake at any custom modules we may ship
 list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 if(${CMAKE_VERSION} VERSION_LESS "3.16.0")
-  # use backported module from 3.15 (introduced 3.12) to support older versions of cmake
-  list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake/python-backport")
+  if(NOT (${CMAKE_VERSION} VERSION_LESS "3.7"))
+    # use backported module from 3.15 (introduced 3.12) to support older versions of cmake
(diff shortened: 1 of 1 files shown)


tbm:
Which system are you on that doesn't have cmake 3.7?

It seems 3.7 came out in November 2016.

I wonder if it's better to make it explicit that a certain version of cmake is needed.

q3aiml:
@tbm I agree an explicit minimum version likely makes sense. That said I think this change is reasonable even if certain older cmake versions are considered not officially supported.

@sanel Nitpick, but I would find it more readable to use `VERSION_GREATER_EQUAL` rather than `NOT ... VERSION_LESS`. Something like this:

```if((${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.7.0") AND (${CMAKE_VERSION} VERSION_LESS "3.16.0"))```

What do you think?

tbm:
> if((${CMAKE_VERSION} VERSION_GREATER_EQUAL "3.7.0") AND (${CMAKE_VERSION} VERSION_LESS "3.16.0"))

That sounds cleaner. @sanel can you update your patch and I'll approve. Thanks

tbm (changes requested):
can you make the change suggested by @q3aiml

sanel:
Yes! Thank you for your suggestions :)

sanel:
Regarding the question for the system without cmake 3.7: cmake 3.5 it is part of Slackware 14.2 and newer version is in the work. Slackware also has community provided packages through [slackbuilds](https://slackbuilds.org/) and the latest Ledger version present out there is 3.1.3.

I presume it can't use the latest Ledger versions because it would have to replace system cmake installation or patch the source code in the fly.

This fix would make it buildable again and the change was small enough it didn't affect Ledger build scripts very much.

tbm:
> Slackware 14.2 and newer version is in the work

Thanks for the clarification.

tbm:
Applied, thank you @sanel!

PR #1991, opened 2021-01-16
Fix contrib bash-completion

mstmob:
Fix shellcheck warning SC2034: command appears unused.
Fix completion script if used by zsh with bashcompinit.


tbm (approved):
Looks good, thank you!

Issue #1998, opened 2021-01-30
Assertion error if multiple commodities are in one posting


Christoph-D:
In some cases, multiple commodities in the same posting trigger an assertion error. A minimal example:

```
2021-01-02 Convert cost basis to EUR (1 EUR = 1 USD)
    Assets            -1 Stock {100 USD} [2021-01-01]
    Assets             1 Stock {100 EUR} [2021-01-01]
    Equity
```

```
$ ledger -f test.leder bal
While parsing file "[...]/test.leder", line 4:
Error: Assertion failed in "[...]/src/commodity.cc", line 551:int ledger::commodity_t::compare_by_commodity::operator()(const ledger::amount_t*, const ledger::amount_t*) const: false
```

This regression was caused by commit 501fbc08ae5493db77bb34f4c4fbe1f3a3bc14e3 ( @jwiegley ) "Change compare_by_commodity to return an integer", because the commit did not only change the return type, it also changed the function's behavior to not always return after the recursive call to `commodity_t::compare_by_commodity()`.

tbm:
Fixed in b155f8928c6a33af42f859b27c83639b72517f5e by @Christoph-D

PR #1999, opened 2021-01-30
compare_by_commodity: Always return the result of the recursive call

Christoph-D:
Commit 501fbc08ae5493db77bb34f4c4fbe1f3a3bc14e3 changed the behavior of this function to not return the "equal" result (==0) from the recursive call. Previously, the function returned the result of the recursive call unconditionally.

The new behavior causes an assertion error for certain postings. The regression test added in this commit shows such a posting.

This commit reverts to the old behavior, which fixes issue #1998.

### DIFF
--- a/src/commodity.cc
+++ b/src/commodity.cc
@@ -465,18 +465,23 @@ int commodity_t::compare_by_commodity::operator()(const amount_t * left,
     amount_t rightprice(*arightcomm.details.price);
 
     if (leftprice.commodity() != rightprice.commodity()) {
-      // Since we have two different amounts, there's really no way
-      // to establish a true sorting order; we'll just do it based
-      // on the numerical values.
-      leftprice.clear_commodity();
-      rightprice.clear_commodity();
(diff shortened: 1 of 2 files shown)


Christoph-D:
Looks like my initial fix makes the new test flaky because the ordering is non-deterministic (only in cases that trigger an assertion error before the fix, so it's no worse than before). I'll see if I can make the ordering deterministic.

Christoph-D:
I updated the commit to add deterministic ordering in case the prices have different commodities and identical numerical values.

tbm:
so it's good to go from your side?

Christoph-D:
Good to go if the tests on Travis CI pass. The tests pass locally for me, but I don't have the multitude of environments that Travis has.

tbm (approved):
Thanks so much!

I'm not familiar with the code, but John is busy, and it seems obvious you've done your homework, so I'll merge it. Thanks!

PR #2018, opened 2021-03-15
Refer to the right amount used in the example

steelman:


tbm (approved):
Looks good.  Thank you!

PR #2021, opened 2021-04-05
Fix using day-of-week names for `xact` command.

kanreki:
Also fix end-of-command boundary checks for keywords "at", "to", etc.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,8 +18,6 @@ Boost libraries consist of inline functions and templates in header files.
 **[Boost.Python]**:  C++ library which enables seamless interoperability
 between C++ and the Python programming language.
 
-**[Cheetah]**: a Python templating engine, used by `./python/server.py`.
-
 **[CMake]**: A cross platform system for building from source code.  It uses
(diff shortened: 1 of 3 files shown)


tbm (commented):
Are there test cases to go with that?

kanreki:
I'll add something to baseline/cmd-entry.

By the way, I apologize for mixing in the additional commit regarding the "Cheetah" doc. I didn't realize that pull requests work this way (I have very little Github experience). Shall I remove that change, and maybe send it in again separately later?

Thank you for your prompt response to my first attempt.

tbm:
> Shall I remove that change, and maybe send it in again separately later?

That would be great

tbm:
fwiw @kanreki there was no need to close this and open a new one.  You can edit existing pull requests.  You can do "git rebase" on your end and then do a "git push --force" to update.

It's all a bit complicated the first time, but works well once you've figured it out.

kanreki:
Thanks for the tip!

Yeah, I figured there must be a way. But I decided to keep things dead simple in this case, until I get a bit more facile with it.

PR #2024, opened 2021-04-06
Show "FAILURE in exit code" when success is expected

### LINKED ISSUE #2023: test failure lacks explanation when unexpected non-zero exit
In Ledger's regression tests, in a case of an unexpected non-zero exit code, the test fails (as it should) but there is no indication what was wrong. This makes it difficult to diagnose.

(For comparison, when a non-zero exit is expected but the wrong non-zero value is produced, the test driver does provide output showing both expected and actual values. And when the expected output text differs from the actual output text, the test driver shows a "diff" listing.)

I have a fix for this which I will propose in a pull request.

kanreki:
Fixes #2023.

### DIFF
--- a/test/RegressTests.py
+++ b/test/RegressTests.py
@@ -182,7 +182,7 @@ class RegressFile(object):
                     printed = True
                 print(" ", line,)
 
-        if test['exitcode'] is None or test['exitcode'] == p.wait():
+        if test['exitcode'] == p.wait():
             if success:
                 harness.success()
             else:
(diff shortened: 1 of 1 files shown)


kanreki:
Could this PR be considered?

tbm (approved):
it's not entirely clear to me when that would be useful but I guess it won't hurt

kanreki:
Thanks, Martin! I tried to explain it in the bug (#2023); sorry if I wasn't clear.

It's useful during development of a test, especially for someone less familiar with the code. If you write a test where a non-zero exit code (failure) is expected, but forget to indicate that in the test description (as I clumsily did), then the test will fail but not provide any output giving a hint to why it failed. Basically the principle that any test failure should provide at least some hint as to why, to help with debugging.

Issue #1809, opened 2019-07-15
transaction ignored silently due to subtraction in amount expression


andres-erbsen:
```
browse@ashryn ~ % cat bad.ledger                                              
commodity $
payee landlord
account reserves
account housing

2019-07-15 landlord
  ; August (4..23]
  housing  (((23-4)*$85))
  reserves
browse@ashryn ~ % ledger --pedantic --check-payees --immediate -f bad.ledger r                                                                   
browse@ashryn ~ % ledger --pedantic --check-payees --immediate -f bad.ledger b
browse@ashryn ~ % 
```

mbbogh:
The **-** is interpreted as negative sign. Adding a space after it works fine.

`  housing  ((23- 4)*$85)`

Same is true for using an addition.

`  housing  ((23+-4)*$85)`

Looks like a parser error to me.

tbm:
Confirmed, same issue as #2001 which has some debug logs showing the subtraction is incorrectly calculated as 0.

tbm:
Fixed in commit 4ba80c37c068dd1

PR #2027, opened 2021-04-11
Fix amount tokenizer re: embedded minus sign.

oaf-edwin:
An amount may have a (single) leading minus sign, but none after that.
Bug #2001 (and #1809).

### DIFF
--- a/src/amount.cc
+++ b/src/amount.cc
@@ -975,8 +975,15 @@ namespace {
   {
     char buf[256];
     char c = peek_next_nonws(in);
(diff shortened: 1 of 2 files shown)


tbm:
This needs a review from @jwiegley

oaf-edwin:
> This needs a review from @jwiegley

OK, sure. What do I need to do to make that happen?

tbm:
> OK, sure. What do I need to do to make that happen?

I already requested review from @jwiegley Hopefully he'll find the time.

PR #2028, opened 2021-04-19
Fix IRC bullet point in README.md

catleeball:
Currently the IRC link is broken in the rendered readme page on Github.
Speculating that Github's markdown renderer might not display the text
as a link since it starts with "irc://" instead of "http://" or similar.

> ledger-readme-list-bug

Added some text to the bullet point specifying channel and server.

I left the link in-place in case the above issue is resolved at some
point, or if other websites render this README.md file without the
issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ Now that you're up and running, here are a few resources to keep in mind:
 
  - [Homepage]
  - [Documentation]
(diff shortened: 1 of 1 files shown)


tbm (approved):
Looks good, thank you!

PR #2034, opened 2021-05-30
Escape all string values output by the lisp report.

kreuter:
Notes were already being escaped. The filename, the transaction code,
the payee, postings' account names, and postings' cost expressions
were not. Everything else the lisp report prints appears to be a
number or a constant-valued non-string token for Lisp.

Here's an example journal file, for reproduction:

```
;; The lisp report prints the quantity as "123 "X Y"",
;; which won't parse properly in Emacs Lisp.
2021/05/30 * Buy some X Y
    A  123 "X Y"
    Z

;; Here's a pathological transaction with lots of stray double quotes
;; and backslashes. The lisp report won't parse right.
2021/05/30 * (1\2) Bad\ly named"payee 
    A:\:B  123 "X Y"
    Z
```

This is the output of the lisp report when that content is in /tmp/bad.ledger:

```
(("/tmp/bad.ledger" 4 (24755 3520 0) nil "Buy some X Y"
  (5 "A" "123 "X Y"" t)
  (6 "Z" "-123 "X Y"" t))
 ("/tmp/bad.ledger" 10 (24755 3520 0) "1\2" "Bad\ly named"payee"
  (11 "A:\:B" "123 "X Y"" t)
  (12 "Z" "-123 "X Y"" t)))
```

And here's the reference to the Elisp manual on String syntax:

https://www.gnu.org/software/emacs/manual/html_node/elisp/Syntax-for-Strings.html

### DIFF
--- a/src/emacs.cc
+++ b/src/emacs.cc
@@ -41,7 +41,7 @@ namespace ledger {
 	void format_emacs_posts::write_xact(xact_t& xact)
 	{
 		if (xact.pos)
-			out << "\"" << xact.pos->pathname.string() << "\" "
+			out << "\"" << escape_string(xact.pos->pathname.string()) << "\" "
 					<< xact.pos->beg_line << " ";
 		else
 			out << "\"\" " << -1 << " ";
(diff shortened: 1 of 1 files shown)


tbm (approved):
I don't know much about this lisp command, but escaping the strings seems like the right thing to do!

tbm:
Merged, thank you!

tbm:
BTW, next time please add a test case to the test suite (and for `emacs` you need to specify `--now`; see the other test case that's there already.)

kreuter:
Thanks, Martin! 

Might I suggest adding an item to the contributor guidelines about adding test cases? I read through CONTRIBUTING.md, but didn't see that; I would have written one up if I'd known that was expected/required.

tbm:
>Might I suggest adding an item to the contributor guidelines about adding test cases? I read through CONTRIBUTING.md, but didn't see that; I would have written one up if I'd known that was expected/required.

It took me a while but I finally added: "Please add a test case under test/regress when possible." Thanks

PR #2040, opened 2021-07-12
CI for Github actions for Mac and Linux

purplefishies:
The first part of the PR is for the Mac and Linux based Github action build

- It builds both mac and linux using the .travis.yml as the basis
- It builds Python3 since that what was requested but it would be easy to expand the matrix and build for python2
- It doesn't build Windows yet since I couldn't not find the chocolately dependencies for gmp, gpgme or mpfr. Hence I left it out.
- It doesn't produce an archive (yet) although that's an easy fix.
- The matrix setup seemed to be the best way going forward to allow for customization per specific platforms

The second part of the PR was to allow me to build Ledger without getting an error.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,66 @@
+name: CMake
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
+  BUILD_TYPE: Release
+
+jobs:
+  build:
+    runs-on: ${{ matrix.config.os }}
+    strategy: 
+      fail-fast: false
+      matrix:
+        # os: [ubuntu-latest,macos-latest]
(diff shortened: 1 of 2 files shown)


tbm:
Thanks for working on this @purplefishies, I appreciate it.

What is the 0c45295 commit? That looks unrelated to CI but some kind of build fix.  If so, it definitely needs a better commit message and probably should be in a separate PR.

tbm (changes requested):
Thanks for working on this @purplefishies, I appreciate it.

What is the 0c45295 commit? That looks unrelated to CI but some kind of build fix. If so, it definitely needs a better commit message and probably should be in a separate PR.

tbm:
@purplefishies ping?

tbm:
@purplefishies any comments regarding my question?

tbm:
I applied the commit that adds the GitHub CI file.  Thank you!

I dropped the other, unrelated commit. If you're fixing a build issue, please open a new PR.

Thank you!

PR #2047, opened 2021-08-04
Update NEWS with all changes up to now

Rahix:
I've went through the commit log since the last release and tried summarizing all changes.  I hope I catched everything of importance...

I included a second commit to move the `NEWS.md` file out of the `doc` folder and into the repository root.  I feel this helps giving it more visibility (after all, I didn't find it at first!) and is in line with what most other projects are doing these days.  If you do not like this change, I can drop the commit.

Cc: @tbm


Rahix:
@tbm, any comments?

tbm (approved):
Thanks for updating the NEWS file!

PR #2053, opened 2021-08-24
Fix for azure ubuntu not getting correct package server

purplefishies:
This update seems to address the problems that others have seen with azure repos disappearing when 
performing an apt-get


tbm (approved):
Let's try that, thanks!

tbm:
It worked, thank you!

PR #2055, opened 2021-08-29
check that commodity format uses correct symbol

### LINKED ISSUE #1948: doesn't reject invalid commodity in commodity format
@zacchiro filed a [bug on ledger2beancount](https://github.com/beancount/ledger2beancount/issues/215) involving this ledger example:

```
commodity CTGXXV
    format 1000.000 CTG25
```

I believe this is invalid ledger input that ledger should reject.

It's invalid in two ways:

1. CTG25 has to be a quoted commodity
2. It doesn't make sense to have two different commodities in the same `commodity` directive.

FWIW, hledger rejects this with:

```
commodity directive symbol "CTGXXV" and format directive symbol "CTG25" should be the same
```

madroach:
closes #1948

### DIFF
--- a/src/textual.cc
+++ b/src/textual.cc
@@ -1126,13 +1126,16 @@ void instance_t::commodity_value_directive(commodity_t& comm, string expr_str)
   comm.set_value_expr(expr_t(expr_str));
 }
 
-void instance_t::commodity_format_directive(commodity_t&, string format)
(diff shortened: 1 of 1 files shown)


tbm (approved):
Thanks, I think this is good.  I'm wondering if this will break something but I can't think of anything.

Could you add the test case to the test suite?

tbm:
Could you add the test case to the test suite?

PR #2056, opened 2021-08-29
do not enforce GNU C++, allow using c++ (clang)

madroach:
I don't know why this is prefilled with `g++`. I was annoyed by it, because my OpenBSD uses clang `c++`. Maybe others, too.

### DIFF
--- a/acprep
+++ b/acprep
@@ -239,7 +239,7 @@ class PrepareBuild(CommandLineApp):
         self.LDFLAGS         = []
 
         self.envvars = {
-            'CXX':      'g++',
+            'CXX':      '',
             'CXXFLAGS': '',
             'LDFLAGS':  '',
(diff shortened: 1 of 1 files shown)


tbm:
@jwiegley, @q3aiml any idea?

Rahix:
If it builds with clang and the resulting binary passes the testsuite, I'd say it is a good idea to not restrict to GNU toolchains.

tbm:
Well, the question is why it's there in the first place.

Maybe it can be removed, but maybe it should be replaced by `c++`.

I was hoping @jwiegley could comment.

kanreki:
Can't you simply specify the desired compiler by setting the CXX environment variable?

jwiegley (approved):

Issue #2058, opened 2021-08-31
Compiling on ARM produces  type-limits warnings


kanreki:
On architectures where the C `char` type defaults to unsigned (for example, ARM), compiling produces warning for these source files:

- query.cc
- times.cc
- token.cc

"warning: comparison is always false due to limited range of data type [-Wtype-limits]"

Briefly [mentioned previously on the mail list](https://groups.google.com/g/ledger-cli/c/bd8v3wiU6oo/m/DS5bufG3AQAJ).

kanreki:
A pull request is incoming shortly.

kanreki:
This affects error reporting. For example, if you invoke Ledger as ...

```
ledger --limit 'date>=[2021/9/1' register myaccount
```

(with missing closing ']'), on Intel, Ledger correctly reports "Missing ']'". The same test on ARM tries to output a hex 0xFF byte as a character  "Invalid char ÿ (wanted ']')".

(Though if your terminal environment is configured for UTF-8 you would see garbage instead of the ÿ. In that case you could pass the output through `iconv --from 8859_1` in order to see it.)

PR #2059, opened 2021-09-01
Use correct int return type for stream input operations

kanreki:
This makes it safe to compare results to -1 to indicate EOF,
regardless of whether char is considered signed or unsigned;
and so eliminates compiler warnings on platforms such as ARM.

Fixes bug #2058.

### DIFF
--- a/src/token.cc
+++ b/src/token.cc
@@ -148,7 +148,7 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
 
   char c = peek_next_nonws(in);
 
-  if (in.eof() || c == -1) {
+  if (in.eof()) {
     kind = TOK_EOF;
     return;
   }
@@ -199,7 +199,7 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
     char buf[256];
     READ_INTO_(in, buf, 255, c, length, c != ']');
     if (c != ']')
-      expected(']', c);
+      expected(']', in.eof() ? -1 : (c && 0xff));
 
     in.get(c);
     length++;
@@ -221,7 +221,7 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
     char buf[4096];
     READ_INTO_(in, buf, 4095, c, length, c != delim);
     if (c != delim)
-      expected(delim, c);
+      expected(delim, in.eof() ? -1 : (c && 0xff));
     in.get(c);
     length++;
     kind = VALUE;
@@ -233,9 +233,11 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
     in.get(c);
     amount_t temp;
     temp.parse(in, PARSE_NO_MIGRATE);
-    in.get(c);
-    if (c != '}')
-      expected('}', c);
+    int i = in.get();
+    if (i != '}')
+      expected('}', i);
+    else
+      c = i & 0xff;
     length++;
     kind  = VALUE;
     value = temp;
@@ -306,7 +308,7 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
       char buf[4096];
       READ_INTO_(in, buf, 4095, c, length, c != '/');
       if (c != '/')
-        expected('/', c);
+        expected('/', in.eof() ? -1 : (c && 0xff));
       in.get(c);
       length++;
 
@@ -420,14 +422,15 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
         if (in.fail() || ! in.good())
           throw_(parse_error, _("Failed to reset input stream"));
 
-        c = static_cast<char>(in.peek());
-        if (c != -1) {
-          if (! std::isalpha(c) && c != '_')
-            expected('\0', c);
+        int peeked = in.peek();
+        if (peeked == -1) {
+          throw_(parse_error, _("Unexpected EOF"));
+        } else {
+          if (! std::isalpha(peeked) && peeked != '_')
+            expected('\0', peeked);
+	  c = static_cast<char>(peeked);
 
           parse_ident(in);
-        } else {
-          throw_(parse_error, _("Unexpected EOF"));
         }
 
         if (! value.is_string() || value.as_string().empty()) {
@@ -503,19 +506,20 @@ void expr_t::token_t::unexpected(const char wanted)
   }
 }
 
-void expr_t::token_t::expected(const char wanted, char c)
+void expr_t::token_t::expected(const char wanted, const int c)
 {
-  if (c == '\0' || c == -1) {
-    if (wanted == '\0' || wanted == -1)
+  if (c == -1) {
+    if (wanted == '\0')
       throw_(parse_error, _("Unexpected end"));
     else
       throw_(parse_error, _f("Missing '%1%'") % wanted);
   } else {
-    if (wanted == '\0' || wanted == -1)
-      throw_(parse_error, _f("Invalid char '%1%'") % c);
+    char ch = c & 0xff;
+    if (wanted == '\0')
+      throw_(parse_error, _f("Invalid char '%1%'") % ch);
     else
       throw_(parse_error,
-             _f("Invalid char '%1%' (wanted '%2%')") % c % wanted);
+             _f("Invalid char '%1%' (wanted '%2%')") % ch % wanted);
   }
 }
 
--- a/src/query.cc
+++ b/src/query.cc
@@ -220,37 +220,9 @@ test_ident:
   return token_t(token_t::UNKNOWN);
 }
 
-void query_t::lexer_t::token_t::unexpected()
+void query_t::lexer_t::token_t::expected(char wanted)
 {
-  kind_t prev_kind = kind;
-
-  kind = UNKNOWN;
-
-  switch (prev_kind) {
-  case END_REACHED:
-    throw_(parse_error, _("Unexpected end of expression"));
-  case TERM:
-    throw_(parse_error, _f("Unexpected string '%1%'") % *value);
-  default:
-    throw_(parse_error, _f("Unexpected token '%1%'") % symbol());
-  }
-}
-
-void query_t::lexer_t::token_t::expected(char wanted, char c)
-{
-  kind = UNKNOWN;
-
-  if (c == '\0' || c == -1) {
-    if (wanted == '\0' || wanted == -1)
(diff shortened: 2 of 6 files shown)


src/token.cc
```diff
@@ -199,7 +199,7 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
     char buf[256];
     READ_INTO_(in, buf, 255, c, length, c != ']');
     if (c != ']')
-      expected(']', c);
+      expected(']', in.eof() ? -1 : (c && 0xff));
```
Rahix, line 202: I don't understand what this is supposed to do:

1. `&&` is a logical AND, so this operation will evaluate to `0` if c is `0` and `1` otherwise.  It looks an awful lot like you wanted to mask for the last byte, i.e. `& 0xff`?
2. `c` is type `char` so it cannot contain bits beyond mask `0xff` anyway... So even if you meant to mask, it would be superfluous?

src/token.cc
```diff
@@ -221,7 +221,7 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
     char buf[4096];
     READ_INTO_(in, buf, 4095, c, length, c != delim);
     if (c != delim)
-      expected(delim, c);
+      expected(delim, in.eof() ? -1 : (c && 0xff));
```
Rahix, line 224: Same here.

src/token.cc
```diff
@@ -306,7 +308,7 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
       char buf[4096];
       READ_INTO_(in, buf, 4095, c, length, c != '/');
       if (c != '/')
-        expected('/', c);
+        expected('/', in.eof() ? -1 : (c && 0xff));
```
Rahix, line 311: And same here.

src/token.cc
```diff
@@ -233,9 +233,11 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
...
     amount_t temp;
     temp.parse(in, PARSE_NO_MIGRATE);
-    in.get(c);
-    if (c != '}')
-      expected('}', c);
+    int i = in.get();
+    if (i != '}')
+      expected('}', i);
+    else
+      c = i & 0xff;
```
Rahix, line 240: What does this change do?  If I understand correctly, it just leads to c becoming 255/-1 depending on whether `char` is `unsigned`/`signed`?  But c is not used afterwards anyway?

src/token.cc
```diff
@@ -420,14 +422,15 @@ void expr_t::token_t::next(std::istream& in, const parse_flags_t& pflags)
...
-        if (c != -1) {
-          if (! std::isalpha(c) && c != '_')
-            expected('\0', c);
+        int peeked = in.peek();
+        if (peeked == -1) {
+          throw_(parse_error, _("Unexpected EOF"));
+        } else {
+          if (! std::isalpha(peeked) && peeked != '_')
+            expected('\0', peeked);
+	  c = static_cast<char>(peeked);
```
Rahix, line 431: You got a tab here which messes up the space indentation used everywhere else, it seems.

Rahix (commented):
Just took a quick look over your changes (I'm not affiliated with the project).  Found a few things which look a bit odd to me...

Just to clarify, is this just about compiler warnings or about actual wrong behavior on ARM?

src/token.cc
kanreki: Yikes!! Your (first) point is correct: it should have been `& 0xff`. Thank you so much for noticing that bug before it got merged!

As for the second point, the `expected()` method is now receiving an `int`, so the `char` is promoted with sign extension in the case of a signed char, I believe. Consider the following test program: 

```
% cat x.cc
#include <iostream>

void f(const char *, const int);

int main() {
  int waltz = 0;
  signed char c = '\377';

  f("bare", waltz ? 42 : c);
  f("with mask operation", waltz ? 42 : (c & 0xff));
}

void f(const char *msg, const int x) {
  std::cout << msg << ": " << x << std::endl;
}
% g++ x.cc -o xyzzy
% ./xyzzy
bare: -1
with mask operation: 255
% 
```

src/token.cc
kanreki: (Sure.)

src/token.cc
kanreki: The `expected()` method apparently relies on getting a -1 to indicate EOF. But `in.get(c)` does not promise to set `c` (at all) in the case of EOF, as far as I could determine, whereas `i = in.get()` (where `i` is `int` rather than `char`) of course does. So, that was the intent of the change.

As for whether `c` is used again afterwards, you're right: we could get rid of that last part. In the original code, `c` does end up holding the value of the character in the non-EOF case. So this change as written preserves that behavior, which makes it easier for a reviewer to read without looking beyond this localized context. I wouldn't mind removing those last couple lines.

src/token.cc
kanreki: Oops; I'll fix that.

src/token.cc
kanreki: And, yes, this set of changes was intended merely to eliminate compiler warnings on ARM. I had not encountered any incorrect calculations or other behavior in running.

Again, thank you for the very helpful review. I would say you are certainly affiliated with the project, just by virtue of this alone. And Github tags you as "Contributor" anyway!

Rahix: I see, so the `& 0xff` is a trick to essentially bitcast the signed integer into an unsigned one, right?  And you're doing this to preserve the most significant bit in non-ASCII input, essentially?  I.e. a `reinterpret_cast<unsigned char&>(c)` would archive the same, I assume?

Considering that, I think it could be shown that UTF-8 characters would break the previous implementation... No clue if UTF-8 is even meant to be properly supported in `ledger`?

> Again, thank you for the very helpful review. I would say you are certainly affiliated with the project, just by virtue of this alone. And Github tags you as "Contributor" anyway!

Haha, that's just because I had my own share of troubles for which I had sent a PR to fix :)

kanreki: As you say, I too struggled with uncertainty about what character encoding(s) Ledger is intended to support. I think we can assume that at least UTF-8 is supposed to be supported, just because we have source files like unistring.h and wcwidth.cc.

Could you help me understand what you mean about UTF-8 breaking the previous implementation? Note that the char `0xff` (`'\377'`) never occurs in properly encoded UTF-8, I believe.

I was trying to make my change in a way that would not break any character encoding that currently works. I noted that some other encodings (e.g., ISO-8859-1 and CP1252) do use position 255.

Specifically, yes, as you say, I'm trying here to handle characters that have their high-order bit set. I'm trying to stuff them back into an `int` because of the (IMHO) broken way that `READ_INTO` is cramming the result of `istream::peek()` into a `char`. It seems to me, `peek()` can return 257 possible values:

- any valid character (assuming 8-bit `char`), or
- -1 for the EOF signal

Actually I think a more correct fix would have been for me to change READ_INTO to operate with `int` values. I kinda wanted to do that, but that would have broadened the scope of the change, and I was afraid the project owners/reviewers would be reluctant to accept that. Perhaps I should have just tried it.

Anyway, your suggestion about the `reinterpret_cast` is an interesting one. At first I was thrown off by the explicit mention of `unsigned char`. But it got me to thinking, my version of the change (with the `0xff`) assumes that `char` is 8 bits. Maybe it's silly to worry about, but I guess theoretically there could be an architecture with a larger char. The `reinterpret_cast` avoids that assumption.

Rahix: > Could you help me understand what you mean about UTF-8 breaking the previous implementation? Note that the char 0xff ('\377') never occurs in properly encoded UTF-8, I believe.

Not `0xff`, but other bytes with the MSB set, e.g. `"💲"` becomes `b"\xf0\x9f\x92\xb2"`.  Or would those silently pass through the existing logic even if technically "negative numbers"?

> It seems to me, `peek()` can return 257 possible values:

Agreed.  It also looks like the previous code kind of used both `0` and `-1` for the purpose of encoding `EOF`?  So if the previous codebase did not properly support internal null-bytes anywhere anyway, maybe just using `\x00` as the `EOF` indicator would also work?

> Perhaps I should have just tried it.

It does sound like the more sane solution to me :)

> At first I was thrown off by the explicit mention of `unsigned char`.

Not sure I understand, what's the problem with that?

> assumes that `char` is 8 bits.

You see how much havoc the "`char` is `signed` or `unsigned` by default" is already wreaking.  I don't even want to imagine how much breaks when `char` is a different width entirely ^^

src/token.cc
Rahix: I'd drop any code which does not serve a useful purpose.  In the end, it will just clutter the file and make it even less obvious to future readers what a function is supposed to be doing.

src/token.cc
kanreki: > Not `0xff`, but other bytes with the MSB set, e.g. `"💲"` becomes `b"\xf0\x9f\x92\xb2"`. Or would those silently pass through the existing logic even if technically "negative numbers"?

I think those would pass through OK.

However, I just noticed the code in `commodity_t::parse_symbol()` (in src/commodity.cc) with hard-coded assumption to the format of UTF-8. Unclear to me why only commodity names get this special treatment; but I think this means other 8-bit encodings wouldn't work

Also, (in [the same commit that introduced that change](https://github.com/ledger/ledger/commit/6f2e3b88649543d625b76ac5542dd8838b28f0b1)) in src/system.hh, note the removal of preprocessor symbol `SUPPORT_UNICODE` and now unconditionally including "utf8.h".

> 
> > It seems to me, `peek()` can return 257 possible values:
> 
> Agreed.

By the way, some pretty good discussion on Stack Overflow on that topic ([here](https://stackoverflow.com/a/35356684/98761); and another one [here](https://stackoverflow.com/q/18013167/98761)).


> > a more correct fix would have been to change READ_INTO to operate with `int` values... I should have just tried it.
> 
> It does sound like the more sane solution to me :)

Maybe I'll try to work up an alternative change (in a separate PR) with that approach.


> > assumes that `char` is 8 bits.

There's a 256-word array `invalid_chars` array in src/commodity.cc, so I guess we can stop worrying about that issue.

kanreki: > > > a more correct fix would have been to change READ_INTO to operate with `int` values... I should have just tried it.
> > 
> > 
> > It does sound like the more sane solution to me :)
> 
> Maybe I'll try to work up an alternative change (in a separate PR) with that approach.

I have created another PR as #2063; and I guess I'll cancel the present PR.

Thanks again for your help!

PR #2060, opened 2021-09-05
Support for `NO_COLOR` environment variable

nagy:
See https://no-color.org/ for more information about the variable.

### DIFF
--- a/src/report.cc
+++ b/src/report.cc
@@ -816,7 +816,7 @@ value_t report_t::fn_format_datetime(call_scope_t& args)
 
 value_t report_t::fn_ansify_if(call_scope_t& args)
 {
-  if (args.has<string>(1)) {
+  if (args.has<string>(1) && !getenv("NO_COLOR")) {
     string color = args.get<string>(1);
     std::ostringstream buf;
     if (color == "black")          buf << "\033[30m";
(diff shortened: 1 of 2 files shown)


Rahix (approved):
Because the great thing about standards is that there are so many, there is also the [`CLICOLOR` standard](https://bixense.com/clicolors/) with a similar goal.  But it looks like `NO_COLOR` is more widely adopted so I think this change is a good idea.

tbm:
`report.h` has:

```
    OPTION_(report_t, no_color, DO() {
        OTHER(color).off();
      });
```

Is there a reason we can't do a similar call (not sure where though) if the environment variable is set?

nagy:
> Is there a reason we can't do a similar call (not sure where though) if the environment variable is set?

That seems like a better way. I will see if I can make it that way.

nagy:
I will undraft this PR once I think it is ready.

nagy:
I have given this a try and I cannot find a suitable place to place the logic for this.

I have also found that I can get the effect that I wanted by doing:

```
export LEDGER_NO_COLOR=1
```

That is enough for my needs, so feel free to close this PR if nobody wants to take it on.

PR #2062, opened 2021-09-08
Fix config path in man page

cbarrete:
By the way, is there an ETA for the next release? It would be really nice to have the XDG_CONFIG stuff in the distribution packages.


tbm (approved):
That's obviously correct, thank you!

tbm:
There has been some work on coordinating a new release (see the mailing list). Hopefully soon.

PR #2063, opened 2021-09-14
Use correct int return type for stream input operations

kanreki:
This makes it safe to compare results to -1 to indicate EOF,
regardless of whether char is considered signed or unsigned;
and so eliminates compiler warnings on platforms such as ARM.

Fixes bug #2058.

### DIFF
--- a/src/amount.cc
+++ b/src/amount.cc
@@ -974,13 +974,13 @@ namespace {
   void parse_quantity(std::istream& in, string& value)
   {
     char buf[256];
-    char c = peek_next_nonws(in);
+    int c = peek_next_nonws(in);
     int max = 255;
     char *p = buf;
     if (c == '-') {
       *p++ = c;
       max--;
-      in.get(c);
+      in.get();
     }
     READ_INTO(in, p, max, c,
               std::isdigit(c) || c == '.' || c == ',');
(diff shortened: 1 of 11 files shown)


tbm:
@Rahix since you had comments on #2059 do you have any comments on this?

kanreki:
@alan-andrade 

> Still pending to try this out compilation to ARM ...

Would it be helpful for me to supply details/output of my build on ARM (I've done it on two different ARM devices)?

kanreki:
> I was able to compile this for ARMv8 64bit without a problem on a buildroot project. I imagine the 64bit fixes this issue "automatically"

I don't think 64-bit makes a difference: one of my devices is an aarch64, and I was getting the warnings there. But note that, IIRC, I think the warning is only produced for "debug" builds, so if you had it configured for "opt" that could explain it.

kanreki:
> Ahhh I see. It's probably that since I imagine "debug" builds are not the default. I went with the default options.

Actually, I believe that "debug" _is_ the default, at least when you run a plain `./acprep`. I'm not familiar with buildroot, though, so I don't know if that changes anything, nor whether it faithfully emulates everything about a native build.

alan-andrade:
It doesn't use the acegrep script since biildroot doesn't have a package
manager, so I had to install all dependencies and libraries. It runs cmake
without extra params.

kanreki:
@tbm Hello again, Martin. Anything I can do to encourage some activity/progress on this?

tbm:
I was hoping @jwiegley would review but he hasn't been around much lately.

jwiegley:
Sorry to be a bit out of touch, just getting back to my GitHub notifications this week.

jwiegley (approved):

tbm:
Thanks @jwiegley!

Thanks @kanreki for finding and fixing this issue (and doing the rework) and to @Rahix for the review!

Issue #1176, opened 2016-08-03
--time-colon mishandles sub-hour times and displays imbalanced transactions (BZ#1176)


tbm:
Note: the issue was created automatically with bugzilla2github

Bugzilla bug ID: BZ#1176
From: unknown reporter

tbm:
Fixed in commit ed58869  by @rafasc

PR #2064, opened 2021-09-25
Fix --time-colon for negative time amounts

rafasc:
While the current formula works for positive numbers, negative numbers
are incorrectly represented.

One of the issues comes from the fact that floor(x) < x for every x.
`amount_t precision` will always be a non negative number and the code
that attempts to fix the issue for negative number will never run.

If we truncate the number instead, the current formula works for both
positive and negative numbers without making negative numbers a corner
case. So let's do that.

### DIFF
--- a/src/amount.cc
+++ b/src/amount.cc
@@ -733,13 +733,9 @@ void amount_t::in_place_unreduce()
 
   if (shifted) {
     if (("h" == comm->symbol() || "m" == comm->symbol()) && commodity_t::time_colon_by_default) {
-      amount_t floored = tmp.floored();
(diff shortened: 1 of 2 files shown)


rafasc:
This should close #1176

tbm (approved):
I can't really evaluate this but since it makes the testcase work and since John is busy I'll approve this.

Thank you!

tbm:
Thank you!

CharlaLuetmer2001:
Same thing here

tbm:
@CharlaLuetmer2001 does the fix from @rafasc work for you?

PR #2072, opened 2021-12-02
Add flake

matthewbauer:
This might be useful to anyone using Nix. This allows you to use the new Nix flakes feature. If not, you can still use shell.nix or default.nix

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,25 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1638407071,
+        "narHash": "sha256-xbveILjgtBVh6B5F6i2k3T0IrE8lZp1vsqfDY+Df/cg=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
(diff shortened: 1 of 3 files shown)


tbm:
I don't know anything about Nix.  Do all of these files have to be in the main directory or could they be moved to a sub-dir?

purcell:
@matthewbauer heads-up that we use `default.nix` to build a ledger dev snapshot for use in the `ledger-mode` CI. The existing default.nix is broken, but it would be good to line things up so that everything works again (ideally without requiring the adoption of flakes in that other repo, personally speaking).

akirak:
I created #2082 which would hopefully fix the CI failure in ledger-mode.

FYI: If you had flake.nix in `ledger-mode`, you could test ledger-mode against this PR without #2082 being merged into master, e.g.:

```sh
nix build --override-input ledger github:matthewbauer/ledger/add-flake
```

jwiegley (approved):

jwiegley:
Pushed!

PR #2082, opened 2022-01-17
Point licenses to lib.licenses

akirak:
`pkgs.stdenv.lib.licenses` is no longer available in recent versions of nixpkgs. Use `lib.licenses` instead.

I think this will fix the build failure in https://github.com/ledger/ledger-mode/pull/323.

Note that this change is already included in #2072, so you don't need this if you merge the PR that adds a flake.


tbm:
@matthewbauer   @purcell can you review this?

purcell (approved):
Yes, looks good, thanks

tbm (approved):
Thank you!

PR #2086, opened 2022-02-01
csv parser: mark guessed account as calculated

madroach:
This will change the output of the `convert` command so that the amount is printed
on the bucket account, not the guessed account.
```
2022/02/01* Blublub
    Expenses:Groceries                   100,00
    Accounts:Cache
```
becomes
```
2022/02/01* Blublub
    Expenses:Groceries
    Accounts:Cache                      -100,00
```
This makes subclassifying after conversion a bit easier


tbm:
I'm not sure I see the advantage of this.  Can you elaborate?

Why not put the amount on both postings; but maybe some people don't like that

madroach:
When importing via csv you always get transactions with two postings. One is the posting on the _real_ account the csv is about. The other one is guessed.
Now I often want something like this:

```
2022-02-02 Supermarket
    Assets:Credit Card    -50
    Expenses:Groceries     10
    Expenses:Household     20
    Expenses:Drinks        20
```

- It is easier to get there if there is already an amount on the asset. Otherwise I need to move the amount from an arbitrary account to the asset and negate it. I want to avoid that.
- It makes sense, because the csv was about the asset, not the guessed account. Indeed the other account has a _calculated_ amount to balance the real asset account.

Putting the amount on both postings would be fine with me, but this would require a change to `print.cc`, which would affect more than just the output of `convert`.

tbm:
Ok, thanks for the clarification.  I agree with your points; otoh, people might not like the negative numbers that the change will cause (even if they accurately represent the transaction).

I think I'll leave this up for John to decide (but he has been busy).

You may want to email the ledger mailing list about this and gather feedback from other users.

JesseFL:
I agree with this change 100%.  I think users will often want to split their income/expenses and moving the posted amount to the bank account would facilitate this.  Further, the CSV is likely provided by a source for **one account**, the bank/credit account, and it would logically make sense for `ledger convert` to attach the amount in the CSV to the referenced account.

tbm:
can someone email the ledger mailing list to discuss this change and see what the feedback is.

I think the arguments here are good, but since it's a change in behaviour I think it's better to discuss it more widely.

JesseFL:
Added
https://groups.google.com/g/ledger-cli/c/_eXMLgohq6Y

JesseFL:
Only received one comment, but it was from @jwiegley, so it may be worth more than one vote ;)

> Sounds like a great feature request to me!
> 
> John

jwiegley:
Could we add an option to do this flipping? It could change the workflow for many people.

tbm:
@madroach what do you think of John's suggest to make it an option.

madroach:
sure, that would be fine with me.

jwiegley:
@madroach Great, just let me know when that change is in!

afh:
@madroach Mind merging the current state of master, which has seen some improvements regarding the checks run on PRs, to this PRs branch too?

afh:
Friendly ping for @madroach

madroach:
I rebased my change. Sorry for taking so long. @afh, @jwiegley

jwiegley (approved):

afh:
I took the liberty to address the breaking tests. The changes can be found [here](https://github.com/ledger/ledger/compare/master...afh:ledger:pr/2086). I'm ot sure how to best update this PR with these changes as this PR's branch is on a fork, hence I created a PR [madroach/ledger#1](https://github.com/madroach/ledger/pull/1). 

@madroach would you be willing to accept that, so this PR's branch gets those changes and we can move forward with merging this PR?

@jwiegley Are you okay with the fixes (see [Fix DocTests](https://github.com/madroach/ledger/pull/1/commits/7805702b751e5d088f1554966337ca4b5308c822) and [tests: Fix Baseline and Regression tests](https://github.com/madroach/ledger/pull/1/commits/355149917db982d47fe313b1986d888470396f3e)) and are any changes to the documentation necessary?

madroach:
Merged. Thanks for the work. I couldn't find the time for it...

afh:
Thanks for merging, @madroach, very much appreciated! No worries regarding not finding the time, often enough other obligations stand in the way of the fun stuff 🙂

jwiegley:
I think this is set to go, it just still needs the option to flip the behavior, which should be relatively easy to add as a `session_t` option.

afh:
@madroach would you be okay with me taking over the work and adding the requested option as well as accompanying documentation and tests?

If yes, I'd close this PR in favor of one I'd create in order to more easily push changes to it. You'd be credited with `Co-authored-by: @madroach` on one of the commits unless you object.

madroach:
Yes, please do! I just won't find the time.

Issue #2075, opened 2021-12-17
Test failure with Boost 1.77.0 (regress/BF3C1F82-2.test)


foutrelis:
Doesn't seem too serious, but I'm not sure what's causing it (also fails with Boost 1.78.0, though 1.76.0 was fine):

```
FAILURE in error output from /build/ledger/src/ledger-3.2.1/test/regress/BF3C1F82-2.test:
--
$ledger -f - reg
--
  @@ -1,2 +1,2 @@

   While parsing file "", line 2:

  -Error: File to include was not found: "./non-existent-ledger-file-BF3C1F82"

  +Error: File to include was not found: "non-existent-ledger-file-BF3C1F82"

E[BF3C1F82-2.test]STDERR:
b''
FAILURE in error output from /build/ledger/src/ledger-3.2.1/test/regress/BF3C1F82-2.test:
--
$ledger -f /dev/stdin reg
--
  @@ -1,2 +1,2 @@

   While parsing file "", line 2:

  -Error: File to include was not found: "./non-existent-ledger-file-BF3C1F82"

  +Error: File to include was not found: "non-existent-ledger-file-BF3C1F82"

E[BF3C1F82-2.test]STDERR:
b''

FAILED (2)
```

yangsheng6810:
Having the same problem. On a ledger file that was reconciled and balanced, I am seeing the following when doing ledger-reconcile:
```
 File to include was not found
```
Also have the problem of
```
Transaction does not balance
```

bcc32:
Yep, this also breaks things like `M-x ledger-add-transaction` in ledger-mode, since it also reads in the contents of the current buffer from stdin.

XieJiSS:
Same problem here. 

```
Test #360: RegressTest_BF3C1F82-2 ................................***Failed    1.89 sec
```

PR #2089, opened 2022-02-14
WIP: fileinfo_t: remove size member

MEschenbacher:
it is currently not required but set. Incidentally, calling file_size()
on a bash pipe e.g. /proc/self/fd/11 fails with:

Error: boost::filesystem::file_size: Operation not permitted: "/proc/self/fd/11"

when executing ledger as

ledger -f <(cat journal)

### DIFF
--- a/src/journal.h
+++ b/src/journal.h
@@ -78,21 +78,19 @@ public:
   struct fileinfo_t
   {
     optional<path>   filename;
-    boost::uintmax_t size;
     datetime_t       modtime;
     bool             from_stream;
 
-    fileinfo_t() : size(0), from_stream(true) {
+    fileinfo_t() : from_stream(true) {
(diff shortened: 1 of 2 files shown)


tbm:
Review requested from @jwiegley

jwiegley:
Further changes are needed:
```
/home/runner/work/ledger/ledger/src/py_journal.cc: In function ‘void ledger::export_journal()’:
/home/runner/work/ledger/ledger/src/py_journal.cc:255:55: error: ‘size’ is not a member of ‘ledger::journal_t::fileinfo_t’
  255 |                   make_getter(&journal_t::fileinfo_t::size),
      |                                                       ^~~~
```

jwiegley:
Tests failing on macOS, I wonder if it's an intermittent failure?

tbm:
the macOS failure is unrelated, see #2075

PR #2092, opened 2022-03-09
Ensure absolute path for include

### LINKED ISSUE #2075: Test failure with Boost 1.77.0 (regress/BF3C1F82-2.test)

kunhtkun:
Ensure the path of file to include in `instance_t::include_directive`
is always an absolute path.

Previously when the journal file is given through stdin, we prepend a
"./" to the filename to include. However, in Boost >= 1.77,
`path::normalize` strips the leading "./" [1]. Our `resolve_path`
function calls `normalize` and thus now it returns "file" for "./file"
instead of the previous "./file".

This change causes a failing test regress/BF3C1F82-2 [2], and also
breaks the `include` directive for stdin input:

    $ touch file-to-include
    $ echo "include file-to-include" | ledger -f - reg

gives

    While parsing file "", line 1:
    Error: File to include was not found: "file-to-include"

Therefore, we change to prepend the `context.current_directory` to
make the filename absolute in this case as well. The test
regress/BF3C1F82-2 is also updated to match the new output.

Fixes #2075.

[1] https://github.com/boostorg/filesystem/commit/16bd89b7c0398d0dc5904148a865ef3fc3ece7ec
[2] https://github.com/ledger/ledger/issues/2075

### DIFF
--- a/src/textual.cc
+++ b/src/textual.cc
@@ -743,7 +743,7 @@ void instance_t::include_directive(char * line)
     DEBUG("textual.include", "parent file path: " << context.pathname);
     path parent_path = context.pathname.parent_path();
     if (parent_path.empty()) {
-      filename = path(string(".")) / line;
+      filename = context.current_directory / line;
     } else {
       filename = parent_path / line;
       DEBUG("textual.include", "normalized path: " << filename.string());
--- a/test/regress/BF3C1F82-2.test
+++ b/test/regress/BF3C1F82-2.test
@@ -4,11 +4,11 @@ include non-existent-ledger-file-BF3C1F82
 test -f - reg -> 1
 __ERROR__
 While parsing file "", line 2:
-Error: File to include was not found: "non-existent-ledger-file-BF3C1F82"
+Error: File to include was not found: "$sourcepath/non-existent-ledger-file-BF3C1F82"
 end test
 
 test -f /dev/stdin reg -> 1
 __ERROR__
 While parsing file "", line 2:
-Error: File to include was not found: "non-existent-ledger-file-BF3C1F82"
+Error: File to include was not found: "$sourcepath/non-existent-ledger-file-BF3C1F82"
 end test


kunhtkun:
By the way, I checked other places that used `resolve_path` and there seemed to be no breakage due to the Boost change.

In 

https://github.com/ledger/ledger/blob/474d2b8d091ba865c17c554a4ecdfb0d05325d3f/src/context.h#L119-L120 

the path is made absolute right after, and in

https://github.com/ledger/ledger/blob/474d2b8d091ba865c17c554a4ecdfb0d05325d3f/src/session.cc#L96

the path is also turned into absolute before actually being read (implicitly through `parse_context_stack_t::push`)

https://github.com/ledger/ledger/blob/474d2b8d091ba865c17c554a4ecdfb0d05325d3f/src/session.cc#L132-L135

Thanks!

kunhtkun:
By the way, I briefly checked and there seems to be no breakage due to the Boost change in the two other places that call `resolve_path`.

Thanks!

tbm:
@kunhtkun thanks so much for submitting a fix for this issue!  That's one of the stoppers for getting a new release out.

I'm not famliar with the code but why do we have to prepand a directory at all?

Would just

> filename = / line;

work?

I'm sure there's a reason this doesn't work.  Just asking.

kunhtkun:
Hi @tbm! You are welcome and thanks for taking care of this awesome project!

Regarding the prepended path, we actually requires an existent parent path:

https://github.com/ledger/ledger/blob/474d2b8d091ba865c17c554a4ecdfb0d05325d3f/src/textual.cc#L758-L767

otherwise `include_directive` will not try to find the file at all. That is also why adding "./" fails after Boost 1.77 -- now it strips the leading "./" which is the parent path.

Honestly the snippet linked above is a bit more complicated than I would imagine, but I guess there is a reason to do so.

tbm (approved):
I've run some tests and I think this works as expected.

Kunht, thanks so much for this fix.  This issue has been outstanding for far too long.

If you have time to fix other issues, please submit PRs and John and I will review. Ledger needs help.

Issue #2057, opened 2021-08-29
Segfault when trying to add commodity to null-commodity


madroach:
I tried to somehow emulate the `D` directives behaviour of `hledger` in `ledger` to configure a default commodity added to any commodityless amount. Doing so I triggered a segfault:

test.journal:
```
commodity €
    format 1.000,00 €

C 1 "" = 1 €

2021-01-01 Testtransaction
    blub        1000 €
    blab
```
```
(gdb) run -f test.journal balance
Starting program: /home/madroach/src/ledger/ledger -f test.journal balance
[New process 36632]

Thread 1 received signal SIGSEGV, Segmentation fault.
ledger::amount_t::in_place_unreduce (this=<optimized out>) at amount.cc:735
735         if (("h" == comm->symbol() || "m" == comm->symbol()) && commodity_t::time_colon_by_default) {
(gdb) print comm
$1 = (ledger::commodity_t *) 0x0
(gdb) bt
#0  ledger::amount_t::in_place_unreduce (this=<optimized out>) at amount.cc:735
#1  0x0000028ab228d087 in ledger::value_t::unreduced (this=<optimized out>) at value.h:476
#2  ledger::report_t::display_value (this=<optimized out>, val=<error reading variable: Unhandled dwarf expression opcode 0xa3>) at report.cc:513
#3  0x0000028ab225945d in ledger::format_accounts::mark_accounts (this=0x28b496f1a00, account=..., flat=<optimized out>) at output.cc:229
#4  0x0000028ab22592fd in ledger::format_accounts::mark_accounts (this=0x28b496f1a00, account=..., flat=<optimized out>) at output.cc:207
#5  0x0000028ab225974f in ledger::format_accounts::flush (this=0x28b496f1a00) at output.cc:251
#6  0x0000028ab22b42c0 in ledger::item_handler<ledger::account_t>::flush (this=0x7f7fffff7498) at chain.h:76
#7  ledger::pass_down_accounts<ledger::basic_accounts_iterator>::pass_down_accounts (this=<optimized out>, handler=..., iter=..., _pred=..., _context=...)
    at filters.h:1069
#8  0x0000028ab228c5e1 in ledger::(anonymous namespace)::accounts_flusher::operator() (this=0x7f7fffff7808) at report.cc:443
#9  0x0000028ab228b7cd in ledger::report_t::accounts_report (this=0x28b2a63e000, handler=...) at report.cc:482
[...]

PR #2093, opened 2022-03-12
Avoid dereferencing null pointer

kunhtkun:
Here `comm` could be a null pointer (the new test regress/2057 shows such a case). So test it before dereferencing to avoid segfault.

This is related to #2057 but does not fully resolve it (the "feature request" part still exists).


tbm (approved):
thanks!

PR #2094, opened 2022-03-12
Add a NEWS entry for ba8ec324

kunhtkun:
I realized that #2092 introduced some user visible change (the error message). So it might be better to have a NEWS entry for it.


tbm (approved):
thanks!

PR #2097, opened 2022-03-18
README.md: Fix for Homebrew and Conda

kunhtkun:
Hi!

If I understand it correctly, the README file in this repo is for developers who want to build Ledger. So we have instructions for installing *build dependencies* on Ubuntu, Fedora, and macOS with MacPorts. However, for Homebrew and Conda the instructions are actually on installing the *compiled Ledger cli itself*. That looks a bit confusing and inconsistent. So I update them to give instructions about installing build dependencies.

For conda-forge, I have tested it on a Linux x64 machine. I did not try the Homebrew instructions but the packages are copied from the CI workflow (with `git` and `gpgme` omitted):

https://github.com/ledger/ledger/blob/f3fd025a35b7328f66e21db1eccfe774cae4cf70/.github/workflows/cmake.yml#L47

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,18 +72,14 @@ Dependency | Version (or greater)
 
 ### macOS
 
-You can use [Homebrew] or [MacPorts] to install Ledger easily on macOS.
+You can use [Homebrew] or [MacPorts] to install build dependencies for Ledger
+easily on macOS.
 
 #### 1. Homebrew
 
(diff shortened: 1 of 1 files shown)


tbm (approved):
You make a fair point.

Maybe there should be installation instructions separately somewhere (although that's probably on the web site already, and people looking at the source probably won't need a package).

PR #2107, opened 2022-04-14
Remove some UTF-8 code that was having no effect

### LINKED ISSUE #2061: some code specific to UTF-8 actually accomplishes nothing
There's some code in `commodity_t::parse_symbol(std::istream&, string&)` (src/commodity.cc) for doing something special with non-ASCII UTF-8 characters. However, it doesn't seem to actually make any difference: it copies bytes into a buffer; but all bytes (except "`invalid_chars`") get copied into the buffer anyway. So this special code seems to be pointless.

Is there some subtlety I'm missing? If not, we might as well excise this useless code.

kanreki:
Fixes #2061.

### DIFF
--- a/src/commodity.cc
+++ b/src/commodity.cc
@@ -307,47 +307,14 @@ void commodity_t::parse_symbol(std::istream& in, string& symbol)
       throw_(amount_error, _("Quoted commodity symbol lacks closing quote"));
   } else {
     char * _p = buf;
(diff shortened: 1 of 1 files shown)


kanreki:
(as discussed with @jwiegley last week)

tbm (approved):
Reviewed by @kcwu and @jwiegley already.

Thank you @kanreki

PR #2109, opened 2022-04-15
When calculating balance assignments, strip away lot annotations

jwiegley:
This allows the following to work:
```
2021/11/22 Entry
    Assets:Foo        100 FOO (Note) {$43.44}
    Income:Bar

2021/11/22 Balance Check
    Assets:Foo                                0 = 100 FOO
```
Without this change, you get an error on the balance check because the lot annotations don't match up.

### DIFF
--- a/src/textual.cc
+++ b/src/textual.cc
@@ -1666,18 +1666,20 @@ post_t * instance_t::parse_post(char *          line,
       balance_t diff = amt;
 
       switch (account_total.type()) {
-      case value_t::AMOUNT:
-        diff -= account_total.as_amount();
+      case value_t::AMOUNT: {
+        amount_t amt(account_total.as_amount().strip_annotations(keep_details_t()));
+        diff -= amt;
         DEBUG("textual.parse", "line " << context.linenum << ": "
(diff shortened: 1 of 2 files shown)


jwiegley:
@tbm I already hear you asking for a test in the back of my mind, so I'll have to add that to this PR once I have a moment free, likely in May.

tbm (commented):
The example you posted already works for me (without your patch), so I'm a bit confused.

jwiegley:
@tbm Then I need a more involved example, that was just my guess at the reduced form.

jwiegley:
@tbm here is an example that shows the fix:
```
2021/11/22 Entry
    Assets:Foo        100 FOO (Note) {$43.44}
    Income:Bar

2021/11/22 Entry
    Assets:Foo        100 FOO (Misc) {$45.44} = 200 FOO
    Income:Bar

2021/11/22 Entry
    Assets:Foo        $100
    Income:Bar

2021/11/22 Balance Check
    Assets:Foo                                0 FOO = 200 FOO
```

tbm (changes requested):
John, can you add the example you posted as a testcase.

jwiegley:
@tbm Added

PR #2112, opened 2022-05-05
Fix unrounding for equity

dmage:
A naive fix that makes `reg --equity --unround` and `equity --unround` work. Today `equity` ignores `--unround` and `reg --equity --unround` doesn't display amounts that are zero when they are rounded. Perhaps we need a more sophisticated fix as it doesn't fix `reg --equity --amount 'unround(amount_expr)'`


tbm (approved):
I can't evaluate your change, but your test case appears to make sense to me, so I'll add it.

PR #2114, opened 2022-05-05
Bump required CMake version

### LINKED ISSUE #1721: CMake warning about deprecated policy related to RPATH

dmage:
CMake 3.21.4 reports that compatibility with CMake < 2.8.12 will be removed from a future version of CMake.

CMake 3.0 has new defaults for MACOSX_RPATH. It seems ledger works fine with them. See CMake Policy CMP0042 for the details.

Fixes #1721.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.5)
+cmake_minimum_required(VERSION 3.0)
 
 PROJECT(ledger)
 


tbm (approved):
CMake 3.0 appears to be *ancient*, so I don't think this should cause problems for anyone.

PR #2116, opened 2022-06-25
Document --exchange can accept multiple commodities

gwymor:
The --exchange option can accept multiple comma-separated commodities,
but this is only mentioned in one place where --exchange is documented,
and notably not in the Texinfo manual.

[ci skip]

### DIFF
--- a/doc/ledger.1
+++ b/doc/ledger.1
@@ -172,11 +172,14 @@ with this command are:
 Show the running average, rather than a running total.
 .It Fl \-current Pq Fl c
 Don't show postings beyond the present day.
-.It Fl \-exchange Ar commodity Pq Fl X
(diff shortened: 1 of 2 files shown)


tbm:
yes, but there must not be a space otherwise it's seen as a different argument.

Does *not* work:

    ledger bal ets:cash -X EUR, PHP

Works:

    ledger bal ets:cash -X EUR,PHP
    ledger bal ets:cash -X "EUR, PHP"

I'm not sure if this should be explicitly mentioned.

tbm:
I'm not so familiar with texinfo syntax, but if that works, I think that would be good.

Maybe also add an example somewhere.

Thanks for working on this!

tbm (approved):
Thank you so much!

PR #2117, opened 2022-06-25
acprep: add support for arch linux

balajisivaraman:
Hello folks,

I just tried to setup the Ledger codebase on Arch Linux and discovered that the `acprep` script does not handle it already.

So I extended it to also work on this distribution.

I followed existing setup for other distros such as Fedora to install all needed dependencies.

So hopefully this is fine.

Thanks.


tbm (approved):
Thank you!

PR #2119, opened 2022-07-01
Update copyright year to 2022 and improve update_copyright_year script

afh:


tbm (approved):
I think you still have commit access. Feel free to commit directly.

tbm:
technically, none of this is correct since copyright should only be updated when there are new changes; but I know that ledger has done it this way (updating all years), so I'll put it in.

afh:
Yes, I still have commit access, yet prefer peer review of proposed changes.

I'm happy update the `tools/update_copyright_year.sh` script, so that it only updates the files that have been changed in a given year.
Should be easy as replacing `git ls-files` with something like `git log --name-only --pretty=format:'' --since=01/01/$(date +%Y) | sort -u`

Do you think that's worthwhile, @tbm?

PR #2121, opened 2022-07-05
Update flake.nix

afh:
This PR brings ledger's `flake.nix` a bit closer to the [`default.nix` in nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/office/ledger/default.nix) by adding the `usePython` and `useGpgme` "options" allowing to specify whether support for Python bindings and gpg encrypted journal files should be built into ledger.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -2,6 +2,7 @@
   description = "A double-entry accounting system with a command-line reporting interface";
 
   outputs = { self, nixpkgs }: let
+    useGpgme = true;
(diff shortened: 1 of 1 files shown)


tbm (approved):
Thanks.

Is there a reason Python is off by default while GPG is on?

afh:
Thanks for merging, @tbm. Yes, the reason Python is off by default is that it was off before this PRs changes.

Issue #2132, opened 2022-10-01
Bug in handling utf-8


horasio:
In latest commits from July 20:

Local config : MacOS 12.6 Monterey, Terminal in UTF8.
I guess the fix above is not enough because the July 19 commit works well with the Euro symbol as a commodity :
Python 3.7.14 (default, Sep 10 2022, 11:16:56)
[Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.

```
import ledger
a=ledger.Amount('€1.3')
a
€1.3
```
But with this commit, it's all buggy :

```
ledger.Amount('€1.3')
ﾬ 1
```

See how my Euro symbol has become a Korean syllable!

afh:
A PR (#2133) addressing this regression is now available, @horasio. Any testing you could do and feedback you could give is greatly appreciated :pray:

horasio:
All good as far as my tiny test goes. See my comment in your PR. Thank you!

PR #2123, opened 2022-07-05
Fix use of deprecated Python legacy Unicode API

afh:
replacing it with APIs introduced in Python 3.3 to ensure ledger's Python bindings continue to function when the legacy Unicode API is removed in Python 3.12.

For details see ["Unicode Objects and Codecs" in the Python/C API Reference Manual](https://docs.python.org/3.10/c-api/unicode.html).

<details><summary>Python C/API Deprecation Warnings Emitted During Compilation of Ledger's Python Bindings</summary>

```
~/Developer/ledger/src/py_utils.cc:117:25: warning: '_PyUnicode_get_wstr_length' is deprecated [-Wdeprecated-declarations]
      Py_ssize_t size = PyUnicode_GET_SIZE(obj_ptr);
                        ^
/nix/store/6a94sw6yyc412y3a0qjnjdyljx3p1b7n-python3-3.10.5/include/python3.10/cpython/unicodeobject.h:261:7: note: expanded from macro 'PyUnicode_GET_SIZE'
      PyUnicode_WSTR_LENGTH(op) :                    \
      ^
/nix/store/6a94sw6yyc412y3a0qjnjdyljx3p1b7n-python3-3.10.5/include/python3.10/cpython/unicodeobject.h:451:35: note: expanded from macro 'PyUnicode_WSTR_LENGTH'
#define PyUnicode_WSTR_LENGTH(op) _PyUnicode_get_wstr_length((PyObject*)op)
                                  ^
/nix/store/6a94sw6yyc412y3a0qjnjdyljx3p1b7n-python3-3.10.5/include/python3.10/cpython/unicodeobject.h:445:1: note: '_PyUnicode_get_wstr_length' has been explicitly marked deprecated here
Py_DEPRECATED(3.3)
^
/nix/store/6a94sw6yyc412y3a0qjnjdyljx3p1b7n-python3-3.10.5/include/python3.10/pyport.h:513:54: note: expanded from macro 'Py_DEPRECATED'
#define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
                                                     ^
~/Developer/ledger/src/py_utils.cc:117:25: warning: 'PyUnicode_AsUnicode' is deprecated [-Wdeprecated-declarations]
      Py_ssize_t size = PyUnicode_GET_SIZE(obj_ptr);
                        ^
[...]

PR #2133, opened 2022-10-01
Fix regression in Python/C++ unicode string introduced with #2123

afh:
The changes in this PR address #2132.

The regression was introduced by casting the return value of the different string access macros (e.g. `PyUnicode_1BYTE_DATA`) to `const char*` instead of using the proper typealias (e.g. `Py_UCS1`), which can be an 8, 16, or 32 bit type (see [`python/cpython/Include/unicodeobject.h`](https://github.com/python/cpython/blob/7546914e3f8157d8bad2998677e0ea3ed4cb7939/Include/unicodeobject.h#L100-L104)).

To properly handle the [Euro sign `€`)](https://en.wikipedia.org/wiki/Euro_sign) 16 bits are needed (UCS-2 codepath), yet ledger only allowed for 8, which truncated upper bits of the character for the Euro sign from `U+20AC` to to `U+FFAC`, i.e. `ﾬ`.

While working on this fix to address #2132 I noticed that certain strings (e.g. `¤ 12.34`) are not properly handled and result a `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa4 in position 0: invalid start byte` is thrown at runtime.
The [generic currency sign `¤`](https://en.wikipedia.org/wiki/Currency_sign_(typography)) can be encoded with 8 bits (UCS-1 codepath) yet directly creating a string from the `PyUnicode_1BYTE_DATA` return value does not result in a proper UTF-8 string.

I decided to replace the `EUR` currency in the `python/demo.py` with `€`, so that this fix for the regression is tested. The `XCD` and `EUR` commodities test the same codepath and seem to provide less value. Let me know if you believe that is too drastic of a change, @tbm…

### DIFF
--- a/python/demo.py
+++ b/python/demo.py
@@ -55,11 +55,11 @@ def assertEqual(pat, candidate):
 comms = ledger.commodities
 
 usd = comms.find_or_create('$')
-eur = comms.find_or_create('EUR')
 xcd = comms.find_or_create('XCD')
 
-# Tests currency symbols encoded with UCS-1. For details see #2132.
-xxx = comms.find_or_create('¤')
+# Tests currency symbols encoded using UCS. For details see #2132.
+eur = comms.find_or_create('€') # UCS-2 / UCS-4
+xxx = comms.find_or_create('¤') # UCS-1
 
 assert not comms.find('CAD')
 assert not comms.has_key('CAD')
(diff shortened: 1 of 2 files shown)


horasio:
I'm not qualified to review the code, but I can confirm that the € bug is gone :-)
`ledger.Amount('€1.23')` seems to work as expected, at least, it parses correctly and I have no more Korean "translation."
Thanks!

tbm (approved):
Thanks @horasio for reporting this and to @afh for fixing it so quickly!

afh:
Thanks, @horasio, for testing this PR and confirming that the changes in this PR fix [the bug you reported](#2132); much appreciated!

Thanks, @tbm, for merging this PR so swiftly 👍

PR #2143, opened 2022-11-11
Add python support for error_context

adamipc:
Added ledger.session.error_context() to py_session.cc to get the error context message when exceptions are thrown.

This is needed for me as the ArithmeticError thrown when a transaction doesn't balance doesn't include the unbalanced remainder. So adding this allowed me to retrieve the amount that the transaction was off by.

First PR so let me know if anything else is needed. I looked at the python unit tests but looks like they were never updated for python 3.

### DIFF
--- a/src/py_session.cc
+++ b/src/py_session.cc
@@ -33,6 +33,7 @@
 
 #include "pyinterp.h"
 #include "pyutils.h"
+#include "error.h"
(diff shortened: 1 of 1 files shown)


tbm:
@afh @jwiegley can you review this

jwiegley (approved):

PR #2156, opened 2022-12-29
Python: test cleanups

davidsklar:
1. Remove references to modules (exceptions, StringIO) no longer needed for Pythn3

2. Use assertEqual instead of assertEquals

3. Clear journal files with a close_journal_files() function that uses the then-current underlying python_session pointer. Calling session.close_journal_files() sometimes leads to segfaults because python_session has changed after it was injected into the python module (as "session") on module startup.

### DIFF
--- a/src/py_session.cc
+++ b/src/py_session.cc
@@ -55,6 +55,9 @@ namespace {
   {
       return str_to_py_unicode(error_context());
   }
+  void py_close_journal_files() {
+    python_session->close_journal_files();
+  }
 }
 
 void export_session()
@@ -74,6 +77,8 @@ void export_session()
 
   scope().attr("session") =
     object(ptr(static_cast<session_t *>(python_session.get())));
+  scope().attr("close_journal_files") =
+    python::make_function(&py_close_journal_files);
   scope().attr("read_journal") =
     python::make_function(&py_read_journal,
                           return_internal_reference<>());
--- a/test/python/JournalTest.py
+++ b/test/python/JournalTest.py
@@ -6,7 +6,7 @@ from ledger import *
 
 class JournalTestCase(unittest.TestCase):
     def tearDown(self):
-        session.close_journal_files()
+        close_journal_files()
 
     def testBasicRead(self):
         journal = read_journal_from_string("""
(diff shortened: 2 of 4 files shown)


afh:
ℹ️ macOS build error related to #2157

afh:
Thanks for this PR, @davidsklar.

1. and 2. look good to me.
Regarding 3.: What would it take in order to address the segfaults? Ledger's session handling is something I've been struggling with ever since. I'm not sure about its benefits and have run into issues with it. Mind sharing your understanding of the issue at hand, @davidsklar?

davidsklar:
>        Regarding 3.: What would it take in order to address the segfaults? Ledger's session handling is something I've been struggling with ever since. I'm not sure about its benefits and have run into issues with it. Mind sharing your understanding of the issue at hand, @davidsklar?

I *think* what's happening here is:

When `export_session()` (in py_session.cc) gets called, it puts the `session` object into the Python module with this:

```c++ 
scope().attr("session") =
    object(ptr(static_cast<session_t *>(python_session.get())));
```

So whatever `session_t *` is held by the `python_session` shared_ptr at that point is what the python object `session` points to / contains.

But, other things can change what `python_session` points to, e.g. `initialize_for_python()` in `pyinterp.cc` 

So, if you try to reference the python `session` object after that, it contains the wrong `python_session` object.

I am not familiar enough with Boost.Python to know if/how there's a way to expose an object via `attr()` such that the value of the object gets calculated at time-of-access, not time-of-definition. If we could do that, it would definitely be cleaner. We want the python `session` object to reference the shared_ptr, and then retrieve the raw pointer inside of it when accessed (vs referencing the raw pointer when the `session` object is defined, which is what happens now.)

Adding the `close_journal_files()` (that uses the C++ `python_session` pointer at time-of-call) was a way around this so that tests could clear out anything in the journal and have independent data to work from.

afh (approved):
Thank you for this PR, @davidsklar. I think these changes are good to me merged, @tbm.

I created #2163 to dig into your helpful analysis of the underlying issue further.

tbm:
Thank you,  @davidsklar!

PR #2168, opened 2023-01-16
Rename quoted_rfc4180 to quoted_rfc, to avoid parser quirk

### LINKED ISSUE #2007: CSV output invalid escape sequence
`ledger csv` renders escaped quotes inside fields (for example, from a description `John's Hardware - bought 2" rubber nails`) as `\"`. Per the grammar in [RFC4180](https://tools.ietf.org/html/rfc4180#section-2), and common parsers, this is incorrect, and should instead be a sequence of two double quotes `""` (as mad as that is).

Concretely:
```ledger
2021/1/1 John's Hardware - 2" rubber nails
    Expenses:DIY     -123
    Assets:Credit
```

renders:
```csv
"2021/01/01","","John's Hardware - 2\" rubber nails","Expenses:DIY","","-123","",""
"2021/01/01","","John's Hardware - 2\" rubber nails","Assets:Credit","","123","",""
```

but we expect:
```csv
"2021/01/01","","John's Hardware - 2"" rubber nails","Expenses:DIY","","-123","",""
"2021/01/01","","John's Hardware - 2"" rubber nails","Assets:Credit","","123","",""
```

oaf-edwin:
Ledger's expression parser considers quoted_rfc4180 to be an amount (quantity 4180 of commodity "quoted_rfc").

Fixes #2007.

### DIFF
--- a/doc/ledger.1
+++ b/doc/ledger.1
@@ -1284,11 +1284,11 @@ for values that have a per-unit cost.
 .It Fn quoted expression
 Surround
 .Ar expression
-with double-quotes.
-.It Fn quoted_rfc4180 expression
+with double quotes.
+.It Fn quoted_rfc expression
 Surround
 .Ar expression
-with double-quotes, compatible with rfc 4180.
+with double quotes, compliant with rfc 4180.
 .It Sy real
 .\" Is there a difference between real and actual?
 Return true if the transaction is real, i.e not an automated or virtual
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -8518,7 +8518,8 @@ Return the quantity of @var{value} for values that have a per-unit cost.
 @end defun
 
 @defun quoted expression
-Surround @var{expression} with double-quotes.  If expression contains a double-quote, it will be escaped with a backslash.
+Surround @var{expression} with double quotes.  If expression contains
+a double quote, it will be escaped with a backslash.
 @smallexample @c command:EAD8AA7,with_input:3406FC1
 $ ledger -f expr.dat --format "%(quoted(account)) %(quoted(amount))\n" reg
 @end smallexample
@@ -8528,16 +8529,10 @@ $ ledger -f expr.dat --format "%(quoted(account)) %(quoted(amount))\n" reg
 @end smallexample
 @end defun
 
-@c @defun quoted_rfc4180 expression
-@c Surround @var{expression} with double-quotes, compliant with RFC 4180.  If expression contains a double-quote, it will be represented with two double-quotes.
-@c @smallexample @c command:EAD8AA7,with_input:3406FC1
-@c $ ledger -f expr.dat --format "%(quoted_rfc4180(account)) %(quoted_rfc4180(amount))\n" reg
-@c @end smallexample
-@c @smallexample @c output:EAD8AA7
-@c "Assets:Cash" "¤ -123,45"
-@c "Expenses:Office Supplies" "¤ 123,45"
-@c @end smallexample
-@c @end defun
+@defun quoted_rfc expression
+Similar, except an embedded double quote would be escaped by
+preceding it with another double quote, as prescribed by RFC 4180.
+@end defun
 
 @defun round
 @value{FIXME:UNDOCUMENTED}
--- a/src/report.cc
+++ b/src/report.cc
@@ -765,7 +765,7 @@ value_t report_t::fn_quoted(call_scope_t& args)
   return string_value(out.str());
 }
 
-value_t report_t::fn_quoted_rfc4180(call_scope_t& args)
+value_t report_t::fn_quoted_rfc(call_scope_t& args)
 {
   std::ostringstream out;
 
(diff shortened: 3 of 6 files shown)


oaf-edwin:
Some history:

- [asked on the Google Group](https://groups.google.com/g/ledger-cli/c/ibnbZzfT0BA/m/BKEtoOPiCwAJ)
- change 221c37a20 (made just a few days later) implemented the suggested "quoted_rfc4180" (but no reply to the Group discussion thread?)
- following year, change edbde4b21 removed the mention of it (and test) from the Info doc
- recently, I noticed this wasn't working, and curiosity drove me to investigate

tbm:
@jwiegley can you review that.

I guess my question would be whether the parser could be fixed.

oaf-edwin:
The parser is not wrong, I believe. The term "quoted_rfc4180" is a valid amount.

I guess we could ponder whether an amount is allowed to appear in a format expression. If it is, then it doesn't seem as though a parser change makes sense.

tbm:
But aren't you calling it as `quoted_rfc4180()`, i.e. with brackets, and that would make it a function?

oaf-edwin:
Oh, I think I see what you mean: instead of complaining "Cannot call an amount as a function," it would defer deciding whether "quoted_rfc4180" is an amount or a function name until it sees the parentheses?

OK, yeah, I suppose that's possible, albeit quite a bit more involved/messy.

tbm:
I don't know anything about the context of your report but it's not obvious to me why `quoted_rfc4180` shouldn't be a valid function name.  Can you show how ledger interprets this as a commodity?

Anyway,  `quoted_rfc4180()` seems like a function to me, so it sounds like a parser issue that should be fixed.

But I don't know.  This is really a question for John... @jwiegley ?

tbm:
@jwiegley any comment on this?

I can merge this but it seems the real problem is in the parser.

jwiegley:
As you can see from the output of `ledger expr 'quoted_rfc4180("hello")'`, the expression parser does not treat this function name as a commodity. If that is happening, it must be due to the context where the function is being used; it's possible that the query parser (a simplified variant of the expression parser) could indeed be doing this, since it's not tuned for the use of functions.

From reading the code changes, it sounds like `%(...)` in a format string is the context where the function name is being misparsed?

Ah, indeed it is:
```
$ ledger format '%(quoted_rfc4180("hello"))'
  quoted_rfc4180("hello")
  ^^^^^^^^^^^^^^^^^^^^^^^
Error: Cannot call an amount as a function
--- Context is first posting of the following transaction ---
2004/05/27 Book Store
    ; This note applies to all postings. :SecondTag:
    Expenses:Books                 20 BOOK @ $10
    ; Metadata: Some Value
    ; Typed:: $100 + $200
    ; :ExampleTag:
    ; Here follows a note describing the posting.
    Liabilities:MasterCard        $-200.00

--- Input format string ---
%(quoted_rfc4180("hello"))

--- Format elements ---
Element:    EXPR  flags: 0x0  min:  0  max:  0  expr: {quoted_rfc4180}("hello")

--- Formatted string ---
"
```
For whatever reason the function name is being interpreted as an amount literal.

As @tbm surmised, this needs to be fixed in the format string parser, and not by changing the name of the function.

tbm:
@oaf-edwin do you think you can investigate the format string code?

@jwiegley any chance you have time to fix this? Maybe it's easy for you.

jwiegley:
Looked deeper and it turns out I'm wrong: even in non-format expression, the presence of numbers causes Ledger to think it's an amount. I don't see a good way around this, so for now I do think we rename the function.

jwiegley:
Please see https://github.com/ledger/ledger/pull/2171

oaf-edwin:
@jwiegley Glad you took a more careful look into it. Your previous comment left me quite befuddled.

> For whatever reason the function name is being interpreted as an amount literal.

The reason is that `quoted_rfc4180` is a perfectly valid amount literal.

```
% python
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ledger
>>> x=ledger.Amount('quoted_rfc4180')
>>> x.commodity.symbol
'quoted_rfc'
>>> x.number()
4180
>>>
```
That, plus the fact that the tokenizer gives precedence to amount literals over identifiers (line 415 of token.cc).

Note that you get the exact same error in the following:

```
% ledger format '%($1.00(payee))'
     ...<snip>...
--- Formatted string ---
"While evaluating value expression:
  $1.00(payee)
  ^^^^^^^^^^^^
Error: Cannot call an amount as a function
```

Ledger allows great flexibility in the format of amounts (specifically the commodity symbol format and placement). The trade-off is that an identifier cannot have digits in it (as long as you're allowing amount literals to appear in expressions).

oaf-edwin:
By the way, there's a related problem elsewhere in the parser. Something like `"ab cd" 17` is also a valid amount. But it can't even be parsed properly in an expression. This time it's because `expr_t::token_t::next()` gives precedence to quoted string before literal amount (token.cc line 219 versus the lines 380ff). Perhaps when it finds a quoted string it ought to look ahead to see if a number follows.

jwiegley:
Really good points, @oaf-edwin. I'll create an issue for the second.

PR #2171, opened 2023-01-23
Rename quoted_rfc4180 to quoted_rfc; disallow numbers in function names

jwiegley:
Fixes https://github.com/ledger/ledger/issues/2007.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -44,6 +44,11 @@
   to include for stdin (`-f -`). Also for `-f -` when `include` cannot find the
   file it reports the error with full path now. (bug #2057 & bug #2092)
 
+- Rename `quoted_rfc4180` to `quoted_rfc`, as numbers used in function names
+  confuses the parser.
+
+- Numbers are no longer permitted in value expression function names.
+
 - Various documentation improvements
 
 ## 3.2.1 (2020-05-18)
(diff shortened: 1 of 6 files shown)


oaf-edwin:
@jwiegley Thanks, John. Added a couple more comments to you in #2168

tbm:
@oaf-edwin is this ok with you?

oaf-edwin:
@tbm Well, I'm just a humble would-be contributor. So of course it's really up to you. But here are a few comments:

1. The essential piece (report.cc and report.h) is the same as what I wrote in the earlier PR. So that's good.
2. However, I had also included relevant new regression tests, which I don't see here.
3. John's changes here touch a commented-out section of ledger3.texi, without uncommenting them. I prefer my earlier version of the changes to that file, which include cleaning up (removing) a pointless reiteration of a "smallexample" (it's pointless because it does not involve escaping of an embedded double quote for either case).
4. John's change in token.cc is probably a good idea. It was not in the way of the main fix needed to produce a working quoted_rfc function, so I hadn't touched that part. And actually I'm not sure whether it has any detectable effect. (A test that demonstrates broken and fixed behavior before and after the change is always reassuring, whenever possible.)

I guess I was a little quick to close the PR that I wrote, and remove the branch on the cloned repo. But I could upload it again if you like (I still have it locally).

jwiegley:
@oaf-edwin I'd prefer to merge in your original PR, if you'd be able to upload the branch and open it again.

oaf-edwin:
@jwiegley OK, I've created #2173. Sorry for the excess churn.

tbm:
I added John's "disallow numbers in function names" patch.

PR #2173, opened 2023-01-24
Rename quoted_rfc4180 to quoted_rfc, to avoid parser quirk

### LINKED ISSUE #2007: CSV output invalid escape sequence

oaf-edwin:
Ledger's expression parser considers quoted_rfc4180 to be an amount (quantity 4180 of commodity "quoted_rfc").

Fixes #2007.

### DIFF
--- a/doc/ledger.1
+++ b/doc/ledger.1
@@ -1284,11 +1284,11 @@ for values that have a per-unit cost.
 .It Fn quoted expression
 Surround
 .Ar expression
-with double-quotes.
-.It Fn quoted_rfc4180 expression
+with double quotes.
+.It Fn quoted_rfc expression
 Surround
 .Ar expression
-with double-quotes, compatible with rfc 4180.
+with double quotes, compliant with rfc 4180.
 .It Sy real
 .\" Is there a difference between real and actual?
 Return true if the transaction is real, i.e not an automated or virtual
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -8518,7 +8518,8 @@ Return the quantity of @var{value} for values that have a per-unit cost.
 @end defun
 
 @defun quoted expression
-Surround @var{expression} with double-quotes.  If expression contains a double-quote, it will be escaped with a backslash.
+Surround @var{expression} with double quotes.  If expression contains
+a double quote, it will be escaped with a backslash.
 @smallexample @c command:EAD8AA7,with_input:3406FC1
 $ ledger -f expr.dat --format "%(quoted(account)) %(quoted(amount))\n" reg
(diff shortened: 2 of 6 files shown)


tbm (approved):
thanks for your patience with us!

jwiegley:
Thank you so much for your persistence and work here, @oaf-edwin. You found a good limitation that I was unaware of.

oaf-edwin:
Thanks, @jwiegley. This turned out to be pretty interesting, didn't it!

It really got me thinking about the tension between notational convenience and expressivity. The design of the journal file format, with its notion of "hard separator" (which, by the way, is sheer genius) means that "amounts" are easy to parse (we always know when to expect them just by looking at the spaces). But it gets really tricky if we want to allow them to appear more generally anywhere in an expression.

Compare it to typical programming language, where it's easy to distinguish an identifier (starts with a letter) from a numeric constant (starts with a digit), or a string constant (demarcated by punctuation).

One thing we might have considered, if we wanted to continue to allow identifiers to have digits, is to require a little more disambiguation for amount literals. For example, we could require people to use a constructor: instead of a bare `EUR42.00` require something like `amount('EUR42.00')`, and maybe you could also allow a "read syntax" shortcut, something like `#(EUR42.00)`. We wouldn't need to require this in journal file transactions, of course; only in expressions.

But I don't think it's really so terrible to disallow digits in identifiers; I'm just kinda mentally noodling here.

jwiegley:
The reason I allow "naked" amounts in expressions is purely for convenience, to be able to say things like:
```
foo($10 + EUR20)
```
Of course, `{}` can always be used to explicitly introduce an amount, but I think that might get cumbersome, giving the number of places value expressions are used in.

PR #2122, opened 2022-07-05
Update utfcpp to 3.2.1

afh:
The local copy of [utfcpp](https://github.com/nemtrif/utfcpp) (v2_0) seems rather old, if not "outdated" given that the latest version is [3.2.1](https://github.com/nemtrif/utfcpp/releases/tag/v3.2.1) released on June 6th, 2021.

ℹ️  The majority of commits in this PR come from the utfcpp subtree; the following commands were used to update utfcpp:

```
% git subtree add --prefix lib/utfcpp/v3 https://github.com/nemtrif/utfcpp.git v3.2.1
% git rm -rf lib/utfcpp/v2_0
```

PR #2180, opened 2023-02-01
Update utfcpp to 3.2.3

afh:
This PR is based on #2122 and replaces it.

ℹ️ The majority of commits in this PR come from the utfcpp subtree; the following commands were used to update utfcpp with the following command having been run…
```
% git subtree pull --prefix lib/utfcpp/v3 https://github.com/nemtrif/utfcpp.git v3.2.3
```

…after the commands from #2122:

```
% git subtree add --prefix lib/utfcpp/v3 https://github.com/nemtrif/utfcpp.git v3.2.1
% git rm -rf lib/utfcpp/v2_0
```


tbm:
I'm not sure how subtrees work.  Will those ~100 commits show up in the commit history of the ledger repo? Is there a way to avoid that? Or is this what we want?

afh:
At first glance it does not seem that the subtree commits show up in ledger's commit history, e.g. when you click on a changeset (SHA) from the subtree in the list of those 100 commits Github shows a page saying that "We couldn’t display that commit for review."

However when listing the commits for the utfcpp repo owner then several commits are listed that at first glance appear to be from the utfcpp repo rather than ledger's: `git log --author 'nemtrif'`.

That said `git subtree add` supports the `--squash option, which "import[s] only a single commit from the subproject, rather than its entire history." [see git-subtree(1)], which seems very reasonable to use.

Let me try and do that.

afh:
Closing in favor of #2182

Issue #2154, opened 2022-12-28
`acprep` doesn't work with Python 3.11


equal-l2:
With python 3.11, `acprep` fails with the following error:
```
acprep: ERROR: Caught exception
Traceback (most recent call last):
  File "/Users/equall2/test/ledger/./acprep", line 196, in run
    argspec = inspect.getargspec(self.main)
              ^^^^^^^^^^^^^^^^^^
AttributeError: module 'inspect' has no attribute 'getargspec'
```

tbm:
@afh @q3aiml

afh:
Thanks for bringing this to our attention, @equal-l2. Can you please verify that #2155 fixes this issue?

Thanks for the ping @tbm :pray:

tbm:
I still get:

```
acprep:198: DeprecationWarning: inspect.getargspec() is deprecated since Python 3.0, use inspect.signature() or inspect.getfullargspec()
```

with

```
sys.version_info(major=3, minor=9, micro=2, releaselevel='final', serial=0)
```

so it seems that warning was already there at least in Python 3.9.

afh:
@tbm I've put up #2183 for review, seems that [`getfullargspec`](https://docs.python.org/3.9/library/inspect.html#inspect.getfullargspec) has been available since Python 3.0…

PR #2183, opened 2023-02-01
acprep: Replace deprecated getargspec with getfullargspec

afh:
Adressing latest comment in #2154.

According to the Python documentation [`getargspec`](https://docs.python.org/3.9/library/inspect.html#inspect.getargspec) is
> [d]eprecated since version 3.0: Use [getfullargspec()](https://docs.python.org/3.9/library/inspect.html#inspect.getfullargspec) for an updated API that is usually a drop-in replacement, but also correctly handles function annotations and keyword-only parameters.

Given that support for Python 2 has been dropped in ledger and [getfullargspec()](https://docs.python.org/3.9/library/inspect.html#inspect.getfullargspec) is available in Python 3 this change seems a good solution forward.

### DIFF
--- a/acprep
+++ b/acprep
@@ -193,10 +193,7 @@ class CommandLineApp(object):
             # not let us differentiate between application errors and a case
             # where the user has not passed us enough arguments.  So, we check
             # the argument count ourself.
-            if sys.version_info < (3, 10):
-                argspec = inspect.getargspec(self.main)
-            else:
-                argspec = inspect.getfullargspec(self.main)
+            argspec = inspect.getfullargspec(self.main)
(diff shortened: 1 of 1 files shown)


tbm (approved):
LGTM!

afh:
Thanks for merging, @tbm. What might be some other issues that would be good to try and get fixed for the next release?

tbm:
@afh I'm not aware of any blockers. I'll try to update NEWS and make a release.

tbm:
Thanks for all your help!

afh:
Thank you, @tbm, for driving this release and your continued maintenance of ledger :pray:

Issue #1932, opened 2020-06-20
docs: mention --python/--no-python


aparkerlue:
Ledger does not seem to recognize the `--import` option after being built with `./acprep update`.

However, the version of Ledger that comes with Ubuntu 18.04 does:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.4 LTS
Release:        18.04
Codename:       bionic
$ apt show ledger
Package: ledger
Version: 3.1.2~pre1+g3a00e1c+dfsg1-5build5
Priority: optional
Section: universe/utils
Origin: Ubuntu
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Original-Maintainer: David Bremner <bremner@debian.org>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 7,888 kB
Depends: libboost-filesystem1.65.1, libboost-iostreams1.65.1, libboost-python1.65.1, libboost-regex1.65.1, libboost-system1.65.1, libc6 (>= 2.14), libgcc1 (>= 1:3.4), libgmp10, libicu60 (>= 60.1-1~), libmpfr6 (>= 3.1.3), libpython2.7 (>= 2.7), libstdc++6 (>= 5.2)
Suggests: elpa-ledger, python-ledger
Homepage: http://ledger-cli.org
Download-Size: 2,257 kB
APT-Manual-Installed: yes
APT-Sources: http://us.archive.ubuntu.com/ubuntu bionic/universe amd64 Packages
Description: command-line double-entry accounting program
 Ledger is a powerful and flexible double-entry accounting system run
 entirely from the command line.  Your accounts ledger is stored in one or
 more plain-text files with a very simple and readable format, and ledger
 does the hard work of balancing your books and reporting.

$ touch m.py
$ ledger --import m.py --version
Ledger 3.1.2-20160801, the command-line accounting tool

Copyright (c) 2003-2016, John Wiegley.  All rights reserved.

This program is made available under the terms of the BSD Public License.
See LICENSE file included with the distribution for details and disclaimer.
```

I tried the same invocation after building each Ledger version since 3.1.2, but they all return an error with `--import`. The fact that `--import` is unavailable when I build 3.1.2 but available in the Ubuntu version of 3.1.2 suggests that I'm not building Ledger quite the right way. How does one build Ledger to recognize the `--import` option?

[...]

PR #2192, opened 2023-02-13
Update readme

afh:
* Correct declared Python minimum required to 3.9
* Mention that some Ledger features require Python (see #1932)
* Change scheme in urls from http to https
* Format markdown table


tbm (approved):
thanks!

PR #2194, opened 2023-02-13
Update required CMake version to 3.12

afh:
While this specific PR only adds little value I'd like to start a conversation about how far back older version of Ledger dependencies we'd like to support.

My first proposal would be to support versions that were released 4 years ago or more recent, any version released before that are unconsidered as unsupported from this project's point of view.
This would allow for a codebase modernisation once a year.

What are your thoughts on this?


tbm:
> My first proposal would be to support versions that were released 4 years ago or more recent, any version released before that are unconsidered as unsupported from this project's point of view

This might be a question for the mailing list.

afh:
Unfortunately the mailing list has been rather unresponsive on the matter, @tbm.

This PR is superseded by #2222.

PR #2196, opened 2023-02-14
Refactor Ledger Python module into Python package

afh:
Allowing to extend the Python Ledger package with pure Python classes and methods as well as adding functionality to existing native classes using Python.

This is a draft PR as I'd like to get a conversation going about the pros and cons of the proposed change.

On the one hand this change will bring a lot of flexibility for future additions written entirely in Python and decouples the Python API for people using the Ledger Python package from what is exposed by Ledger's C++ native code via Boost.Python.

On the other hand this may break backwards compatibility, yet the change to migrate script is fairly straight-forward: Change `import ledger` to `import ledger.core as ledger`.

I'm not using `ledger python` much, so things may break there, for example it was necessary to change the module name in `test/baseline/feat-value_py3.test` from `ledger` to `_core` in order to make the tests pass.
There may be ways around it, I'm not sure yet.


tbm:
Adding @davidsklar in case he has input.

afh:
Closing this PR as as it has served its purpose to explore how to add a Ledger Python Package. I'm working on improving this further, but what work in progess is neither in draft nor review state at the moment.

PR #2197, opened 2023-02-14
Revert to build without Python

afh:
Since I'm _always_ building ledger with Python support I accidentally flipped the default in the flake.nix. Mea culpa :/

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,7 +3,7 @@
 
   nixConfig.bash-prompt = "ledger$ ";
   outputs = { self, nixpkgs }: let
-    usePython = true;
+    usePython = false;
     gpgmeSupport = true;
     forAllSystems = f: nixpkgs.lib.genAttrs systems (system: f system);
     nixpkgsFor = forAllSystems (system: import nixpkgs { inherit system; });


tbm (commented):
is there a reason not to leave it on by default?

afh:
I think the assumption is that more people use ledger without Python support rather than with. Maybe @marsam who made the change in [NixOs/nixpkgs#215289](https://github.com/NixOS/nixpkgs/pull/215289) can chime in?

elamperti:
As a user, I would like it to stay enabled. My main reason for not using Python is the lack of documentation (I ended up parsing ledger's output with nodejs). Having to re-compile it or find a package with Python enabled would definitely be another adoption impediment 😕

afh:
Thanks for chiming in @elamperti, I've been down that road although I used ruby. I feel your pain in regards to lack of documentation of the Python API (hopefully I can remedy that at some point). Happy to share some of my experience with you if you are willing to share specific use cases.

Please note, that this PR would disable Python by default only for Nix flake users, using ledger's flake directly, e.g. `nix shell github:ledger/ledger`.

Out of curiosity how do you install and use ledger?

afh:
Closing due to lack of activity and interest.

PR #2206, opened 2023-03-02
Do not perform commodity reduction when parsing a format directive

### LINKED ISSUE #2205: Regression when specifying format for commodity "h"
We’re using `ledger` as time management tool and have added the following lines to allow appropriate German formatting of hours:

```
commodity h
    format 1.000,00 h
    default
```

When upgrading to ledger 3.3.0, running any ledger command results in the error message:

    While parsing file "/time/user.journal", line 722:
    Error: commodity directive symbol h and format directive symbol s should be the same
    
This seems to be the same error as [this one reported on reddit](https://www.reddit.com/r/plaintextaccounting/comments/10xgg4x/comment/j95k8y7/?utm_source=share&utm_medium=web2x&context=3) which allocates the error message in https://github.com/ledger/ledger/blame/master/src/textual.cc#L1137

This is a regression from 3.2.1 and older where the above statement worked perfectly fine.

jwiegley:
Fixes #2205

### DIFF
--- a/src/textual.cc
+++ b/src/textual.cc
@@ -1132,10 +1132,12 @@ void instance_t::commodity_format_directive(commodity_t& comm, string format)
   // observational formatting.
   trim(format);
   amount_t amt;
-  amt.parse(format);
+  amt.parse(format, PARSE_NO_REDUCE);
   if (amt.commodity() != comm)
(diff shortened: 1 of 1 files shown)


tbm (approved):
Thanks John!  This test case, and another one I tried, works.  I'll add those test cases.

PR #2217, opened 2023-03-20
Build the Nix flake in CI

purcell:


purcell:
@afh This should at least make sure that the Nix flake doesn't get quietly broken.

afh:
Right, I always forget about flake.loc… I couldn't agree more on regularly updating the nix flake and ideally ledger dependencies. Do you have an opinion on [my message to the mailing list](https://groups.google.com/g/ledger-cli/c/8CZtMmOVmLI)?

tbm (approved):
thanks

PR #2218, opened 2023-03-20
flake: Run pre and post hooks in checkPhase

afh:


tbm (approved):
I have no idea what this is doing but I guess you know what you're doing

afh:
Thanks for merging, @tbn :pray:

Nixpkgs builds packages in phases, e.g. `configurePhase`, `buildPhase`, `installPhase`, etc. Each of these phases can be override (ledger's flake overrides the `checkPhase`) and fine-tuned by defining pre- and postHooks which are executed and the start and end of a phase.

To quote the [Nixpkgs documentation about Phases](https://nixos.org/manual/nixpkgs/stable/#sec-stdenv-phases):
> When overriding a phase, for example installPhase, it is important to start with runHook preInstall and end it with runHook postInstall, otherwise preInstall and postInstall will not be run. Even if you don’t use them directly, it is good practice to do so anyways for downstream users who would want to add a postInstall by overriding your derivation.

So the changes in this PR ensure that nixpkgs best-practices are followed.

PR #2224, opened 2023-03-23
Change a use of is_realzero to just is_zero

jwiegley:
This is needed to avoid a possible divide by zero.


afh:
Nice! Thanks, @jwiegley 🙏

PR #2225, opened 2023-03-28
Fix Python version check

afh:
I got the version check for Python wrong when updating ledger's dependencies and starting to require Python 3.9.

We want cmake to check that the available Python version is greater than or equal to ledger's minimun required Python version.


tbm (approved):
Thanks

afh:
Thanks for merging, @tbm 👍

PR #2232, opened 2023-04-11
Cmake doxygen cleanup

afh:
I came across the [doxygen_add_docs](https://cmake.org/cmake/help/latest/module/FindDoxygen.html?highlight=finddoxygen#command:doxygen_add_docs) CMake function _"intended as a convenience for adding a target for generating documentation with Doxygen"_ and thought it would be a good modernization of `doc/CMakeListst.txt`.


jwiegley (approved):

afh:
Thanks for the review and approval, @jwiegley 🙂

PR #2238, opened 2023-04-15
Cleanup system defines

afh:
the idea is that any features specific to ledger are defined with a value, e.g. `#define HAVE_GETTEXT 0` and "external" features, like compiler specifics are checked using `#ifdef _GLIBCXX_THROW` or `#if !defined(__clang__)`

### DIFF
--- a/src/history.cc
+++ b/src/history.cc
@@ -467,13 +467,6 @@ commodity_history_impl_t::find_price(const commodity_t& source,
 
   const commodity_t * last_target = &target;
 
-#if REVERSE_PREDECESSOR_MAP
-  typedef tuple<const commodity_t *, const commodity_t *,
-                const price_point_t *> results_tuple;
-  std::vector<results_tuple> results;
-  bool results_reversed = false;
-#endif
-
   vertex_descriptor v = tv;
   for (vertex_descriptor u = predecessorMap[v];
        u != v;
@@ -494,23 +487,7 @@ commodity_history_impl_t::find_price(const commodity_t& source,
     const commodity_t * u_comm = get(namemap, u);
     const commodity_t * v_comm = get(namemap, v);
 
-#if REVERSE_PREDECESSOR_MAP
-    if (v == tv && u_comm != last_target && v_comm != last_target)
-      results_reversed = true;
-
-    results.push_back(results_tuple(u_comm, v_comm, &point));
-  }
-
-  if (results_reversed)
-    std::reverse(results.begin(), results.end());
-
-  foreach (const results_tuple& edge, results) {
-    const commodity_t *  u_comm = edge.get<0>();
-    const commodity_t *  v_comm = edge.get<1>();
-    const price_point_t& point(*edge.get<2>());
-#else
     assert(u_comm == last_target || v_comm == last_target);
-#endif
 
     bool first_run = false;
     if (price.is_null()) {
--- a/src/history.cc
+++ b/src/history.cc
@@ -467,7 +467,7 @@ commodity_history_impl_t::find_price(const commodity_t& source,
 
   const commodity_t * last_target = &target;
 
-#if defined(REVERSE_PREDECESSOR_MAP)
+#if REVERSE_PREDECESSOR_MAP
   typedef tuple<const commodity_t *, const commodity_t *,
                 const price_point_t *> results_tuple;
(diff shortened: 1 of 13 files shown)


src/history.cc
```diff
@@ -494,7 +494,7 @@ commodity_history_impl_t::find_price(const commodity_t& source,
     const commodity_t * u_comm = get(namemap, u);
     const commodity_t * v_comm = get(namemap, v);
 
-#if defined(REVERSE_PREDECESSOR_MAP)
+#if REVERSE_PREDECESSOR_MAP
```
afh, line 497: @jwiegley can you share some context on this define, which I wasn't able to find anywhere else in the codebase?

What is its intent?
Can it be removed?

src/utils.h
```diff
@@ -45,30 +45,8 @@
...
-#define TRACING_ON  0
-//#define NO_LOGGING  1
-#else
-#define DEBUG_ON    0
-#define VERIFY_ON   0
-#define TRACING_ON  1           // use --trace X to enable
-#endif
-
-/*@}*/
-
```
afh, line 71: I couldn't explain why these defines are set in `src/utils.h` as opposed to `src/system.hh.in`—except maybe for pre-compilation reasons—and thought it would be a good clean-up to keep all ledger related defines used by several source files in system.hh.in.

Can you think of any issues with this changes?

src/history.cc
jwiegley: I can't find mention of it anywhere, not even Google. So yes, I think it can be removed.

src/utils.h
jwiegley: They aren't in system.hh because they are system-level defines, they are specific to Ledger. System.hh really only exists to collect all of the non-Ledger code we want to import into a single PCH.

afh: I'll figure something out and come back with another proposal, if that's alright with you?

src/history.cc
afh: Could it be related to [Boost's `dijkstra_shortest_paths`](https://www.boost.org/doc/libs/1_72_0/libs/graph/doc/dijkstra_shortest_paths.html)?

src/utils.h
afh: @jwiegley What do you think about introducing `ledger.hh` (generated by CMake from `ledger.hh.in` similar to `system.hh.in`) that provides the version information as other application specific defines for the rest of the source (see 1fe7d71cf130074f2b345ebe70f9e76e8899b072)?

jwiegley: Yes, that sounds like a good approach to me.

jwiegley: `system.hh` really exists for only one purpose: to create the PCH. So we should avoid putting anything meaningful in there.

afh: I went ahead an [`DEBUG_MODE` and `NO_ASSERTS`](https://github.com/ledger/ledger/pull/2238/files#diff-bff4a86f314c161c10f267650141495d4921fd839bc8bfb1583d2a2b53a4585bR67-R68) should moved into [`ledger.hh.in`](https://github.com/ledger/ledger/pull/2238/files#diff-050964adfe522ee4701297304e4e126c6eef2bf6787cee3fea45dfb44d24b34c), @jwiegley, and took the liberty to re-order the system.hh includes to something that felt a bit more useful to me.

Please let me know if there is anything that you'd like to see changed before approving this PR, so it can be merged.

src/history.cc
jwiegley: Oh, that does ring a bell...

jwiegley (approved):

src/history.cc
afh: In your mind what is a better way forward, @jwiegley?
Leave the code parts referencing `REVERSE_PREDECESSOR_MAP` as is or delete them and make a not of it, so that should you or someone else familiar remember the intent can provide context and details and it can be implemented as desired or necessary?

jwiegley: It was introduced in 72fc1824d01bb4fe50405ed183afb57b0e129d69, but was apparently just an idea and never used. I'd say it's safe to delete that code.

afh: ✅ Done 😃

jwiegley (approved):

PR #2222, opened 2023-03-23
Update dependencies

afh:
This PR supersedes #2194.

As per the [conversation on the mailing list regarding updating the Ledger dependencies regularly](https://groups.google.com/g/ledger-cli/c/8CZtMmOVmLI/m/Fm25oWOpAAAJ?pli=1) I took the liberty to update all major dependencies as follows:

Dependency | From | Age | To  | Release Date
--------------------|------|-----|-----|-------------
[CMake][] | 3.0.0 | 4 years (Dec 2018) | 3.16.2 | Dec 2019
[Boost] | 1.49 | 11 years (Feb 2012) | 1.72.0 | Dec 2019
[Gmp] | 4.2.2 | 16 years (Sep 2007) | 6.2.1 | Dec 2016
[Mpfr] | 2.4.0 | 14 years (Jan 2009) | 4.0.2 | Jan 2019
[Python] | 3.0 | 13 years (Feb 2010) | 3.9 | Oct 2020
[Gpgmepp] | — | — | 1.13.1 | Jun 2019
[utfcpp] | 2.3.4 | 8 years (Sep 2015) | 3.2.3 | Dec 2022¹

¹ Since utfcpp is included in the source and its version is currently not checked by the CMakeLists.txt I felt it was acceptable to bump this to a more recent than 4 years release.

[CMake]: https://github.com/Kitware/CMake/releases/
[Boost]: https://sourceforge.net/projects/boost/files/boost-binaries/
[Gmp]: https://gmplib.org/download/gmp/
[Mpfr]: https://www.mpfr.org/history.html
[utfcpp]: https://github.com/nemtrif/utfcpp/releases
[Python]: https://www.python.org/downloads/
[Gpgmepp]: https://www.gnupg.org/ftp/gcrypt/gpgme/
[gettext]: https://ftp.gnu.org/pub/gnu/gettext
[doxygen]: https://github.com/doxygen/doxygen/releases
[graphviz]: https://www.graphviz.org/download/source/
[lcov]: https://github.com/linux-test-project/lcov.git
[libedit]: https://thrysoee.dk/editline/
[Texinfo]: https://ftp.gnu.org/gnu/texinfo/

PR #2240, opened 2023-04-16
gpgme: Remove code handling versions prior to required version

afh:
Since ledger now required Gpgme >= 1.13.1 (see #2222 for details) code handling prior versions can be removed.

/cc @spacefrogg

### DIFF
--- a/src/gpgme.cc
+++ b/src/gpgme.cc
@@ -136,15 +136,9 @@ shared_ptr<Data> decrypted_stream_t::decrypt(shared_ptr<Data> enc_d) {
     ctx = nullptr;
     dec_d = enc_d;
   } else {
-#if GPGME_VERSION_NUMBER < 0x010d00
-    ctx = unique_ptr<Context>(Context::createForProtocol(enc_d->type() == Data::PGPEncrypted
-                                                          ? Protocol::OpenPGP
-                                                          : Protocol::CMS));
-#else
     ctx = Context::create(enc_d->type() == Data::PGPEncrypted
                           ? Protocol::OpenPGP
                           : Protocol::CMS);
-#endif
     if (!ctx)
       throw runtime_error("Unable to establish decryption context");
 
@@ -167,11 +161,7 @@ static inline void init_lib() {
 }
 
 static inline void rewind(Data * d) {
-#if GPGME_VERSION_NUMBER < 0x010c00
-  d->seek(0, SEEK_SET);
-#else
   d->rewind();
-#endif
 }
 
 istream* decrypted_stream_t::open_stream(const path& filename) {
--- a/src/gpgme.cc
+++ b/src/gpgme.cc
@@ -160,10 +160,6 @@ static inline void init_lib() {
     throw_(runtime_error, _f("%1%: %2%") % err.source() % err.asString());
 }
 
-static inline void rewind(Data * d) {
-  d->rewind();
-}
-
 istream* decrypted_stream_t::open_stream(const path& filename) {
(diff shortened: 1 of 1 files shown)


src/gpgme.cc
```diff
@@ -167,11 +161,7 @@ static inline void init_lib() {
 }
 
 static inline void rewind(Data * d) {
```
spacefrogg, line 163: Would you mind dropping `rewind()` entirely and replacing all calls to it with `d->rewind()`? The API differences between GPGME versions were the reasons for its existence.

src/gpgme.cc
```diff
@@ -193,15 +179,14 @@ decrypted_stream_t::decrypted_stream_t(path& filename)
...
-  rewind(dec_d.get());
+  auto dec_d = decrypt(enc_d).get();
 
   if (is_encrypted(enc_d)) {
     fclose(file);
     file = nullptr;
   }
 
-  set_rdbuf(new data_streambuffer_t(*dec_d.get()));
+  set_rdbuf(new data_streambuffer_t(*dec_d));
```
afh, line 189: I took the liberty of simplifying this code some more, does this still look correct to to @spacefrogg?

I seemed that since previously on line 197 and 204 only `.get()` was used the return value could be stored temporarly and used to call `rewind` and then pass to the `data_streambuffer_t` ctor.

What are your thoughts on this?

src/gpgme.cc
afh: Excellent suggestion, thanks!

src/gpgme.cc
spacefrogg: Please rename `dec_d` to `dec_data`, as it is no longer of the same type as `enc_d`.

You also dropped the call to `rewind()` which is crucial before creating a new stream buffer from it.

afh: Apologies for having missed the accidential removal of the `rewind` call. That's why we have PR review :)

src/gpgme.cc
```diff
@@ -183,15 +179,15 @@ decrypted_stream_t::decrypted_stream_t(path& filename)
 
   file = open_file(filename);
   auto enc_d = setup_cipher_buffer(file);
-  dec_d = decrypt(enc_d);
-  rewind(dec_d.get());
+  auto dec_d = decrypt(enc_d).get();
```
spacefrogg, line 182: I guess my previous comment got lost, because updates to comments do not get communicated via e-mail.
This still doesn't work. `dec_d` is a class member and you just made it a local variable. We must keep the shared pointer to the decrypted Data for the life time of the stream. So, please remove the `auto dec_d` declaration again and just keep the assignment.

spacefrogg: It should look like before, i.e.:
```cpp
dec_d = decrypt(enc_d);
dec_d.get()->rewind();
...
...(new data_streambuffer_t(*dec_d.get()));
```

afh: Thank you for you thorough review and helpful comments, I should've avoided to do needles code-optimisations… 🤷

Anyhow I think the PR is now as you requested it to be.

spacefrogg (approved):

afh:
Thanks for reviewing and approving the changes and thanks for contributing support for encrypted journals to ledger, it's an A+ feature👌

In order to be able to merge this PR it needs an approval from @ledger/ledger-hackers.

jwiegley (approved):

afh:
Thanks for reviewing, approving, and merging this, @jwiegley :+1:

PR #2242, opened 2023-04-16
Prepare for language translations

afh:
This PR adds support for international languages by fully enabling gettext support in Ledger.

Since this is a larger PR I'll tag is as a draft, so folks can chime in and provide feedback.

The changes are widespread throughout the codebase as the way that Ledger's creates its output messages had to change. Please review the inline comments on this PR for questions, details of note, and further context on the thinking that went into this PR.

Some of Ledgers messages were pieced together in output streams, yet to better support translations to languages that have a grammar that is very different from English or similar Roman languages the use of format strings is preferred. Other messages would benefit from comments made available to translators as to provide
more context about the message and its placeholder parameters.

Example:

| | `src/draft.cc`
| --- | ---
| **Before** | `out << _("Date:       ") << *date << std::endl;`
| **After** | `out << _f("Date:       %1%") % *date << std::endl;`

☝️P.S.: This is the result of a few days work on many areas of the code, hence I kindly request thorough review should something have slipped my attention, while wrangling all the necessary changes to make this a reality.

✌️P.P.S.: separate PR will follow that adds the first international language to Ledger.

Closes #1241

PR #2243, opened 2023-04-16
German translation

afh:
This PR adds a very first (read: bad) translation of Ledger's messages to :de:. As a native speaker I should be able to do a better job (and will). Yet my focus for this PR was to get an impression on all (~500 😮‍💨) the Ledger messages and the infrastructure to support international languages in Ledger.

I will look for inspiration from [translationproject](https://translationproject.org), oftentime when developing (and thinking) in English it is hard for me to switch back and forth to German when trying to come up with good translation.

Please see below for inline comments and context about the changes.

One thing to emphasise are the changes related to handling special metadata tags, e.g. `Payee` und `UUID`. Ledger currently allows these to be translated (and it does make sense) yet for compatibility I believe ledger should always recognize and support the English version.

This is a companion PR that stacks on top of #2242.

Tested using nix 🥳:tada:

```
% nix run . -- --version
Ledger 3.3.2, the command-line accounting tool
with support for gpg encrypted journals and with Python support

Copyright (c) 2003-2023, John Wiegley.  All rights reserved.

This program is made available under the terms of the BSD Public License.
See LICENSE file included with the distribution for details and disclaimer.
% /usr/bin/env LC_MESSAGES=de nix run . -- --version
Ledger 3.3.2, das kommandozeilenbasierte Buchführungsprogramm
mit Unterstützung für gpg verschlüsselte Journale und mit Unterstützung für Python

Copyright (c) 2003-2023, John Wiegley.  Alle Rechte vorbehalten.

Dieses Programm wird unter den Bedinungen der BSD Public License zur Verfügung gestellt.
Die in der Distribution enthaltente LICENSE Datei enthält weitere Details und die Ausschlussklausel.
```

## TODO
- [ ] Add tests
- [ ] Improve translation

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -1,0 +1,2393 @@
+# German translations for Ledger package.
+# Copyright (C) 2023 John Wiegley
+# This file is distributed under the same license as the Ledger package.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Ledger 3.3.2\n"
+"Report-Msgid-Bugs-To: https://bugs.ledger-cli.org\n"
+"POT-Creation-Date: 2023-04-16 15:36+0200\n"
+"PO-Revision-Date: 2023-04-15 10:04+0200\n"
+"Last-Translator: Alexis Hildebrandt\n"
+"Language-Team: none\n"
+"Language: de\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+
+#: times.cc:1410 times.cc:1429
+#, boost-format
+msgid "   range: %1%"
+msgstr "   Bereich: %1%"
+
+#: times.cc:1412 times.cc:1431
+#, boost-format
+msgid "   start: %1%"
+msgstr "   Beginn: %1%"
+
+#: times.cc:1414 times.cc:1433
+#, boost-format
+msgid "  finish: %1%"
+msgstr "  Ende: %1%"
+
+#: format.cc:256
+msgid "%$ field reference must be a digit from 1-9"
+msgstr "%# Feldverweis muss eine Zahl von 1-9 sein"
+
+#: format.cc:268
+msgid "%$ reference to a non-existent prior field"
+msgstr "%# verweist auf ein nicht-existents vorheriges Feld"
+
+#. TRANSLATORS:
+#. %1% signifies desc
+#. %2% signifies a filename
+#: item.cc:596
+#, boost-format
+msgid "%1% from \"%2%\""
+msgstr "%1% aus \"%2%\""
+
+#. TRANSLATORS: %1% signifies desc
+#: item.cc:588
+#, boost-format
+msgid "%1% from streamed input:"
+msgstr "%1% aus streamed Eingabe:"
+
+#: parser.cc:116 parser.cc:140 parser.cc:157 parser.cc:198 parser.cc:228
+#: parser.cc:295 parser.cc:327 parser.cc:354 parser.cc:380 parser.cc:389
+#: query.cc:254 query.cc:350 query.cc:379 query.cc:403
+#, boost-format
+msgid "%1% operator not followed by argument"
+msgstr "%1% Operator nicht von einem Argument gefolgt"
+
+#. TRANSLATORS: This is an error message when the gpgme encryption library
+#. could not be initialized.
+#. %1% refers to the source of the error
+#. %2% refers to the actual error as a string.
+#: gpgme.cc:170
+#, boost-format
+msgid "%1%: %2%"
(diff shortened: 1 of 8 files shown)


src/post.cc
```diff
@@ -126,6 +126,8 @@ string post_t::payee_from_tag() const
 {
   if (optional<value_t> post_payee = get_tag(_("Payee")))
     return post_payee->as_string();
+  else if (optional<value_t> post_payee = get_tag("Payee"))
+    return post_payee->as_string();
```
afh, line 130: Assume someone has a Ledger journal and has made use of the special metadata tag `Payee`. With the changes in this PR they are able to make Ledger <HeavyGermanAccent>speak German</HeavyGermanAccent>.

Without [this change](https://github.com/ledger/ledger/pull/2243/commits/ac4f561fa2586fe0545abf4344474812b5aad951) their journal would no longer recognize the English `Payee` keyword they have in their journal as Ledger would look for the German equivalent, i.e. `Empfänger`.

Therefore I believe Ledger should always support and recognize the English name of special metadata tags.

po/de.po
```diff
@@ -0,0 +1,2393 @@
+# German translations for Ledger package.
```
afh, line 1: Several words like `commodity` have not (yet) been translated. This was already quite a bit of work as it is and my brain went 👻 while thinking in English when working on this PR and switching to translation mode when expanding on this file.

jwiegley:
Ich glaube wirklich nicht, dass ich qualifiziert bin, die Übersetzungen speziell zu überprüfen, obwohl ich die Codeänderungen gutheiße. Vielleicht könnte ein anderer Muttersprachler seine Zustimmung geben, wenn es fertig ist?

afh:
Ich wusste gar nicht, daß Du so gut Deutsch sprichst, John! Heißt das, daß ich die PR Beschreibungen zukünftig komplett in Deutsch schreiben kann? 😅

jwiegley:
Ach, es gab eine Zeit, in der ich das genossen hätte. Während meiner Highschool- und Collegejahre habe ich mit großer Intensität Deutsch gelernt und wollte tatsächlich nach Deutschland ziehen, um dort als Programmierer zu arbeiten. Doch dann vergingen die Jahre und es gab zu wenig Gelegenheiten, die Sprache hier in den Staaten anzuwenden. Und realistisch betrachtet ist es viel effizienter, wenn Deutsche mit mir auf Englisch kommunizieren, als wenn ich meine brüchigen Sprachkenntnisse an ihnen ausprobiere. Und so habe ich am Ende den Großteil meiner Fähigkeiten verloren. Glücklicherweise ist ChatGPT ein großartiger Assistent, der mir hilft, die Lücken zu füllen, wo das Vergessen mich eingeholt hat, und zumindest kann ich das Ergebnis Korrektur lesen und sicherstellen, dass ich nicht wie ein verrückter Mensch klinge.

And so yes, much better to continue in English, to avoid the inefficiency of translation getting in the way of a timely response. :)

afh:
I came across an article ["[o]n writing better error messages"](https://onlineornot.com/write-better-error-messages) recently and feel somewhat inspired to re-work the existing messages Ledger prints.

My thinking is around coming up with a "scheme" for message-categories, i.e. how can error messages be differentiated linguistically from info messages.

How would you feel about a PR that may or may not rewrite the majority of Ledger's messages, @jwiegley?


ℹ️ The article's essence is:

* **Who** caused the error?
* **What** happened, and why?
* **When** will it be fixed?
* **How** can the user respond to the error?

jwiegley:
Can you show me a few examples? That would better help me determine if the heavy lift is worth your time...

afh:
That is very considerate, @jwiegley, thank you. I believe it is worth the time to do soon, i.e. before folks volunteer to translate into other languages, as such a potentially large scale change will effectively require the re-translation.

Here are some quick examples:

| Current Message | Possibly Improved Message
| --- | ---
| Cannot add an uninitialized amount to an amount | Initialize the uninitialized amount %1% with a value to add it to the amount %2%
| Cannot add two uninitialized amounts | Initialize the uninitialized amounts with a value to add them together
| Cannot 'not' %1% | The 'not' operator could not be applied to %1%
| Error in the pager | The pager program %1% exited with error status %2%

And maybe a first good take is consolidating use of 'Cannot' vs' Could not' and proper quoting of arguments within messages.

I just had an interesting idea: We could use gettext to get a feel for rewording by simply providing po/en.po with different messages 😆

jwiegley:
The latter two are definitely an improvement, but I find the top two to be more confusing. They don't directly inform the user about the mistake they made. By rephrasing it in terms of what would have worked, it can make the actual problem harder to spot...

jwiegley:
Oh yes, that sounds great, Alexis! Good work you're doing here, and I think it will be helpful to many once we hammer it out.

jwiegley:
Closing draft after much time. Please re-open when work is ready to resume!

afh:
Thanks for the clean-up! 👍

PR #2250, opened 2023-04-25
Fix compilation errors with NO_ASSERT

afh:
See inline comments for details.

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -44,6 +44,7 @@
 #pragma once
 
 #include <boost/uuid/detail/sha1.hpp>
+#include <ledger.hh>
 
 #define TIMERS_ON   1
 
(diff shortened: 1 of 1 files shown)


src/utils.h
```diff
@@ -44,6 +44,7 @@
 #pragma once
 
 #include <boost/uuid/detail/sha1.hpp>
+#include <ledger.hh>
```
afh, line 47: `ledger.hh`. contains the definition of NO_ASSERT, without its inclusion the build fails at the linker stage with missing symbols ot `ledger::debug_assert`

src/utils.h
```diff
@@ -94,7 +95,7 @@ namespace ledger {
 
 #else // !NO_ASSERTS
 
-#define assert(x)
+#define assert(x) ((void)(x))
```
afh, line 98: The Python API uses the assert macro in several places (e.g. [`main/Objects/unicodeobject.c:118`](https://github.com/python/cpython/blob/main/Objects/unicodeobject.c#L118).

I'm uncertain whether this change is the best change possible, yet it allows the Ledger to compile successfully.

jwiegley (approved):

PR #2255, opened 2023-05-22
doc: Fix ledger3.texi:4306: unknown command `Chevron'

afh:


tbm (approved):
Doh!  Thanks for fixing this.

PR #2258, opened 2023-06-05
Use OpenBSD readline instead of editline

madroach:
OpenBSD ships with GNU readline in its base system. There is no port for NetBSD editline. Nevertheless ledger's configure system recognises the GNU readline library and sets `HAVE_EDIT`.
The OpenBSD ledger port deals with this by explicitely disabeling libedit support:
``` make
# libedit support requires readline headers that aren't installed on OpenBSD
CONFIGURE_ARGS +=	-DHAVE_EDIT:=Off
```
I did not test it, but ledger links fine against OpenBSD's libedit when `main.cc` knows the right headers, as done in this PR. This is a hack, because I don't test for the feature, but for the OS.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -40,7 +40,12 @@
 #endif
 
 #if HAVE_EDIT
-#include <editline/readline.h>
+  #ifdef __OpenBSD__
+    #include <readline/readline.h>
+    #include <readline/history.h>
+  #else
+    #include <editline/readline.h>
+  #endif
 #endif
 
 using namespace ledger;


afh:
Thanks for this PR, @madroach, let me try those changes on my OpenBSD workstation and get back to you.

afh:
@madroach Looking more closely at the issue I was wondering whether addressing this issue within cmake would allow to also support GNU readline alongside OpenBSD's libedit.
The result is PR #2259, what are your thoughts on this approach?

afh:
@madroach Are you okay with closing this in favor of #2259?

madroach:
Sure. Thanks for your efforts :)

PR #2265, opened 2023-06-17
Let dependabot help us keep Actions config up to date

purcell:


afh:
Thanks for the suggestion @purcell. Not being familiar with Dependabot for GitHub Actions, can you please share links to related documentation and use-cases that you think ledger will benefit from?
This would help me understand better how this can be used. 🙏

purcell:
Essentially, with this config Dependabot will automatically open PRs in the repo to bump the versions of GitHub Actions we use here, e.g. `install-nix-action`. I've been doing this across my repos today exactly because v20 of that action is currently broken on MacOS, and v22 is fixed — Dependabot just makes such updates easy to apply.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot

afh (approved):
Automatically created PR to update GitHub Actions you say? I'm all for it!! 🥳

afh:
Thanks for making this change, @purcell. I took the liberty to apply the same changes to [ledger/ledger.github.io](https://github.com/ledger/ledger.github.io), so that things are up-to-date there as well.

purcell:
Oh, good thinking. I'd already done it for `ledger-mode` too. 😄

afh:
Excellent, @purcell! :tada:

@alerque is this something that [`ledger/vim-ledger`](https://github.com/ledger/vim-ledger) could possibly benefit from too?

PR #2271, opened 2023-07-13
Fix null posting causing error on balanced multi-commodity transaction

### LINKED ISSUE #2270: multiple commodities causes error: There cannot be null amounts after balancing a transaction
I have a journal like so:
```
commodity USD
  format 0.00 USD ; error
  ;format 0.000 USD ; no error

2023-04-01
  acc0  0.123 BTC @ 30508.00 USD
  acc1  -3752.48 USD
  fee  0.0002 BTC
  acc0  -0.0002 BTC
  adjustments
```
I expect ledger to figure out that the adjustments account should be deducted 0.004 USD, but instead I get:
```
While parsing file "tmp.dat", line 10:
Error: There cannot be null amounts after balancing a transaction
```
The error message goes away if I change the precision of USD to three decimal places. This does not make sense to me since ledger is supposed to internally use infinite-precision rational values.

And anyway, it would be nice if ledger could resolve a null value to zero instead of giving an error. I don't always want to have to figure out whether it will be zero to know whether I need to add an adjustments posting. I want to just add an adjustments posting whenever it may be necessary, and if it isn't used then great.

dbear496:
This fixes #2270 . The cause of the issue was balancing postings were not properly generated when the transaction balance was already zero.

### DIFF
--- a/src/balance.cc
+++ b/src/balance.cc
@@ -273,7 +273,7 @@ balance_t::strip_annotations(const keep_details_t& what_to_keep) const
 void balance_t::sorted_amounts(amounts_array& sorted) const
 {
   foreach (const amounts_map::value_type& pair, amounts)
-    if (pair.second)
+    if (! pair.second.is_null())
       sorted.push_back(&pair.second);
   std::stable_sort(
     sorted.begin(), sorted.end(),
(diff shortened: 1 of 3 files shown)


dbear496:
I am not sure whether my change to filter.cc is necessary. It passes all tests either way.

jwiegley (approved):

jwiegley:
It all looks like a safer and more robust form of the logic, so I'm good with it if it passes all tests.

PR #2272, opened 2023-07-14
copy cost from automatic transaction postings

### LINKED ISSUE #2268: trading between commodities in automatic transaction does not balance
When an automatic transaction deals with trading between commodities -- even with an explicit price -- the transaction will not balance. For example, the following does not balance when the automatic transaction is applied even though 100 apples at $0.20 per apple is clearly $20.

```
= expr 'tag("thetag") == "foo"'
  Assets:My Larder           100 apples        @ $0.200000
  Assets:Checking           $-20

2010/05/31 Farmer's Market
  Assets:My Larder           100 apples        @ $0.200000
  Assets:My Larder           100 pineapples    @ $0.33
  Assets:My Larder           100 "crab apples" @ $0.04
  Assets:Checking
    ; thetag: foo
```

dbear496:
This fixes #2268 by preserving the cost from automatic transactions and reflecting it in target transactions.

### DIFF
--- a/src/xact.cc
+++ b/src/xact.cc
@@ -782,6 +782,8 @@ void auto_xact_t::extend_xact(xact_base_t& xact, parse_context_t& context)
         // the automated xact's one.
         post_t * new_post = new post_t(account, amt);
         new_post->copy_details(*post);
+        if(post->cost)
+          new_post->cost = post->cost;
 
         // A Cleared transaction implies all of its automatic posting are cleared
         // CPR 2012/10/23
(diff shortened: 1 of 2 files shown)


dbear496:
It is possible that copying the cost should go in post_t::copy_details instead of in xact_t::extend_xact. But it is not entirely clear to me what the copy_details method is supposed to do. It really begs the question, "which details?" Why does it not already copy things like cost, assigned_amount, account, etc.? Also, I do not really understand why copy_details is called from the copy constructor--that appears redundant.

dbear496:
Hey, I'm not sure why this PR never got reviewed and also PR #2274 , but I'd like to bring them to the surface. @afh

jwiegley (approved):

PR #2277, opened 2023-08-12
Update readme

monkeywithacupcake:
Makes the cli code copy/pastable by removing the leading `$` and adds a note/warning for macOS users regarding python / python3

This is doc only (only touching readme)

[ci skip]

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,12 +25,17 @@ and easy to use.
 I know, you just want to build and play.  If you have all the [dependencies](#dependencies)
 installed, then simply do this:
 
-    $ git clone git@github.com:ledger/ledger.git
-    $ cd ledger && ./acprep update  # Update to the latest, configure, make
+```
+git clone git@github.com:ledger/ledger.git
+cd ledger && ./acprep update  # Update to the latest, configure, make
+```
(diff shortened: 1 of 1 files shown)


afh:
Thanks for this PR, @monkeywithacupcake. I think it is worthwhile to begin each code block with:

```
```sh
```

README.md
```diff
@@ -103,52 +115,63 @@ platforms (for example, `linux-64`), which can be used with [Conda] or [mamba].
...
 The next step is preparing your environment for building.  While you can use
 `cmake .` and make, I've prepared a script that does a lot more of the
 footwork for you:
 
-    $ ./acprep update
+```
+./acprep update
+```
     # or, if you want to use the Boost libraries with suffix -mt, install in
     # $HOME/local and build with 2 processes in parallel
```
afh, line 170: Please unindent these 2 lines and remove the leading `#`, so they appear as regular text.

afh (changes requested):
@monkeywithacupcake Please update PR, so that only `$ ` is removed from the code examples instead of enclosing them in a ```-block?

PR #2280, opened 2023-08-22
Support regex in aliases

badbob:


badbob:
This pull request can not be merged as is, because the latency of `ledger` increased **A LOT**. 
This need to be refactored and optimized by using [RE-tree approach](https://citeseerx.ist.psu.edu/document?repid=rep1&type=pdf&doi=56abfcbe30e55691e765341733ac06f5ad850427) or something else.

afh:
@badbob Is this something you'd be willing to continue pursuing?

jwiegley:
One improvement I can see right away is to use a regex object for the account keys, rather than a string. The current implementation re-builds the reegex object (`std::regex`) every time a lookup is performed, for every account in the map until the one that matches. This is a large amount of repeated work that could be avoided.

PR #2288, opened 2023-09-05
Add support for Ubuntu 22.04 Jammy in ./acprep

burntraisin:
Running `./acprep dependencies` in Ubuntu 22.04 Jammy while building ledger from source returns:

    acprep: INFO: Invoking primary phase: dependencies
    acprep: INFO: Executing phase: dependencies
    acprep: INFO: Installing Ledger's build dependencies ...
    acprep: INFO: Looks like you are using APT on Ubuntu jammy
    acprep: INFO: I do not recognize your version of Ubuntu!

This error was also present when installing on Ubuntu 20.04 Focal. The solution for Focal is the same for Jammy, which is to add `ubuntu-jammy` to `BoostInfo` and `jammy` as another `release`.

### DIFF
--- a/acprep
+++ b/acprep
@@ -63,7 +63,7 @@ class BoostInfo(object):
         if system in ['centos']:
             return [ 'boost-devel' ]
 
-        elif system in ['ubuntu-focal', 'ubuntu-bionic', 'ubuntu-xenial',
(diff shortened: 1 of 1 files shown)


burntraisin:
The solution for Focal that I'm referring to is [here](https://github.com/ledger/ledger/pull/2037).

afh (approved):
Thank you, @burntraisin 👍

PR #2289, opened 2023-09-06
print: Align amounts even when account names are long

gwymor:
When the account name is longer than the --account-width (default 36), the amounts stop aligning:

    2023/01/01 Transaction with long account names
        Assets:Very:Long:Account:Name:That:Will:Push:The:Amount     -10 ABC
        Assets:Another:Long:Account:Name:That:Will:Push:The:Amount     -10 ABC
        Expenses:Short                            20 ABC

One can set a larger --account-width, but that is not a great solution for cases where you have only a few accounts with problematically long names. Instead, keep the current account width wherever possible, but when an account name is longer than the account width, account for that and still align the values:

    2023/01/01 Transaction with short account names
        Assets:Short                             -10 ABC
        Assets:Short                             -10 ABC
        Expenses:Short                            20 ABC

    2023/01/01 Transaction with long account names
        Assets:Very:Long:Account:Name:That:Will:Push:The:Amount        -10 ABC
        Assets:Another:Long:Account:Name:That:Will:Push:The:Amount     -10 ABC
        Expenses:Short                                                  20 ABC

This is similar to hledger's behavior.

### DIFF
--- a/src/print.cc
+++ b/src/print.cc
@@ -100,6 +100,33 @@ namespace {
     }
   }
 
+  std::ostringstream format_account_name(xact_t& xact, post_t * post)
+  {
+    std::ostringstream pbuf;
+
+    if (xact.state() == item_t::UNCLEARED)
+      pbuf << (post->state() == item_t::CLEARED ? "* " :
+              (post->state() == item_t::PENDING ? "! " : ""));
+
+    if (post->has_flags(POST_VIRTUAL)) {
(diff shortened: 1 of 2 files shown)


afh:
What are your thoughts around this functionality in formats (`--format` option) rather than the default ledger behaviour?

jwiegley (approved):
I like this change, and in fact do this myself manually in my own Ledger files when it happens.

PR #2300, opened 2023-11-23
Add support for hash chaining to detect modifications in postings

jwiegley:
The following details of a posting contribute to its hash:

- fullname of account
- string representation of amount

Each posting hashes contributes to the transaction hash, which is compromised of:

- previous transaction’s hash (as encountered in parsing order)
- actual date
- optional auxiliary date
- optional code
- payee
- hashes of all postings

Note that this means that changes in the “code” or any of the comments

### DIFF
--- a/.github/workflows/nix-flake.yml
+++ b/.github/workflows/nix-flake.yml
@@ -13,7 +13,7 @@ jobs:
         os: [ubuntu-latest, macos-latest]
 
     steps:
-    - uses: cachix/install-nix-action@v24
+    - uses: cachix/install-nix-action@v25
       with:
         nix_path: nixpkgs=channel:nixos-unstable
 
--- a/acprep
+++ b/acprep
@@ -100,7 +100,6 @@ class CommandLineApp(object):
 
     force_exit    = True           # If true, always ends run() with sys.exit()
     log_handler   = None
-    boost_major   = "1_52"
 
     def __init__(self):
         "Initialize CommandLineApp."
@@ -358,25 +357,6 @@ class PrepareBuild(CommandLineApp):
             self.log.error("Execution failed: " + e)
             sys.exit(1)
 
-    def get_stdout(self, *args):
-        try:
-            self.log.debug('Executing command: ' + ' '.join(args))
-
-            proc    = Popen(args, shell=False, stdout=PIPE)
-            stdout  = proc.stdout.read()
-            retcode = proc.wait()
-            if retcode < 0:
-                self.log.error("Child was terminated by signal",
-                               -retcode)
-                sys.exit(1)
-            elif retcode != 0:
-                self.log.error("Execution failed: " + ' '.join(args))
-                sys.exit(1)
-            return stdout[:-1]
-        except OSError as e:
-            self.log.error("Execution failed:" + e)
-            sys.exit(1)
-
     def isnewer(self, file1, file2):
         "Check if file1 is newer than file2."
         if not exists(file2):
@@ -502,10 +482,10 @@ class PrepareBuild(CommandLineApp):
 
         self.log.info("Installing Ledger's build dependencies ...")
 
-        system = self.get_stdout('uname', '-s').decode(locale.getpreferredencoding())
+        system = os.uname().sysname
 
         if system == 'Darwin':
-            if exists('/opt/local/bin/port'):
+            if shutil.which('port'):
                 self.log.info('Looks like you are using MacPorts on macOS')
                 packages = [
                     'sudo', 'port', 'install', '-f',
@@ -522,7 +502,7 @@ class PrepareBuild(CommandLineApp):
                 ] + BoostInfo().dependencies('darwin-macports')
                 self.log.info('Executing: ' + ' '.join(packages))
                 self.execute(*packages)
-            elif exists('/usr/local/bin/brew') or exists('/opt/local/bin/brew'):
+            elif shutil.which('brew'):
                 self.log.info('Looks like you are using Homebrew on macOS')
                 packages = [
                     'brew', 'install',
@@ -531,15 +511,21 @@ class PrepareBuild(CommandLineApp):
                 ] + BoostInfo().dependencies('darwin-homebrew')
                 self.log.info('Executing: ' + ' '.join(packages))
                 self.execute(*packages)
-            elif exists('/sw/bin/fink'):
+            elif shutil.which('fink'):
                 self.log.info('Looks like you are using Fink on macOS')
                 self.log.error("I don't know the package names for Fink yet!")
                 sys.exit(1)
+            elif shutil.which('nix'):
+                self.log.info('Looks like you are using Nixpkgs on macOS')
+                packages = [
+                    'nix', 'develop',
+                ]
+                self.log.info('Executing: ' + ' '.join(packages))
+                self.execute(*packages)
 
         elif system == 'Linux':
             if exists('/etc/issue'):
-                issue = open('/etc/issue')
-                issue_name = issue.readline()
+                issue_name = open('/etc/issue').readline()
                 if issue_name.startswith('Ubuntu'):
                     info = dict([line.strip().split('=', 1)
                                  for line in open('/etc/lsb-release')])
@@ -709,6 +695,33 @@ class PrepareBuild(CommandLineApp):
                     self.log.info('Executing: ' + ' '.join(packages))
                     self.execute(*packages)
 
+            if exists('/etc/debian_version'):
+                release = open('/etc/debian_version').readline()
+                if release.startswith('trixie'):
+                    self.log.info(f'Looks like you are using APT on Debian {release}')
+                    packages = [
+                        'sudo', 'apt', 'install', '-y', '--no-install-recommends',
+                        'clang',
+                        'cmake',
+                        'ninja',
+                        'libboost-dev',
+                        'libmpfr-dev',
+                        'libgmp-dev',
+                        'libicu-dev',
+                        'libedit-dev',
+                        'libutfcpp-dev',
+                        'gettext',
+                        'texinfo',
+                        'doxygen',
+                        'graphviz',
+                        'lcov',
+                        'sloccount',
+                    ]
+                    self.log.info('Executing: ' + ' '.join(packages))
+                    self.execute(*packages)
+                else:
+                    self.log.info(f"I do not recognize your version '{release}' of Debian!")
+
             if exists('/etc/redhat-release'):
                 release = open('/etc/redhat-release').readline()
                 if release.startswith('CentOS'):
@@ -796,7 +809,7 @@ class PrepareBuild(CommandLineApp):
             self.configure_args.append(self.source_dir)
 
     def setup_for_system(self):
-        system = self.get_stdout('uname', '-s').decode(locale.getpreferredencoding())
+        system = os.uname().sysname
         self.log.info('System type is => ' + system)
 
         if self.options.enable_doxygen:


doc/ledger3.texi
```diff
@@ -6697,6 +6697,21 @@ $ ledger reg Expenses --group-by "payee" --group-title-format "-----------------
...
+reported. This provided value can also just be a prefix, at whatever
+length the user wishes, in which case only this amount of text is
+compared to the generated hash. Printing the entry with hashes enabled
+will display both this explicit metadata setting and the internally
+generated hash value.
+
+Hashes depend on previous entries, such that setting a single hash value
+is sufficient to guarantee the shape of the entire history leading up to
+that entry.
+
```
simonmichael, line 6714: Not the focus of this PR, but I see duplication is required between the man page and info manual which must create some friction for future edits. (And they are slightly different already, maybe it's intended that the man page has shorter text.)

test/baseline/opt-hashes-neg.test
```diff
@@ -0,0 +1,23 @@
+2008/01/01 January
+    Expenses:Books                            $10.00
+    Assets:Cash
+
+2008/01/31 End of January
+    Expenses:Books                            $10.00
+    Assets:Cash
+
+2008/02/01 February
+    ; Hash: f80797304b8d9acb2084772364ee717f0b724cddb1a9ccb6a3fb8d9cb803fe3b9b7a458211a7fe818578b6ec75a1a31acbcf8e2926ea349141568252966189aa
```
simonmichael, line 10: Hashes are rather long, eg filling one and a half lines of an 80-char terminal in print output. Could you get similar benefit with a shorter one ?

doc/ledger3.texi
```diff
@@ -6702,9 +6702,15 @@ Record the hash of each transaction in a @var{Hash} metadata value,
 according to the hashing algorithm given by the @var{ALGO}
 argument. Note that if a @var{Hash} metadata value is explicitly
 provided and does not match what would have been generated, an error is
-reported. Hashes depend on previous entries, such that setting a single
-hash value is sufficient to guarantee the shape of the entire history
-leading up to that entry.
+reported. This provided value can also just be a prefix, at whatever
+length the user wishes, in which case only this amount of text is
+compared to the generated hash. Printing the entry with hashes enabled
```
simonmichael, line 6707: On the surface this seems like a complexity that discourage users - if it's unavoidable, perhaps say a little more about the motivation and recommendations for use ? Would a different default behaviour work instead ?

simonmichael (commented):
Hi @johnw, reviewing as requested - no C++ code review, just some high level thoughts:

> The following details of a posting contribute to its hash:
> 
>   fullname of account
>   string representation of amount
> 
> Each posting hashes contributes to the transaction hash, which is compromised
> of:
> 
>   previous transaction’s hash (as encountered in parsing order)
>   actual date
>   optional auxiliary date
>   optional code
>   payee
>   hashes of all postings
> 
> Note that this means that changes in the “code” or any of the comments

Maybe the above details should appear in docs also ? Apologies if I missed it.

Posting's "string representation of amount" - that's the representation the journal file, I assume (not what print or reg would show).

> --hashes option requires an argument to specify the algorithm
> At the moment only "sha512" or "SHA512" is accepted, but this could extend to
> more algorithms in the future.

Cool feature!

As we discussed in chat, one obvious user benefit it promises is being able to warn when any past entries have changed in the journal files. VCS users can already detect this before commit, but this does not require a VCS and would be available to every Ledger user without setup. VCS users who don't check the diff before committing  might find it helpful to avoid accidentally committing fat-finger edits, eg.

I think users fairly often want to clean up small mistakes, whitespace, or even make bigger cleanups to old files and entries. And the commit messages above make me think this will be very sensitive - to any edits, to changes in hashing algorithm, to any rearrangement of included files or to different order of file arguments on the command line (because of "in parsing order"). So it's my guess users of this will quite often need to regenerate hashes for all of their data. Maybe that's not a problem, I'm not totally clear on the workflow. I imagine it would mean replacing at least some explicit Hash metadata values (tags) in journal entries in all old files (and committing those changes in VCS).

It seems to me to be a prototype that will need field testing and tweaking to find its best design and usage patterns. Possibly it's worth signalling this status to users by mentioning "Experimental" in descriptions.

As mentioned in chat Tackler has some similar-ish features described at https://tackler.e257.fi/docs/auditing - perhaps not this exactly, but there might be some interesting related ideas there.

Hope this helps! I appreciate this exploration and will follow with interest.

test/baseline/opt-hashes-neg.test
jwiegley: It will only require a comparison match up to the length given, so you can enter the first half of the hash, for example, and it will still give you a huge amount of assurance.

doc/ledger3.texi
jwiegley: The motivation is answered in the previous comment... perhaps it should be worded differently here?

jwiegley: Thanks, I've added this and more to the Texinfo file.

test/baseline/opt-hashes-neg.test
afh: Nixpkgs uses a shorter hashes, e.g. `hash = "sha512-IakZy6XsAE39awjzQI+R11JCPeQSaibx6+uX8Iea5WdlCundeovnPwSAi6RzzZl9dr2UftzzEiF4Aun8VMtqVA==";` (see [`musikcube/default.nix:41`](https://github.com/NixOS/nixpkgs/blob/24682ced2e23fff4144b790c3a0d18382b713a8c/pkgs/applications/audio/musikcube/default.nix#L41)), yet I do not know how these are shortened/encoded.

Nix offers [libutil](https://github.com/NixOS/nix/tree/master/src/libutil) although I'm uncertain how mature the library is on terms of use from third-party software and how sensible and feasible it is to integrate it into ledger.

doc/ledger.1
```diff
@@ -634,6 +634,24 @@ Set the format for the headers that separate reports section of
 a grouped report.  Only has effect with a
 .Fl \-group-by Ar EXPR
 register report.
+.It Fl \-hashes Ar ALGO
+Record the hash of each transaction in a
+.Ar Hash
+metadata value, according to the hashing algorithm given by the
+.Ar ALGO
+argument. Support algorithms are
```
afh, line 642: Fix spelling

```suggestion
argument. Supported algorithms are
```

doc/ledger.1
```diff
@@ -634,6 +634,24 @@ Set the format for the headers that separate reports section of
...
+.Ar sh512_256
+where both use SHA512, but the latter only stores the first half of
+the hash value.
+.Pp
+If a
+.Ar Hash
+metadata value is explicitly provided and does not match what would
+have been generated, an error is reported. Hashes depend on previous
+entries, such that setting a single hash value is sufficient to
+guarantee the shape of the entire history leading up to that entry.
```
afh, line 654: Nice!

doc/ledger.1
```diff
@@ -634,6 +634,24 @@ Set the format for the headers that separate reports section of
...
 register report.
+.It Fl \-hashes Ar ALGO
+Record the hash of each transaction in a
+.Ar Hash
+metadata value, according to the hashing algorithm given by the
+.Ar ALGO
+argument. Support algorithms are
+.Ar sha512
+or
+.Ar sh512_256
```
afh, line 645: The texinfo refers to this as `sha512_256` so I assume it is a typo here…

```suggestion
.Ar sha512_256
```

doc/ledger3.texi
```diff
@@ -6697,6 +6697,57 @@ $ ledger reg Expenses --group-by "payee" --group-title-format "-----------------
...
+they do not match, an error is reported. You may also write just a
+prefix of the @var{Hash}, which is less verbose but still gives quite
+good assurance.
+
+The support algorithms are:
+
+@table @code
+@item sha512
+Use the SHA512 hashing algorithm.
+@item sha512_256
```
afh, line 6715: Is `sha512_256` a common and known identifier for the first 256 bits of a SHA512 hash? If yes, what other software makes use of these? I'm curious.

doc/ledger3.texi
```diff
@@ -6697,6 +6697,57 @@ $ ledger reg Expenses --group-by "payee" --group-title-format "-----------------
...
+@itemize
+@item fullname of the account
+@item amount value
+@end itemize
+
+In addition, these details are hashed from the transaction itself:
+
+@itemize
+@item actual date
+@item auxiliary date (if provided)
```
afh, line 6741: effective date and auxiliary date are used somewhat interchangeably in the documented, maybe mentioning it here too could be helpful?

afh (commented):
Does this bring ledger closer to being a triple-entry accounting system? 😃 

I left some comments from first glance below, and will take a closer look when trying out the proposed changes on my machine.

src/sha512.cc
```diff
@@ -0,0 +1,479 @@
...
+typedef struct _SHA512_CTX {
+	uint64_t	state[8];
+	uint64_t	bitcount[2];
+	uint8_t	buffer[SHA512_BLOCK_LENGTH];
+} SHA512_CTX;
+#endif /* do we have sha512 header defs */
+
+void SHA512_Init(SHA512_CTX*);
+void SHA512_Update(SHA512_CTX*, void*, size_t);
+void SHA512_Final(uint8_t[SHA512_DIGEST_LENGTH], SHA512_CTX*);
```
afh, line 62: ℹ️ [Function implementation](https://github.com/ledger/ledger/pull/2300/files#diff-7113cc560e7e08fdcea571c3a876323d65b8365be8661b5ca07ab0364548b632R440) uses `sha2_byte[]` as first parameter type. Could this also be the cause for the [Nix Flake build failure](https://github.com/ledger/ledger/actions/runs/7052072331/job/19196340805?pr=2300):

```
       > /build/source/src/sha512.cc: In function 'unsigned char* SHA512(void*, unsigned int, unsigned char*)':
       > /build/source/src/sha512.cc:475:5: error: 'SHA512_Final' was not declared in this scope; did you mean 'SHA512_Init'?
       >   475 |     SHA512_Final(digest, &ctx);
       >       |     ^~~~~~~~~~~~
       >       |     SHA512_Init
```

```suggestion
void SHA512_Final(sha2_byte[SHA512_DIGEST_LENGTH], SHA512_CTX*);
```

doc/ledger3.texi
jwiegley: We use it for the blockchain at my place of work, and I've seen it as an algorithm in other crypto libraries. If you Google around, you'll see that it's a pretty well-known thing.

afh:
Thanks for the context on sha512_256, @jwiegley.

I did a bit of research into libraries supporting SHA-512/256, e.g. [libtomcrypt](https://github.com/libtom/libtomcrypt/blob/develop/src/hashes/sha2/sha512_256.c), [Botan-3](https://github.com/randombit/botan/blob/master/src/lib/hash/sha2_64/sha2_64.cpp), [pycryptodome](https://www.pycryptodome.org/src/hash/sha512#module-Crypto.Hash.SHA512). These libraries support it in a different manner than this PR suggests in the sense, that a truncated hash _"is not equivalent to simply truncating the output digest"_ ([pycryptodome](https://www.pycryptodome.org/src/hash/sha512#module-Crypto.Hash.SHA512)).

Possibly this is _"for users to be able to distinguish between a SHA-512 digest which has been truncated and a SHA512/256 digest, [offering] new initialization constants, analogous to those used in SHA-384."_ — https://eprint.iacr.org/2010/548.pdf
Is this to also thwart length extension attacks? (see https://news.ycombinator.com/item?id=21981874)

I took the liberty to hack on a little Nix Flake to get a feel for the API of the different implementations and evaluate the feasibility of base64 encoding. The Flake uses LibreSSL, Botan-3, and tomcrypt to compute the SHA-512 and SHA-512/256 hashes for the given arguments and print the hashes as a hex and a base64 encoded string.

What are your thoughts on using one of the aforementioned libraries or another third-party implementation?

```sh
nix build https://projects.surryhill.net/ledger/sha512-test-0.1.3.tgz
./result/bin/sha512-test 'Heureka!'
input           = Heureka!
LibreSSL
sha512          = 83bcf89b75e21ab7d9fe332a6f82ca4d1e94ec587cec1e137d50087fcc6b7518f366ee9e2ba086346bdcc0561a522db4b3bdebc53483199f58ac7139531ded7c
sha512/evp      = g7z4m3XiGrfZ/jMqb4LKTR6U7Fh87B4TfVAIf8xrdRjzZu6eK6CGNGvcwFYaUi20s73rxTSDGZ9YrHE5Ux3tfA==
sha512_256      = 83bcf89b75e21ab7d9fe332a6f82ca4d1e94ec587cec1e137d50087fcc6b7518
sha512_256/evp  = N/A
Botan3
SHA-512         = 83BCF89B75E21AB7D9FE332A6F82CA4D1E94EC587CEC1E137D50087FCC6B7518F366EE9E2BA086346BDCC0561A522DB4B3BDEBC53483199F58AC7139531DED7C
SHA-512.b64     = g7z4m3XiGrfZ/jMqb4LKTR6U7Fh87B4TfVAIf8xrdRjzZu6eK6CGNGvcwFYaUi20s73rxTSDGZ9YrHE5Ux3tfA==
SHA-512/256     = 5EABC68E077C6338A305D388F20AE3A04200F6D942164FFBFD659E345C39D0A7
SHA-512/256.b64 = XqvGjgd8YzijBdOI8grjoEIA9tlCFk/7/WWeNFw50Kc=
tomcrypt
SHA-512         = 83bcf89b75e21ab7d9fe332a6f82ca4d1e94ec587cec1e137d50087fcc6b7518
SHA-512.b64     = g7z4m3XiGrfZ/jMqb4LKTR6U7Fh87B4TfVAIf8xrdRjzZu6eK6CGNGvcwFYaUi20s73rxTSDGZ9YrHE5Ux3tfA==
SHA-512/256     = 5eabc68e077c6338a305d388f20ae3a04200f6d942164ffbfd659e345c39d0a7
SHA-512/256.b64 = XqvGjgd8YzijBdOI8grjoEIA9tlCFk/7/WWeNFw50Kc=
```

If you'd like to inspect the small utility closer have a look at `main.cc` below or download and unpack the flake archive from https://projects.surryhill.net/ledger/sha512-test-0.1.3.tgz and open main.cc in your `$EDITOR`.

<details><summary><code>main.cc</code></summary>

```cpp
#include <string>
#include <sstream>
#include <iomanip>
#include <iostream>
#include <vector>

#include <openssl/crypto.h>
#include <openssl/sha.h>

#include <openssl/hmac.h>
#include <openssl/evp.h>
#include <openssl/bio.h>
#include <openssl/buffer.h>

#include <botan-3/botan/hash.h>
#include <botan-3/botan/hex.h>
#include <botan-3/botan/base64.h>

#include <tomcrypt.h>

// Convert buffer to hex string. Kudos to https://github.com/ledger/ledger/pull/2300
std::string bufferToHex(const unsigned char* buffer, std::size_t size) {
    std::ostringstream oss;
    oss << std::hex << std::setfill('0');
    for(std::size_t i = 0; i < size; ++i)
        oss << std::setw(2) << static_cast<int>(buffer[i]);
    return oss.str();
}

// Encode input as base64 using LibreSSL BIO. Kudos to https://ioncannon.net/programming/34/howto-base64-encode-with-cc-and-openssl/
char *bio_base64(const unsigned char *input, int length) {
  BIO *bmem, *b64;
  BUF_MEM *bptr;
 
  b64 = BIO_new(BIO_f_base64());
  bmem = BIO_new(BIO_s_mem());
  b64 = BIO_push(b64, bmem);
  BIO_write(b64, input, length);
  BIO_flush(b64);
  BIO_get_mem_ptr(b64, &bptr);
 
  char *buff = (char *)malloc(bptr->length);
  memcpy(buff, bptr->data, bptr->length-1);
  buff[bptr->length-1] = 0;
 
  BIO_free_all(b64);
 
  return buff;
}

// Encode input as base64 using LibreSSL EVP. Kudos to mtrw https://stackoverflow.com/a/60580965
char *evp_base64(const unsigned char *input, int length) {
  const auto pl = 4*((length+2)/3);
  auto output = reinterpret_cast<char *>(calloc(pl+1, 1)); // +1 for the terminating null that EVP_EncodeBlock adds on
  const auto ol = EVP_EncodeBlock(reinterpret_cast<unsigned char *>(output), input, length);
  if (pl != ol) { std::cerr << "Whoops, encode predicted " << pl << " but we got " << ol << "\n"; }
  return output;
}

// Encode input as base64 using tomcrypt. Kudos to https://techoverflow.net/2012/11/20/cc-base64-codec-using-libtomcrypt/
std::string encodeBase64(const char* input, const unsigned long inputSize) {
    unsigned long outlen = inputSize + (inputSize / 3.0) + 16;
    unsigned char* outbuf = new unsigned char[outlen]; //Reserve output memory
    base64_encode((unsigned char*) input, inputSize, outbuf, &outlen);
    std::string ret((char*) outbuf, outlen);
    delete[] outbuf;
    return ret;
}

int main(int argc, char*argv[]) {

  // Setup program arguments
  std::vector<std::string> args;
  if (argc > 1)
    args.assign(argv+1, argv + argc);

  // Initialize LibreSSL
  OPENSSL_init_crypto(0, NULL);

  // Initialize Botan3
  const auto bt_sha512_256 = Botan::HashFunction::create_or_throw("SHA-512-256");
  const auto bt_sha512     = Botan::HashFunction::create_or_throw("SHA-512");

  // Initialize tomcrypt
  register_all_ciphers();
  register_all_hashes();
  int tc_sha512_256 = find_hash("sha512-256");
  int tc_sha512     = find_hash("sha512");

  for (const auto& arg : args) {
    std::cout << "input           = " << arg << std::endl;

    // Compute Hash using LibreSSL
    const unsigned char* input = (const unsigned char*)arg.c_str();
    unsigned char* hv = SHA512(input, arg.length(), NULL);
    std::cout << "LibreSSL" << std::endl
      << "sha512          = " << bufferToHex(hv, 64)
      << std::endl
      << "sha512/evp      = " << evp_base64(hv, SHA512_DIGEST_LENGTH)
      << std::endl
      << "sha512_256      = " << bufferToHex(hv, 32)
      << std::endl
      << "sha512_256/evp  = " << "N/A"
      // bio_base64 includes a new line every 64 bytes, which is impractical
      // for ledger's use-case, i.e. single line checksum.
      //<< std::endl
      //<< "sha512/bio      = " << bio_base64(hv, SHA512_DIGEST_LENGTH)
      << std::endl;

    // Compute Hash using Botan3
    bt_sha512->update(input, arg.length());
    auto bt_sha512f = bt_sha512->final();
    std::cout << "Botan3" << std::endl
      << "SHA-512         = " << Botan::hex_encode(bt_sha512f)
      << std::endl
      << "SHA-512.b64     = " << Botan::base64_encode(bt_sha512f)
      << std::endl;

    bt_sha512_256->update(input, arg.length());
    auto bt_sha512_256f = bt_sha512_256->final();
    std::cout
      << "SHA-512/256     = " << Botan::hex_encode(bt_sha512_256f)
      << std::endl
      << "SHA-512/256.b64 = " << Botan::base64_encode(bt_sha512_256f)
      << std::endl;

    // Compute Hash using tomcrypt
    unsigned long outl = MAXBLOCKSIZE;
    unsigned char* out = (unsigned char*)XMALLOC(MAXBLOCKSIZE);
    if (hash_memory(tc_sha512, (unsigned char*)input, arg.length(), out, &outl) != CRYPT_OK)
      continue;
    std::cout << "tomcrypt" << std::endl
      << "SHA-512         = " << bufferToHex(out, 32)
      << std::endl
      << "SHA-512.b64     = " << encodeBase64((const char*)out, outl)
      << std::endl;

    if (hash_memory(tc_sha512_256, (unsigned char*)input, arg.length(), out, &outl) != CRYPT_OK)
      continue;
    std::cout
      << "SHA-512/256     = " << bufferToHex(out, 32)
      << std::endl
      << "SHA-512/256.b64 = " << encodeBase64((const char*)out, outl)
      << std::endl;
  }

  OPENSSL_cleanup();
  return 0;
}
```

</details>

src/sha512.cc
```diff
@@ -0,0 +1,479 @@
...
+typedef struct _SHA512_CTX {
+	uint64_t	state[8];
+	uint64_t	bitcount[2];
+	uint8_t	buffer[SHA512_BLOCK_LENGTH];
+} SHA512_CTX;
+#endif /* do we have sha512 header defs */
+
+void SHA512_Init(SHA512_CTX*);
+void SHA512_Update(SHA512_CTX*, void*, size_t);
+void SHA512_Final(sha2_byte[SHA512_DIGEST_LENGTH], SHA512_CTX*);
```
afh, line 62: Apologies my previous suggestion broke the build as `sha2_byte` is `typedef`'d on line 99 below and not available yet here.

How come this file is named `sha512.cc` when its comments refer to itself as `sha2.c`?
From a maintainer's perspective I wonder whether adding another third-party dependency to handle calculating the hash sums is more convenient (see [my other comment](https://github.com/ledger/ledger/pull/2300#issuecomment-1845374048)).
What do you think?

```suggestion
void SHA512_Final(uint8_t[SHA512_DIGEST_LENGTH], SHA512_CTX*);
```

doc/ledger3.texi
jwiegley: Done!

afh (commented):
@jwiegley I see several changes to the newly added `src/sha512.cc` file. I'd prefer to treat it as third-part code that is integrated into ledger verbatim (just like utfcpp) or even better replace it with a third-party library (see my [previous comment](https://github.com/ledger/ledger/pull/2300#issuecomment-1845374048)) that provides calculation of SHA-512 and SHA-512/256 ideally with support for base64 encoding.

jwiegley:
> @jwiegley I see several changes to the newly added `src/sha512.cc` file. I'd prefer to treat it as third-part code that is integrated into ledger verbatim (just like utfcpp) or even better replace it with a third-party library (see my [previous comment](https://github.com/ledger/ledger/pull/2300#issuecomment-1845374048)) that provides calculation of SHA-512 and SHA-512/256 ideally with support for base64 encoding.
> 
> What are your thoughts?

I can reduce the number of changes down to just `s/uint8_t/unsigned char/`, which I can even do before including it elsewhere, so let me try removing all changes to it.

jwiegley:
@afh I've reverted all of my changes to the SHA512 code. What I would like to understand now is why the flake build here on GitHub fails, when it succeeds just fine on my machine, using either `nix build` or `nix develop` followed by make.

jwiegley:
I was able to reproduce the build failure in a Linux VM, and found that all we're missing are two standard system headers.

jwiegley:
> What are your thoughts on using one of the aforementioned libraries or another third-party implementation?

I'm not really excited about new dependencies, they come with so many other costs (maintenance, licensing, keeping up-to-date, etc). This is a stable, simple algorithm, and we can crack the can on using a 3rd party library if it becomes a popular feature and people end up wanting other algorithms besides the default ones offered.

jwiegley:
I'm current awaiting a close review of `xact_t::hash` from @afh before merging this in.

afh:
I have this on my agenda, yet it'll likely take me until after the holidays and probably New Year's before I get to this. 🎄🎇

jwiegley:
@afh Any luck with that review?

PR #2307, opened 2023-12-03
Improve finding ICU libraries

afh:


afh:
Thanks for reviewing and approving the changes, John, very helpful :)

PR #2325, opened 2024-01-17
Improve and clean-up cmake

afh:


CMakeLists.txt
```diff
@@ -164,7 +162,7 @@ endif()
 cmake_push_check_state()
 
 set(CMAKE_REQUIRED_INCLUDES ${CMAKE_INCLUDE_PATH} ${Boost_INCLUDE_DIRS} ${ICUE_INCLUDE_DIRS})
-set(CMAKE_REQUIRED_LIBRARIES ${Boost_LIBRARIES} ${ICU_LIBRARIES} ${PROFILE_LIBS})
+set(CMAKE_REQUIRED_LIBRARIES Boost::regex ${ICU_LIBRARIES} ${PROFILE_LIBS})
```
afh, line 165: I'm not exactly sure why the `check_cxx_source_runs` check following this line fails when `${Boost_LIBRARIES}` includes boost_python, yet it does. Given that the `check_cxx_source_runs` check only includes the `<boost/regex/icu.hpp>` header, it seems reasonable to only require the boost regex library.

test/CMakeLists.txt
```diff
@@ -19,7 +24,7 @@ macro(add_ledger_harness_tests _class)
     file(GLOB ${_class}_TESTS *.test)
     foreach(TestFile ${${_class}_TESTS})
       get_filename_component(TestFile_Name ${TestFile} NAME_WE)
-      string(FIND ${TestFile_Name} "_py.test" TestFile_IsPythonTest)
+      string(FIND ${TestFile} "_py.test" TestFile_IsPythonTest)
```
afh, line 27: The `TestFile_Name` variable contains the _"File name with neither the directory nor the longest extension" (see [documentation for `get_filename_component`](https://cmake.org/cmake/help/latest/command/get_filename_component.html)), so either the string to search for should be `"_py"` or the variable to search should be `${TestFile}`.

jwiegley (approved):
Thank you, Alexis.

PR #2326, opened 2024-01-18
Expand the size of an arbitrary safety limit

jwiegley:
This limit is just a sanity check, but there is no reason it should be one size or another. I ran into a real world use case where the current limit was much too small.

### DIFF
--- a/src/error.cc
+++ b/src/error.cc
@@ -92,7 +92,7 @@ string source_context(const path&                  file,
     return _("<no source context>");
 
   assert(len > 0);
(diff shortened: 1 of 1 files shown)


afh (approved):
I'm fine with increasing the sanity check limit, yet I'm wondering
what a real-world example of where this assert would be helpful to developers or users is?

jwiegley:
All you need is a really big transaction, and when Ledger attempts to report an error by displaying this transaction back to you, it will hit this limit.

PR #2327, opened 2024-01-18
Change an assertion into an if test

jwiegley:
The given assertion is too strict. If the source and target are the same for a market value valuation, simply return `none`. Note that this happened to me on my data when I was playing around with the `market` function.


afh (approved):
LGTM 👍

PR #2333, opened 2024-02-25
Improve support for Python version specific code

### LINKED ISSUE #2332: Ledger 3.3.2 cannot be built with Python 3.12 support
Python deprecated a long time ago the PyUnicode_AS_UNICODE macro and PyUNICODE type, per https://peps.python.org/pep-0624/ .  With Python 3.12, these have been finally removed.

As a result, building on Fedora 39 with Python support is not possible anymore, because Fedora comes with Python 3.12.

I'm not sure what needs to be fixed, but it looks like a simple fix of changing a few macros.  Thanks in advance for looking into it.

afh:
for details see individual commits.

The use of raw strings `r''` addresses the following warning `SyntaxWarning: invalid escape sequence \s warning`

Fixes #2332

### DIFF
--- a/acprep
+++ b/acprep
@@ -413,16 +413,16 @@ class PrepareBuild(CommandLineApp):
 
             version_m4 = open('CMakeLists.txt', 'r')
             for line in version_m4.readlines():
-                match = re.match('^set\(Ledger_VERSION_MAJOR ([0-9]+)\)', line)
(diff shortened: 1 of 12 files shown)


jwiegley (approved):
All looks good to me.

Rudd-O:
And thank you so much @afh!

Issue #2031, opened 2021-05-02
Change 0018c884dbf228c54e30f0bc8b7586cc35c56b0f breaks conversion from time to money


Gumnos:
I used `git-bisect` to track down that [this commit causes breakage](https://github.com/ledger/ledger/commit/0018c884dbf228c54e30f0bc8b7586cc35c56b0f#commitcomment-50290475). Prior to this commit, I could convert time-log entries to dollars using my price-db. After this commit, it gives me errors. Sample/test files and the command-invocation are shown in that commit comment.

Gumnos:
This was found [in this conversation over on Reddit](https://old.reddit.com/r/plaintextaccounting/comments/mzd6pd/converting_timelog_hours_to_usd_not_respecting/?context=10) and tracked down that it worked on Ledger version 3.1.3-20190331 which I had on OpenBSD 6.8, and broke when I upgraded packages from 6.8 to 6.9, bringing me to Ledger version 3.2.1-20200518. The latest version gives me different/updated error

    While parsing file "/home/gumnos/test.ledger", line 6:
    Error: Only one posting with null amount allowed per transaction

Which is misleading in a different direction.

tbm:
Let's copy @daraul whose change broke this.  @daraul can you look into this please.

tbm:
@daraul there's a test case: https://github.com/ledger/ledger/commit/0018c884dbf228c54e30f0bc8b7586cc35c56b0f#commitcomment-50290475

daraul:
Oh great, thanks. I'm swamped atm so I can't promise it'll be fixed anytime soon, but it's on the TODO list.

daraul:
I'm confused, but it looks as though this was fixed without me. @Gumnos, could you try reproducing the error on the master branch? I tried bisecting to find the “bad” commit that fixed it, but ended up at https://github.com/ledger/ledger/commit/051e4ab1ef82c58dc9563281ec14c02fc250e20f (which was made before you found this problem). I was able to confirm that 0018c884dbf228c54e30f0bc8b7586cc35c56b0f results in the error you mentioned, though.

daraul:
I've written the test case out [here](https://github.com/daraul/ledger/commit/3631cab0cdcc19910964f8e31d7ab5ac5023924e), but I'd appreciate @Gumnos confirming whether or not `$37` is what he's expecting to see in `Assets:Bank:Checking`, based on his test command.

daraul:
[...]

PR #2336, opened 2024-04-15
Fix for #2031

daraul:
This patch includes a solution for #2031, but also causes two tests to fail, that don't fail on the master branch:

- 55 - BaselineTest_feat-balance_assert_split (Failed)
- 378 - RegressTest_ACE05ECE (Failed)

What I found happening is that time log entries were treated as virtual postings, but I am not sure if that is the intended way to treat them, or just a workaround to ensure they can get past the usual requirement that a posting should balance -- my solution assumes the latter. 

My solution there adds a new flag to state whether a posting is for a time log entry, and thereby returns false from the post class' `must_balance` method, when that flag is set on the posting. 

The two tests that fail include a virtual(?) posting that I have never seen before (and cannot find mentioned in the docs), and a time log entry that is asserted to have been posted to a virtual account.


tbm:
Unfortunately  don't know anything about this.  @jwiegley can you comment on this?

jwiegley:
This looks like a good change to me, I'll look into why those tests are failing.

daraul:
Closing this; it's now superceded by John's change: https://github.com/ledger/ledger/pull/2435

PR #2337, opened 2024-04-25
Fix build with Boost 1.85.0

cho-m:
Currently, the build with Boost 1.85.0 fails due to usage of removed `convenience.hpp` https://github.com/boostorg/filesystem/commit/5df060e95ca844fe91b29001b4ae22bdb65635c6

I didn't see any usage of removed functionality from header. Main question would be if any indirect headers were used. At least I didn't see any missing headers when compiling with this change.

Ref: https://www.boost.org/doc/libs/1_85_0/libs/filesystem/doc/deprecated.html

---

Seen while updating Boost in https://github.com/Homebrew/homebrew-core/pull/169237

### DIFF
--- a/src/system.hh.in
+++ b/src/system.hh.in
@@ -140,7 +140,6 @@
 #include <boost/date_time/posix_time/posix_time_io.hpp>
 #include <boost/date_time/gregorian/gregorian_io.hpp>
 
-#include <boost/filesystem/convenience.hpp>
 #include <boost/filesystem/exception.hpp>
 #include <boost/filesystem/fstream.hpp>
 #include <boost/filesystem/operations.hpp>


afh:
Thanks for raising this, @cho-m. Would be it be possible to include the header only if a boost version compile flag indicates that boost >= 1.85.0 is being used?

cho-m:
> Thanks for raising this, @cho-m. Would be it be possible to include the header only if a boost version compile flag indicates that boost >= 1.85.0 is being used?

The change is removing the header so would be inverse (boost < 1.85.0)

I see the minimum supported Boost is 1.72 (https://github.com/ledger/ledger/blob/master/CMakeLists.txt#L2) so question would be is there anything that older Boost versions need https://www.boost.org/doc/libs/1_72_0/boost/filesystem/convenience.hpp

I don't think `extension()`, `basename()`, or `change_extension()` is used (otherwise build will fail as these are removed from Boost), so only reason would be headers, e.g. `boost/config.hpp`, `boost/system/error_code.hpp` (which should be directly included if so).

---

Anyway, I am not too sure on compile flags. A common way is to do a macro check on `BOOST_VERSION`, e.g.
```hpp
#include <boost/version.hpp>
#if BOOST_VERSION < 108500
#include <boost/filesystem/convenience.hpp>
#endif
```

https://www.boost.org/doc/libs/1_72_0/boost/version.hpp
https://www.boost.org/doc/libs/1_85_0/boost/version.hpp

afh:
Ledger already defines `BOOST_FILESYSTEM_NO_DEPRECATED` ([`CMakeLists.txt:23`](https://github.com/ledger/ledger/blob/master/CMakeLists.txt#L23))when compiling, so I'm a bit surprised this wasn't noticed earlier.

Thanks for digging deeper into the headers, very helpful. I think understanding what is necessary and why is the right path forward. I'll run some builds to gain some insight.

Correct, the version check would need to be inverted, my bad…

afh:
ℹ️ `boost/system/error_code.hpp` is also included by `boost/filesystem/operations.hpp`, which is directly inluded by `system.hh.in`, so that only leaves to check if `boost/config.hpp` is somehow indirectly being included by any of the boost related includes

afh (approved):
Given that #2345 fails to build to the issue that is addressed by this PR I'm going to approve and merge this.

Thanks for your contribution and investigation @cho-m, much appreciated!

Issue #2340, opened 2024-04-27
<cctype> abuse


riastradh:
Various parts of ledger pass values of type char, derived from arbitrary user input, into the standard C++ <cctype> functions std::isspace, std::isalpha, &c. (which are just copies of the standard C <ctype.h> functions).  For example:

https://github.com/ledger/ledger/blob/c679e3cbd5a44fa876c42b6912d9aa7729427c8f/src/textual.cc#L343

Here `line[...]` has type char.

The ctype functions take inputs of type int.  But when the input to the <cctype> functions is not representable by type unsigned char -- that is, when the value represented by the object of type char is negative -- and the value is not the value of the integer constant macro EOF, this is undefined behaviour:

> The header <ctype.h> declares several functions useful for classifying and mapping characters.  In all cases the argument is an int, the value of which shall be representable as an unsigned char or shall equal the value of the macro EOF.  If the argument has any other value, the behavior is undefined.
> 
> [_ISO C11, Sec. 7.4 Character handling <ctype.h>_](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf#page=218)

For example, given `char x[2] = {0xa2,0}`, which represents a NUL-terminated string of POUND SIGN (£) in the locale en_GB.ISO8859-1, on platforms where char is signed such as [the usual amd64 ABI](http://www.x86-64.org/documentation/abi.pdf) and EOF is -1, `isspace(x[0])` is undefined behaviour because the value of `x[0]` is -94, which is neither the value of EOF nor representable by the type unsigned char.  (It can be _converted_ to unsigned char, giving the different number 162.)

The undefined behaviour may manifest as silent wrong answers, as in https://github.com/ledger/ledger/issues/2338, or as crashes if the virtual page preceding a ctype table is mapped unreadable.  Even in cases where it is not undefined behaviour, the answer may be wrong.  For example, given `char x[2] = {0xff, 0}`, which might represent a NUL-terminated string of LATIN SMALL LETTER Y WITH DIAERESIS (ÿ) in the locale fr_FR.ISO8859-1, `isalpha(x[0])` will return the same result as `isspace(EOF)` on most systems (where EOF is -1), giving false when the correct answer is true.

The reason for this peculiar interface is that the ctype functions were intended to handle the results of C functions like fgetc, or C++ functions like std::istream::peek, which return values of type int, which can be _either_ EOF (typically -1) _or_ a value that is representable in unsigned char (0, 1, 2, ..., 255):

```
[...]

Issue #2338, opened 2024-04-27
Encoding error


riastradh:
The attached file [in.txt](https://github.com/ledger/ledger/files/15139116/in.txt) does not round-trip under `ledger print`, which yields a file [out.txt](https://github.com/ledger/ledger/files/15139117/out.txt) with invalid UTF-8 encoding.

Somehow, the input octet sequence `e7 94 b0` (i.e., Unicode code point U+7530, CJK UNIFIED IDEOGRAPH-7530) is mapped to the output octet sequence `e7 94`, which is invalid UTF-8 because it is missing the `b0` octet.

Reproduced in ledger 3.2.1-20200518 (which I'm still actively using in production because of https://github.com/ledger/ledger/issues/2318) and in ledger 3.3.2-20230330, under NetBSD 9.3.

I have also seen a variant of this where the output is _sometimes_ encoded correctly and _sometimes_ encoded incorrectly, varying randomly from invocation to invocation of the same ledger executable on the same input file.

PR #2341, opened 2024-04-27
Avoid ctype abuse.

riastradh:
fix https://github.com/ledger/ledger/issues/2338
fix https://github.com/ledger/ledger/issues/2340

### DIFF
--- a/src/amount.cc
+++ b/src/amount.cc
@@ -1013,7 +1013,8 @@ namespace {
               std::isdigit(c) || c == '.' || c == ',');
 
     string::size_type len = std::strlen(buf);
-    while (len > 0 && ! std::isdigit(buf[len - 1])) {
+    while (len > 0 &&
+           ! std::isdigit(static_cast<unsigned char>(buf[len - 1]))) {
       buf[--len] = '\0';
       in.unget();
     }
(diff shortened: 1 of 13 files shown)


src/amount.cc
```diff
@@ -1034,7 +1035,7 @@ bool amount_t::parse(std::istream& in, const parse_flags_t& flags)
     commodity_t::parse_symbol(in, symbol);
 
     if (! in.eof() && ((n = static_cast<char>(in.peek())) != '\n')) {
-      if (std::isspace(static_cast<char>(in.peek())))
+      if (std::isspace(in.peek()))
```
afh, line 1038: Can you provide more context please, as to why the `static_cast` should be removed here, please?

src/option.cc
```diff
@@ -132,7 +132,8 @@ void process_environment(const char ** envp, const string& tag,
         if (*q == '_')
           *r++ = '-';
         else
-          *r++ = static_cast<char>(std::tolower(*q));
+          *r++ = static_cast<char>(std::tolower(
+            static_cast<unsigned char>(*q)));
```
afh, line 136: Is first casting to `unsigned char` and then casting to `char` intentional here? If yes, what are the implications?

afh (commented):
Thank you for this PR @riastradh, very much appreciated!
Would you be willing to also add tests for the changes? I think that tests would help explain the issue(s) with the code prior to this PR and avoid regressions in the future.

src/option.cc
afh: Thanks for taking the time to reply in detail, @riastradh. I agree with your reasoning.

src/amount.cc
afh: That's helpful, thanks! Do you know any platforms where EOF is -1? Once I again, I'd like to kindly ask for tests that catch ledger's current misbehaviour and ensures regressions to this fix are caught. Would you be willing to do that? If there are questions on how to write tests for ledger, I'm happy to help.

jwiegley:
I'd love to merge it, but there is a CI build failure yet to be addressed.

jwiegley (approved):

Issue #2350, opened 2024-06-18
Issue with generated accounts in Automated Transactions


igbanam:
## tl;dr

When an automated transaction uses a complex query to specify the rule, if a new account gets introduced in that rule, this account is omitted from the balance report.

## long form

Consider this simple transaction

```ledger
2024-06-18 * Payee
    Expenses    42 EUR
    Assets:Checking
```

A simple balance query shows

```
  -42 EUR  Assets:Checking
   42 EUR  Expenses
---------
        0
```

Now, let's assume this purchase is from a store which has the VAT baked in; but we want to know how much VAT we pay per purchase. A simple rule for an [automated transaction](https://ledger-cli.org/doc/ledger3.html#Automated-Transactions) would look like

```ledger
= ^Expenses
    Expenses:Tax:VAT  0.13
    $account  -0.13
```

Our balance query now yields

```
  -42 EUR  Assets:Checking
   42 EUR  Expenses
    5 EUR    Tax:VAT
---------
        0
```

This makes sense… until we want to target VAT to certain postings. The `^Expenses` rule no longer works. We would need some [complex query](https://ledger-cli.org/doc/ledger3.html#Complex-expressions) to properly target these postings. One way I thought of doing this is using tags. Our ledger now becomes…

```ledger
= %/vat/
    Expenses:Tax:VAT  0.13
    $account  -0.13

2024-06-18 * Payee ; :vat:
    Expenses    42 EUR
    Assets:Checking
```

>Note the tag is on the payee line

With this, the balance query looks like

```
  -42 EUR  Assets:Checking
   42 EUR  Expenses
---------
        0
```

…no sign of the `Expenses:Tax:VAT` account.

When I shift the tag into the posting, like so

```ledger
= %/vat/
    Expenses:Tax:VAT  0.13
    $account  -0.13

2024-06-18 * Payee
    ; :vat:
    Expenses    42 EUR
    Assets:Checking
```

Note the posting is now applies to the Expenses directly. The balance query now reports

```
  -37 EUR  Assets:Checking
   37 EUR  Expenses
---------
        0
```

This is what I expect — 5 EUR out of this should be in `Expenses:Tax:VAT` — but there's no `Expenses:Tax:VAT` listed in the balance report. 

When I try to switch out `= %/vat/` for `= expr has_tag("vat")`, `= expr has_tag(/vat/)`, or any other expression which isn't the short form, I get

```
While parsing file "/path/to/auto-tags.ledger", line 1:
[...]

Issue #872, opened 2013-01-07
parse error in '= expr has_tag("TAG")' (BZ#872)


tbm:
Note: the issue was created automatically with bugzilla2github

Bugzilla bug ID: BZ#872
From: Ethan Glasser-Camp (@glasserc)
CC:   unknown user, @thdox

tbm:
Issue #1185 contains a similar parsing issue with "expr any()".

PR #2351, opened 2024-06-18
Make the expr syntax explicit in the docs

igbanam:
ledger/ledger#2350 shows me it's possible for the docs to be misinterpreted. This could lead to errors as explained in the issue. This change makes the syntax more explicit so people can translate for their use cases easier


tbm:
afaik the quotes are only needed because of a parsing issue; so the real fix would be to solve the parsing issue (#872.)

Issue #2353, opened 2024-06-28
Stock sale transactions using the implicit commodity do not balance, they are balanced only when using an explicit commodity


hoylen:
Stock sale transactions are balanced when explicit commodities are used to represent prices, but do not balance when the implicit commodity is used.

This is the example from [section 4.5.2 Buying and Selling Stock](https://ledger-cli.org/doc/ledger3.html#Buying-and-Selling-Stock) of the manual.

Values are represented using explicit commodities. The explicit commodity is `$`, with  the _lot price_ being `$30.00` and the _sale price_ of `$50.00`.

```
2005-08-01 Stock sale
    Assets:Broker                                -50 AAPL {$30.00} @ $50.00
    Expenses:Broker:Commissions               $19.95
    Income:Capital Gains                  $-1,000.00
    Assets:Broker                          $2,480.05
```

But the same transaction is incorrectly treated as not balanced, when values are represented using the implicit commodity.  When the _lot price_ is `30.00` and _sale price_ is `50.00`.

## Expected behaviour

The following transaction should be balanced.

```
2005-08-01 Stock sale
    Assets:Broker                                -50 AAPL {30.00} @ 50.00
    Expenses:Broker:Commissions                19.95
    Income:Capital Gains                   -1,000.00
    Assets:Broker                           2,480.05
```

## Actual behaviour

Ledger 3.3.2-20230330 says the transaction is not	balanced.

```
While parsing file "commodity-bug.ledger", line 10:
While balancing transaction from "commodity-bug.ledger", lines 6-10:
> 2005-08-01 Stock sale
>     Assets:Broker                                -50 AAPL {30.00} @ 50.00
>     Expenses:Broker:Commissions                19.95
>     Income:Capital Gains                   -1,000.00
>     Assets:Broker                           2,480.05
Unbalanced remainder is:
               -1000
Amount to balance against:
                2500
Error: Transaction does not balance
```

It appears to be incorrectly using the _sale price_ as the _lot price_. Because the transaction would be balanced if using either:

- `-50 AAPL @ 30.00` or
- `-50 AAPL {30.00} @ 30.00`

But, of course, those are not the correct way to represent the sale posting. They do not correspond with how the posting is represented when explicit commodities are used, which is: _amount_ _stock_ `{` _lot price_ `}` `@` _sale price_.

tbm:
[...]

PR #2354, opened 2024-06-28
Documented need for an explicit commodity in stock sale postings.

hoylen:
Added a sentence to the documentation to mention that implicit commodities do not work with stock sale transactions.

To address the issue reported in https://github.com/ledger/ledger/issues/2353. So users know and don't create ledger files that won't work properly.

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -1816,6 +1816,8 @@ The @samp{@{$30.00@}} is a lot price.  You can also use a lot date,
 same price/date and your taxation model is based on
 longest-held-first.
 
+The lot price and sale price must include an explicit commodity name.
+
 @node Fixing Lot Prices, Complete control over commodity pricing, Buying and Selling Stock, Currency and Commodities
 @subsection Fixing Lot Prices
 @cindex fixing lot prices


tbm:
I think ideally, in a section about amounts, there should be a paragraph saying that while commodity-less amounts work in basic scenarios, they shouldn't be used because a lot of stuff doesn't work if you don't have a commodity.

John made a remark about this a few years ago but I couldn't find it when I looked for it earlier today.

tbm:
For the record, I think that's a difference between ledger and hledger.  hledger applies the `D` directive to commodity-less amounts whereas ledger does not.  (Not that this matters for the issue at hand)

hoylen:
In the past, I think the commodity name was optional.

Certainly _ledger_ does not complain if it is  not present. And the basic features still work, which gives users a false sense of security, until they come across something that does not.

I see [section 3.4](https://ledger-cli.org/doc/ledger3.html#Commodities-and-Currencies) of the current manual now says, "Ledger makes no assumptions about the commodities you use; it only requires that you specify a commodity."

So I think my PR should be rejected. A better solution is for _ledger_ to raise a warning or error when the commodity name is missing... but that is a big change.

tbm:
> In the past, I think the commodity name was optional.

It's still optional but certain things don't work, as you found out, so it's not recommended

Issue #2330, opened 2024-02-16
Assertion failure when using group-by and amount


taviso:
Test ledger File:

```
2024/01/01 * Test1
    Expenses:Test                            $1.00
    Assets:Checking
2024/01/01 * Test2
    Expenses:Test                            $1.00
    Assets:Checking
```

```
$ ledger --args-only --file test.ldg bal --group-by 'payee' --invert ^Expenses:
terminate called after throwing an instance of 'ledger::assertion_failed'
  what():  Assertion failed in "ledger/src/op.h", line 260:void ledger::expr_t::op_t::release() const: refc > 0
```

```
$ ledger --version
Ledger 3.3.2-20230330, the command-line accounting tool
```

taviso:
I would like to fix this bug, I can see the problem is that in `post_t::add_to_value`, it tries to create a `bind_scope_t` where `scope->parent == scope`, so that just recurses infinitely in `ledger::bind_scope_t::define`.

I don't understand how it got to this state.

This is where it starts to recurse forever:

https://github.com/ledger/ledger/blob/master/src/scope.h#L188

```
(gdb) list
185
186       virtual void define(const symbol_t::kind_t kind, const string& name,
187                           expr_t::ptr_op_t def) {
188         parent->define(kind, name, def);
189         grandchild.define(kind, name, def);
190       }
(gdb) p this == parent
$1 = true
```

Here is the debug output before that happens:

```
   54ms  [DEBUG] value = (Expenses:Test)
              $-1.00  Expenses:Test
   54ms  [DEBUG] Binding scope 0x7ffffffebe80 with 0x81b5330 <-- This is creating the loop
   54ms  [DEBUG] Before compilation:
0x81b8510         O_SEQ (1)
0x81b8330          VALUE: null (1)
0x81b84c0          IDENT: __tmp_amount_expr (1)
0x81b8280           SCOPE: 0x81b7b00 (3)
0x81b8230            O_SEQ (1)
0x81b7cd0             VALUE: null (1)
0x81b81e0             O_SEQ (1)
0x81b8040              VALUE: null (1)
0x81b8190              IDENT: amount_expr (1)
0x81b7ff0               SCOPE: 0x81b7f00 (5)
0x81b7fa0                O_NEG (1)
0x81b7f50                 IDENT: amount_expr (1)
0x81b7c80                  SCOPE: 0x81b7b50 (1)
0x81b7c30                   IDENT: amount (1)
[...]

Issue #2343, opened 2024-05-08
historical option isn't handled well with grouping such as monthly


xkikeg:
Input

```
2023/12/01 Some Bank
    Assets:Bank                                 1000 JPY
    Assets:Bank                                -5.96 CHF

2020/01/01 Some Bank
    Assets:Bank                                 1000 JPY
    Assets:Bank                                -8.90 CHF

2024/01/10 Foo Shop
    Expenses:Grocery                            1000 JPY
    Assets:Bank

2024/01/20 Baz
    Expenses:Grocery                              10 CHF
    Assets:Bank

2024/01/25 Some Bank
    Assets:Bank                                 1000 JPY
    Assets:Bank                                -5.86 CHF

2024/02/10 Foo Shop
    Expenses:Grocery                            1000 JPY
    Assets:Bank

2024/02/20 Baz
    Expenses:Grocery                              10 CHF
    Assets:Bank

2024/03/01 Some Bank
    Assets:Bank                                 1000 JPY
    Assets:Bank                                -5.87 CHF

2024/03/10 Foo Shop
    Expenses:Grocery                            1000 JPY
    Assets:Bank

2024/03/20 Baz
    Expenses:Grocery
```

```shell
 % ledger -f historical.ledger reg Expenses:Grocery --historical --group-by payee
Baz
24-Jan-20 Baz                                       Expenses:Grocery                                                 10.00 CHF                 10.00 CHF
24-Feb-20 Baz                                       Expenses:Grocery                                                 10.00 CHF                 20.00 CHF
24-Mar-20 Baz                                       Expenses:Grocery                                                 10.00 CHF                 30.00 CHF
terminate called after throwing an instance of 'ledger::assertion_failed'
  what():  Assertion failed in "./src/op.h", line 259:void ledger::expr_t::op_t::release() const: refc > 0
zsh: IOT instruction  ledger -f historical.ledger reg Expenses:Grocery --historical --group-by paye
```

Posted onto https://groups.google.com/u/1/g/ledger-cli/c/z_drSTSPTmo, but I found it's caused by reusing stack address which is already destroyed. Let me send a PR once I get done.

xkikeg:
[...]

PR #2356, opened 2024-06-30
Restore expr context after calc

taviso:
Part of the `expr_t::compile()` process is to store the current scope, but In `post_t::add_to_value` that scope is temporary and on the stack.

Restore the original context after that process is complete.

Fixes: #2330 #2343

### DIFF
--- a/src/post.cc
+++ b/src/post.cc
@@ -642,11 +642,12 @@ void post_t::add_to_value(value_t& value, const optional<expr_t&>& expr) const
       add_or_set_value(value, xdata_->compound_value);
   }
   else if (expr) {
-    bind_scope_t bound_scope(*expr->get_context(),
-                             const_cast<post_t&>(*this));
+    scope_t *ctx = expr->get_context();
+    bind_scope_t bound_scope(*ctx, const_cast<post_t&>(*this));
 #if 1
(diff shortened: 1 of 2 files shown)


src/post.cc
```diff
@@ -642,11 +642,12 @@ void post_t::add_to_value(value_t& value, const optional<expr_t&>& expr) const
...
   }
   else if (expr) {
-    bind_scope_t bound_scope(*expr->get_context(),
-                             const_cast<post_t&>(*this));
+    scope_t *ctx = expr->get_context();
+    bind_scope_t bound_scope(*ctx, const_cast<post_t&>(*this));
 #if 1
     value_t temp(expr->calc(bound_scope));
     add_or_set_value(value, temp);
+    expr->set_context(ctx);
```
xkikeg, line 650: * If we can copy `expr` to `tmp_expr` and modify the `tmp_expr` to `bound_scope` that'll be cleaner, but `expr_t` is reference counted pointer and maybe not that easy
* `set_context` doesn't invalidate the `compile` flag, is that ok? (I don't have deep understanding, so just a speculation)

jwiegley:
This looks reasonable to me.

jwiegley (approved):

Issue #1765, opened 2019-02-14
`ledger bal --basis` not balancing to zero.


itsjohncs:
[I prepared a reduced test case](https://gist.github.com/itsjohncs/ab5d730f9f12c956e23520fb4509aab2).

jwiegley:
Well, that's a great find. Thank you.

simonmichael:
Slightly more reduced:
```
2019/01/01
    a  A0.5 @ $5.430   ; <- 2.715
    b  $-2.71

2019/01/01
    a  A-0.5 @ $10.93
    c  $2.77
    d  $2.695
```

tbm:
Thanks @simonmichael this is very insightful.  I wonder if this is due to #1663.

dmage:
I found a similar problem:

```
$ cat foo.dat
2019/12/01 * foo
    a  1430 XXX @ 1.0488 CZK
    b  -1499.78 CZK

2019/12/02 * bar
    a  1881 XXX @ 1.0629 CZK
    b  -1999.31 CZK

$ ledger -f foo.dat bal -B
         3499.10 CZK  a
        -3499.09 CZK  b
--------------------
            0.01 CZK
```

simonmichael:
What I think should happen above:
```
could not balance this transaction (real postings are off by $0.005)
2019/01/01
    a    A0.5 @ $5.430  ; $2.715
    b          $-2.710  ; checked for balancedness to 2 decimal places ($-2.710 would fail)
```

In hledger transaction balancedness is checked like this:

- sum the posting amounts at full precision
- render the sum at display precision
- does that appear to be zero (no non-zero digits displayed) ?

simonmichael:
Actually, better if that error message would show the $-2.71 as it was written in the journal.

I think the basic problem is that balanced-checking and display are using different precision. This case could be handled better, eg scanning the journal to find the display precision before doing any balanced checking, or by doing what hledger does. I'm not sure if it's possible to handle every case perfectly, though; there are still some where hledger will display numbers that appear not to balance.

dmage:
Two years later I think my example is incorrect and ledger should complain about such transactions.

```
2022/05/06
    a  1 XXX @ 0.995 CZK
    b  -1.00 CZK

2022/05/06
    a  1 XXX @ 0.995 CZK
    b  -1.00 CZK

2022/05/06
    a  1 XXX @ 0.995 CZK
    b  -1.00 CZK

2022/05/06
[...]

Issue #1663, opened 2018-07-03
0.005 rounded down instead of up


tbm:
3 * 1.595 = 4.785 which should imho be rounded up to 4.79 but ledger seems to round it down to 4.78.  I'm surprised because I'm sure someone else would have noted that already.
```
ledger -f rounding.ledger bal cash
           -4.78 EUR  Assets:Cash
```
Test case:
```
D 1000.00 EUR

; 3 * 1.595 is 4.785 which should imho be rounded to 4.79 but
; ledger shows 4.78.

2018-01-02 * Bought FOO
    Assets:Investments           3 FOO @ 1.595 EUR
    Assets:Cash
```

tbm:
I spoke to John about this yesterday:

```
20:34 < tbm> hey johnw is it expected that 0.955 gets rounded to 0.95 instead of 0.96?
22:03 < johnw> tbm: let me check
22:04 < johnw> tbm: https://github.com/ledger/ledger/blob/next/src/amount.cc#L685
22:04 < johnw> we use ceil(), so I would have thought it would be 0.96
22:05 < johnw> but maybe that's only for roundto
22:05 < tbm> yeah, 0.96 seems correct to me but that's not what I got
22:06 < johnw> oh, it's this:
22:06 < johnw> mpfr_asprintf(&buf, "%.*RNf", precision, tempfb)
22:06 < johnw> https://github.com/ledger/ledger/blob/next/src/amount.cc#L144
22:06 < johnw> according to MPFR docs: 'N' round to nearest
22:06 < johnw> so yeah, we round toward 0
22:07 < johnw> or rather, in this case we should round away from 0....
22:07 < johnw> hmm
22:07 < johnw> maybe some debug code in stream_out_mpq would clarify this
22:07 < johnw> we also use GMP_RNDN, which should also mean nearest
22:07 < johnw> I wonder if something gets lost in the math that we do there
22:10 < johnw> you can set --debug=amount.convert
22:11 < tbm> Ok, I'll try that another day
22:15 < johnw> but please open a bug on this
22:15 < johnw> we shouldn't lose this test case that you have
22:16 < tbm> yeah, will do tomorrow
```

tbm:
With `--debug=amount.convert`:

```
   14ms  [DEBUG] Rational to convert = -957/200
   14ms  [DEBUG] num prec = 394
   15ms  [DEBUG] den prec = 392
[...]

Issue #2329, opened 2024-02-14
Error: Balance assertion off by $-0.00


maxnikulin:
I have puzzled by the following error message having just zeroes in discrepancy amount and I have no idea how to debug the issue:
```
While parsing file "/tmp/roundto.ledger", line 5:
While parsing posting:
  Account  0  = $-700.67
                ^^^^^^^^
Error: Balance assertion off by $-0.00 (expected to see $-700.67)
```
the test file is
```
2024-02-01 Transaction
        Account   (-roundto($12.34 * 56.78, 2))
        Expenses  $700.67
2024-02-02 Assertion
        Account  0  = $-700.67

```
The idea is to check that billing is correct: per unit price, volume and total amount are in agreement.

An attempt to specify excessive precision
```
commodity $
        format $1000.0000000000000
```
failed with a similar message with more zeroes
```
While parsing file "/tmp/roundto.ledger", line 5:
While balancing transaction from "/tmp/roundto.ledger", lines 3-5:
> 2024-02-01 Transaction
>       Account   (-roundto($12.34 * 56.78, 2))
>       Expenses  $700.67
Unbalanced remainder is:
    $0.0000000000000
Amount to balance against:
  $700.6700000000000
Error: Transaction does not balance
```

```
ledger --version
Ledger 3.3.0-20230208, the command-line accounting tool
```
Earlier (ledger-3.1.3) I used a bit annoying trick with no `roundto` call and an extra posting to correct rounding and an automatic transaction adding a virtual posting to cancel rounding correction. Currently another trick is required since virtual and regular accounts have diverged and assertions work in different way (#1959).

How to find the real discrepancy hidden behind zeroes?

maxnikulin:
> This is a precision problem.
> 
> I read somewhere — maybe in the docs — that ledger defaults to 6 places of precision.

My expectation from user point of view is that `roundto()` result should match parsing decimal number having similar precision. IIRC converting from decimal numbers is deterministic in IEEE 754. Perhaps it is something with low level rounding mode in `roundto()` implementation.

> Stepping through this example, the diff is below.

Thanks, however my another expectation is that a tool should provide enough information to figure out what happens without running a debugger. I have ledger installed using `apt install ledger`. Likely I can add the repository with debug symbols, but I suspect it would still inconvenient with configure debug option disabled. I would prefer to avoid building tools from source.

> (lldb) p diff.to_amount().to_double()
> (double) -4.0927261579781769E-14

[...]

Issue #1983, opened 2020-12-29
roundto() rounding issue: negative and positive not the same


tbm:
```
ledger eval 'roundto(10 * 0.0775, 2)'
0.78
ledger eval 'roundto(-10 * 0.0775, 2)'
-0.77
```

You get 0.78 and -0.77, which means that certain transactions won't balance.

Here's a test case:

```
= /^Expenses:Pizza$/
    Expenses:Pizza:Taxes           (roundto(amount * 0.0775, 2))
    Liabilities:CreditCard        (roundto(-amount * 0.0775, 2))

2020-12-29 * Test
    Expenses:Pizza                   10.00 USD
    Liabilities:CreditCard          -10.00 USD
```

gives:

```
Unbalanced remainder is:
            0.01 USD
Amount to balance against:
           10.78 USD
While applying automated transaction ...
```

Note that this issue is different to the rounding issue in #1663

the-solipsist:
This bug arises because `roundto` uses the function `ceil()`: https://github.com/ledger/ledger/blob/next/src/amount.cc#L685

The `ceil()` function rounds up towards +∞, as seen in [the examples on this page](https://www.geeksforgeeks.org/ceil-floor-functions-cpp/).  For financial accounting, one should use a rounding function rounds up/down away from 0, as I explained in a [comment to another issue](https://github.com/ledger/ledger/issues/1663#issuecomment-752182284).

the-solipsist:
If `ceil()` is used internally for purposes other than providing the `roundto()` function for users, then that should be fixed.

However, insofar as `roundto()` is concerned, an easy workaround (as pointed out [on the mailing list](https://groups.google.com/g/ledger-cli/c/f47jLxyLwyQ/m/x4xSF8oEBgAJ)), is to place to the minus sign outside of the `roundto()` function.

```
ledger eval 'roundto(10 * 0.0775, 2)'
0.78
ledger eval '(-roundto(-10 * 0.0775, 2))'
-0.78
```

If it isn't worth replacing the `ceil()` function, then this behaviour should be documented in the manual.

maxnikulin:
To my surprise there is a unit test that fails if `roundto` behavior in respect to sign inversion is fixed:
https://github.com/ledger/ledger/commit/185656e191#diff-13332abb7fc3cbdcab10016305ee16e23a7be2ac9229d6b29b6ff633e530d9acR8231-R8237
```
Assets:Cash ¤ -123,40
[...]

PR #2361, opened 2024-07-08
Fix denominator of roundto result

maxnikulin:
Multiprecision rational created from a double value may have large power of 2 denominator since fractional decimal numbers can not be represented as binary floating point numbers. It leads to failed assertion when result is compared to a value converted directly from strings. Use integer multiprecision arithmetics to round numbers to ensure proper denominator. Inspired by python gmpy2 package <https://github.com/aleaxit/gmpy/blob/3e4564ae9d/src/gmpy2_mpq_misc.c#L315>

See #2329

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -8660,7 +8660,7 @@ $ ledger -f expr.dat --format "%(account) %(roundto(amount, 1))\n" reg
 @end smallexample
 @smallexample @c output:B4DFB9F
 Assets:Cash ¤ -123,40
-Expenses:Office Supplies ¤ 123,50
+Expenses:Office Supplies ¤ 123,40
 @end smallexample
 @end defun
 
--- a/src/amount.cc
+++ b/src/amount.cc
@@ -684,8 +684,45 @@ void amount_t::in_place_roundto(int places)
 {
   if (! quantity)
     throw_(amount_error, _("Cannot round an uninitialized amount"));
-    double x=ceil(mpq_get_d(MP(quantity))*pow(10, places) - 0.49999999) / pow(10, places);
-    mpq_set_d(MP(quantity), x);
+
+  // <https://github.com/ledger/ledger/issues/2362>
+  // _dup() call (see in_place_ceiling and in_place_floor)
+  // leads to 2 failures in the balance/testRound test,
+  // however in its absence this method affects copies of amount_t instances.
+  // Remove expected_failures from amount/testRound
+  // and update balance/testRound
+  // when uncommenting the following line.
+  // _dup();
+
+  mpz_t& scale(temp);
+  if (places)
+    mpz_ui_pow_ui(scale, 10, labs(places));
+
+  if (places > 0) {
+    mpz_mul(mpq_numref(MP(quantity)), mpq_numref(MP(quantity)), scale);
+  } else if (places < 0) {
+    mpz_mul(mpq_denref(MP(quantity)), mpq_denref(MP(quantity)), scale);
+  }
+
+  auto whole(mpq_numref(tempq));
+  auto reminder(mpq_denref(tempq));
+  mpz_fdiv_qr(whole, reminder, mpq_numref(MP(quantity)), mpq_denref(MP(quantity)));
+  mpz_mul_2exp(reminder, reminder, 1);
+  const int rem_denom_cmp = mpz_cmp(reminder, mpq_denref(MP(quantity)));
+  if (rem_denom_cmp > 0
+      || (rem_denom_cmp == 0 && mpz_odd_p(whole)))
+    mpz_add_ui(whole, whole, 1);
+
+  if (places > 0) {
+    mpq_set_num(MP(quantity), whole);
+    mpq_set_den(MP(quantity), scale);
+    mpq_canonicalize(MP(quantity));
+  } else if (places == 0)
+    mpq_set_z(MP(quantity), whole);
+  else {
+    mpq_set_ui(MP(quantity), 0, 1);
+    mpz_mul(mpq_numref(MP(quantity)), whole, scale);
+  }
 }
 
 void amount_t::in_place_unround()
--- a/test/regress/2329.test
+++ b/test/regress/2329.test
@@ -1,0 +1,83 @@
+
+; roundto() result should not have large power 2 denominator.
+; Pairs to test positive/negative value symmetry.
+assert roundto(1.1, 1) == 1.1
+assert roundto(-1.1, 1) == -1.1
+
+; positive places
+assert roundto(1.13, 1) == 1.1
+assert roundto(1,17, 1) == 1.2
+assert roundto(1.10, 1) == 1.1
+assert roundto(-1.13, 1) == -1.1
+assert roundto(-1,17, 1) == -1.2
+assert roundto(-1.10, 1) == -1.1
+
+; zero places
+assert roundto(0.12, 0) == 0
+assert roundto(123.89, 0) == 124
+assert roundto(-0.12, 0) == 0
+assert roundto(-123.89, 0) == -124
+
+; negative places
+assert roundto(123.45, -1) == 120
+assert roundto(98765, -3) == 99000
+assert roundto(24500, -2) == 24500
+assert roundto(-123.45, -1) == -120
+assert roundto(-98765, -3) == -99000
+assert roundto(-24500, -2) == -24500
+
+; round to even
+assert roundto(0.5, 1) + roundto(1.5, 1) + roundto(2.5, 1) + roundto(3.5, 1) + roundto(4.5, 1) == roundto(0.5 + 1.5 + 2.5 + 3.5 + 4.5, 1)
+assert roundto(-0.5, 1) + roundto(-1.5, 1) + roundto(-2.5, 1) + roundto(-3.5, 1) + roundto(-4.5, 1) == roundto(-0.5 - 1.5 - 2.5 - 3.5 - 4.5, 1)
+
+assert roundto(0.12345, 4) == 0.1234
+assert roundto(0.9875, 3) == 0.988
+assert roundto(-0.12345, 4) == -0.1234
+assert roundto(-0.9875, 3) == -0.988
+assert roundto(1234.5, 0) == 1234
+assert roundto(367.5, 0) == 368
+assert roundto(-1234.5, 0) == -1234
+assert roundto(-367.5, 0) == -368
+assert roundto(29500.00, -3) == 30000
+assert roundto(72250, -2) == 72200
+assert roundto(-29500.00, -3) == -30000
+assert roundto(-72250, -2) == -72200
+
+; Assertion in transactions, no error:
+comment
+
+While parsing posting:
+  Account:NOK  0  = -700.67 NOK
+                    ^^^^^^^^^^^
+Error: Balance assertion off by -0.00 NOK (expected to see -700.67 NOK)
+end comment
+
+2024-02-01 Regular precision
+    Account:NOK   (-roundto(12.34 NOK * 56.78, 2))
+    Expenses:NOK  700.67 NOK
+2024-02-02 Assertion
+    Account:NOK  0  = -700.67 NOK
+
+; Transaction balance, no error:
+comment
+
+> 2024-02-01 Excessive precision
+>     Account:USD   (-roundto($12.34 * 56.78, 2))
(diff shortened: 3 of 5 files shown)


jwiegley:
Love it, once the tests pass!

tbm:
@maxnikulin can you take a look at the test failure.

maxnikulin:
Almost certainly tests are failing because I added `_dup()` call similar to `amount_t::in_place_floor()` and `amount_t::in_place_ceiling()`. Is the following behavior expected?
```c++
balance_t b;
amount_t a("$ 123.123");
b += a;
// b.to_string() == "$ 123.123"
a.in_place_roundto(2);
// b.to_string() == "$ 123.120"
```
I am surprised that changing amount after adding it to a balance affects balance value.

Do you have any idea concerning content of ledger file that is sensitive to `_dup` call during ceiling/floor/round operations? I do not like current inconsistency.

maxnikulin:
The following tests fail with original code and I do not like that `in_place_roundto()` affects the variable from which amount was copied. I believe, my current variant is more correct:
```c++
  amount_t a1("$ 123.123");
  amount_t a2(a1);
  a2.in_place_roundto(2);
  // Original version fails due to denominator difference with
  // error: in "balance/testRoundMinimal": check amount_t("$ 123.120") == a2 has failed [$ 123.120 != $ 123.120]
  // But why not "$ 123.12"?
  BOOST_CHECK_EQUAL(amount_t("$ 123.120"), a2);
  // Original version fails with
  // error: in "balance/testRoundMinimal": check amount_t("$ 123.123") == a1 has failed [$ 123.123 != $ 123.120]
  BOOST_CHECK_EQUAL(amount_t("$ 123.123"), a1);
```

By the way, `amount_t::roundto(int)` ignores its argument. Looks like another bug.

maxnikulin:
I have update the commit, `_dup()` is not called any more. I have created #2362 to track it.

I had to update another test case. I suppose `-123,40` vs. `123,50` rounding for the same absolute value is bug #1983 and this change fixes it.

https://github.com/ledger/ledger/blob/92db49d2e0d649c31b93be1e0cd0b5cde5ce57d8/doc/ledger3.texi#L8658-L8664

https://github.com/ledger/ledger/blob/92db49d2e0d649c31b93be1e0cd0b5cde5ce57d8/doc/ledger3.texi#L8430-L8435

jwiegley:
Your assessment of the issues around `_dup` makes sense to me, and I'd definitely like to see this addressed in a follow-up PR. Thank you for giving this so much thought.

maxnikulin:
> Your assessment of the issues around `_dup` makes sense to me, and I'd definitely like to see this addressed in a follow-up PR. Thank you for giving this so much thought.

I am sorry, but `round()` and `unround()` are beyond my knowledge related to ledger and these functions are not documented. I have no idea what is purpose of their magic. That is why I can not guess what `balance/testRound` is supposed to test (ledger/ledger@c5f41b8519). Unfortunately tripun who made greet job adding tests has not been active for a decade.

My questions related to the change:
1. Are you OK that I modified expectation of a unit test?
2. Do you agree with round to even implemented by this commit and thus
   ```
   build/ledger -f /dev/null eval 'roundto(0.005, 2)'
   0
   ```
   or do you prefer
   ```
   ledger -f /dev/null eval 'roundto(0.005, 2)'
   0.01
   ```
   like in #1663 and
   ```
   ledger -f /dev/null eval -- '-roundto(0.005 + 0.015 + 0.025 + 0.035 + 0.045, 2) + roundto(0.005, 2) + roundto(0.015, 2) + roundto(0.025, 2) + roundto(0.035, 2) + roundto(0.045, 2)'
   0.02
   ```
   The proposed commit gives
   ```
   build/ledger -f /dev/null eval -- '-roundto(0.005 + 0.015 + 0.025 + 0.035 + 0.045, 2) + roundto(0.005, 2) + roundto(0.015, 2) + roundto(0.025, 2) + roundto(0.035, 2) + roundto(0.045, 2)'
   0
   ```
   It should be possible to avoid `0.49999999` magic constant, but achieve 0.005 to 0.01 and -0.005 to -0.01 rounding.
3. In second patch revision I tried to add aliases to `temp` and `tempq`, but perhaps it should be reverted since `mpz_t` is a kind of trick while actual `__mpz_struct` is not a part of API.

jwiegley:
Amounts are infinite precision internally (stored as fractions), but they are _displayed_ at the precision observed in relation to the commodity.

`round()` causes this rounded to occur immediately and then be stored back into the fraction. For example, if the amount were `$0.3333333333333333`, this would be represented by the fraction `1/3`. It might display as `$0.33`.

`round()` would change the fraction from `1/3` to `33/100`.

`unround()` tells that particular amount to never use commodity-driven rounding logic, but to display with as much precision as is allowed by the context.

Strangely (now that I think about it), there is no way to undo the effect of `unround()`. It should probably be a method like `autoround()` or some such.

maxnikulin:
> `round()` causes this rounding to occur immediately

Unfortunately I can not figure out what I have missed. `amount_t::in_place_round()` just sets a flag that affects `quantity->prec` after following operations. It does not affect arithmetic.

That is why I can not get the point of the following assertions in `balance/testRound`:
```
(0.00 + $ 123.123 + EUR 456.567).rounded() ==
 0.00 + $ 123.120 + EUR 456.570
```
and the same with `in_place_round()`.

The test looks like a copy-paste of `testCeiling` or `testFloor` that is not split into `round` and `roundto`. The latter is more suitable for the used checks.

jwiegley:
Yeah, that test is very fishy. It is asserting something that is not generally true:
```
(x + y).round() == x.round() + y.round()
```

maxnikulin:
> Yeah, that test is very fishy.

Do you have any idea what test for `balance_t` may be meaningful for `rounded()`/`in_place_round()` and similar code?

It may be made valid if `roundto(2)` and `in_place_roundto(2)` are used instead. In addition it requires a fix in `amount_t::roundto(int)` however.

Out of curiosity I have tried `amount_t::in_place_truncate()` implemented as `in_place_roundto(display_precision())` from this pull request to avoid conversion to and from string. It seems no tests are failing.

maxnikulin:
I have rebased the commit to master and added references to #2362.

Attempts of further fixes would either depend on this pull request or would conflict with it. Feel free to cherry-pick specific chunks and close this pull requests. I am realizing that sometimes it is not easy to synchronize intervals of activity.

jwiegley:
So, materially, how do you think this could affect people's current data? I imagine we will see reported amounts changing if we merge this?

maxnikulin:
> So, materially, how do you think this could affect people's current data? I imagine we will see reported amounts changing if we merge this?

It is hard to reason for me since my attempt of using `roundto()` in a ledger file failed. I had a hope to use it as a workaround for a case close to [tbm. Rounding in ledger. Jun 24, 2014, 5:06:19 AM](https://groups.google.com/g/ledger-cli/c/m-TgILbfrwA/). I am trying to use assertions as extensive as possible to catch typos in positngs and errors in automatic transactions. Certainly behavior will change for people using `roundto()` in custom reports formats. At least in some cases output will be less confusing, like in the example in the docs (see above).

Again, if you strongly believe that halves should be rounded away from zero, not to nearest even then it can be tried.

For a while, comparison of results for some expressions evaluated by original and patched versions is the best what I can provide:
```
ledger eval "..."                    orig        fix match

# Denominator of rounded value
roundto(1.1, 1) == 1.1                  0          1     -
(roundto(1.1, 1) - 1.1)*1,000,000,000,000,000,000
\_                                   88.8          0     -
roundto(11, 0) == 11                    1          1     +
roundto(123, -1) == 120.00              1          1     +

# Rounding of halves for positive and negative values
roundto( 2.125, 2)                   2.13       2.12     -
roundto(-2.125, 2)                  -2.12      -2.12     +
roundto( 3.135, 2)                   3.14       3.14     +
roundto(-3.135, 2)                  -3.13      -3.14     -

# Accumulated error
## Alternating
0.005 + 0.015 + 0.025 + 0.035        0.08       0.08     +
roundto( 0.005, 2) + roundto( 0.015, 2) + roundto( 0.025, 2) + roundto( 0.035, 2)
\_                                    0.1       0.08     -
roundto(-0.005, 2) + roundto(-0.015, 2) + roundto(-0.025, 2) + roundto(-0.035, 2)
\_                                  -0.06      -0.08     -
## Odd
1.015 + 1.035 + 1.055 + 1.075        4.18       4.18     +
roundto( 1.015, 2) + roundto( 1.035, 2) + roundto( 1.055, 2) + roundto( 1.075, 2)
\_                                    4.2        4.2     +
roundto(-1.015, 2) + roundto(-1.035, 2) + roundto(-1.055, 2) + roundto(-1.075, 2)
\_                                  -4.16       -4.2     -
## Even
2.005 + 2.025 + 2.045 + 2.065        8.14       8.14     +
roundto( 2.005, 2) + roundto( 2.025, 2) + roundto( 2.045, 2) + roundto( 2.065, 2)
\_                                   8.16       8.12     -
roundto(-2.005, 2) + roundto(-2.025, 2) + roundto(-2.045, 2) + roundto(-2.065, 2)
\_                                  -8.12      -8.12     +

# Close to half values
roundto(1.1499999999,  1)             1.2        1.1     -
roundto(1.1499000,     1)             1.1        1.1     +
roundto(2.2499999999,  1)             2.3        2.2     -
roundto(2.2499000,     1)             2.2        2.2     +
roundto(-2.1500000001, 1)            -2.1       -2.2     -
roundto(-2.15001,      1)            -2.2       -2.2     +
roundto(-3.2500000001, 1)            -3.2       -3.3     -
roundto(-3.25001,      1)            -3.3       -3.3     +
```

jwiegley:
I certainly think that your numerical tests should define the standard of truth, and the implementation should measure up to them, or else what we have today is wrong.

I'm inclined to accept this Pull Request. Is there anyone who has an objection, or feels that the current behavior should in any way be preserved?

tbm:
> Is there anyone who has an objection 

I'm not sure how many people are following this PR. Maybe this change should be mentioned on the mailing list.

But it sounds like this will fix #1663 which I feel personally very strongly about. I think it should also fix #1765 (@maxnikulin can you check?) which is another great bug report showing the broken behaviour (`-B` not balancing to zero).

maxnikulin:
> > Is there anyone who has an objection
> 
> I'm not sure how many people are following this PR. Maybe this change should be mentioned on the mailing list.

This is a public pull request, so feel free to discuss it in any channel you find appropriate.

> But it sounds like this will fix #1663 which I feel personally very strongly about.

As I have already written, the expectation from the original report "3 * 1.595 = 4.785 which should imho be rounded up to 4.79" is not fulfilled. I am not sure what should be concluded from other comments to that issue.

Notice:
```
roundto( 2.125, 2)                   2.13       2.12     -
```
If you expect a different result, please, specify desired values for the cases from my previous comment and maybe add some more corner cases. If requirements are consistent then the code may be adjusted.

> I think it should also fix #1765 (@maxnikulin can you check?) which is another great bug report showing the broken behaviour (`-B` not balancing to zero).

It is a higher level issue.

This pull request is intended to address #2329, #1983, and #1925 with the following exception:
```
Error: Assertion failed: roundto(12.345, 2) == 12.35
```
however
```
ledger eval 'roundto(12.345, 2) == 12.34'
1
```

jwiegley:
I think our rounding behavior should not be based on historical expectations, but on numeric tables. It is very unfortunate that I never implemented such tests in order to nail down the semantics of `roundto`. So, even though this may change some past report outputs, I think we should accept this PR based on the test results it declares.

jwiegley (approved):

maxnikulin:
> @maxnikulin What is your suggestion for next steps?

@jwiegley, sorry for disturbing you, I have no suggestions yet. I am still confused, otherwise I would create another issue or even a pull request.

Yesterday I was filing another bug when I realized that I have some changes in my local clone of the repository. Some time ago I tried to eliminate floating point arithmetic from `amount_t::in_place_unreduce` in a similar way as it was done for roundto in this pull request. From my point of view, the code of unreduce resembles what you described in response to my question concerning `round()`. That is why I decided to ask if I completely missed the point of round/unround and reduce/unreduce or you described behavior of unreduce, but by mistake named it unround.

Actually I do not use reduce/unreduce, so my motivation to improve unreduce almost exhausted when realized that it may require more than a few lines of code. We may just stop discussion here.

By the way, this pull request was caused by #2329. You may close it if you think, there is no point in improving diagnostics for similar cases.

PR #2363, opened 2024-07-12
Add display_account_note format element

matclab:
It allows to display account note in register command (and not only the
post or transaction note)

### DIFF
--- a/src/post.cc
+++ b/src/post.cc
@@ -338,6 +338,12 @@ namespace {
     }
     return acct;
   }
+  value_t get_display_account_note(call_scope_t& args)
+  {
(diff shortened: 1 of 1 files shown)


matclab:
Shall I open a ticket accordingly ?

matclab:
Ok, in the process of reading the code, I just understand that `display_account.note` does the trick !!!

tbm:
Can you check if this is properly documented, and if not, submit a PR with a doc improvement?

tbm:
I don't know either, sorry

matclab:
First try in #2364.

Looks like there are other functions in need of doc.

Issue #2078, opened 2021-12-31
Ledger-cli in WebAssembly


MisterY:
Perhaps a crazy idea but an interesting one, nonetheless. Is it possible to compile ledger-cli into wasm?
Has anyone tried, what are the challenges? Opening this issue as a placeholder for the topic.

The benefits should be obvious, I guess. Having Ledger compiled to WASM would allow running it natively in browsers, meaning PWAs and also Android.
I've seen some people compiling Boost with Emscripten but I'm quite new to the topic at the moment.

MisterY:
Seems there is also a virtual filesystem available, which is convenient for storing and reading the files.

- https://stackoverflow.com/questions/47313403/passing-client-files-to-webassembly-from-the-front-end
- https://kripken.github.io/emscripten-site/docs/porting/files/file_systems_overview.html

jwiegley:
Interesting idea! It would permit me to run Ledger directly on the Internet Computer too.

MisterY:
Filesystem extension
https://github.com/jvilk/BrowserFS

MisterY:
# Boost 2 WebAssembly links

- https://stackoverflow.com/questions/57046348/emscripten-and-boost-library-how-to-compile-existing-project-for-webassembly

MisterY:
Link to [WebAssembly support](https://github.com/boostorg/context/issues/109) in Boost repo.

alensiljak:
There is already #2004 but nowadays it seems that a WASI target would be a perfect fit for Ledger-CLI. 

In addition, OPFS is now available is all major browsers, effectively providing the conditions for Ledger to run in a browser, using wasm target.

alensiljak:
There is also Chirp - https://github.com/leaningtech/cheerp-meta - compiler.

gudzpoz:
I just managed to get ledger compiled with emscripten following the steps stated in https://github.com/ledger/ledger/issues/2004#issuecomment-1878164915 with changes in https://github.com/gudzpoz/ledger/commit/3d131345ae86616b92c0e35013698e14a0089114 applied.

By the way, currently it seems to me that emscripten is the de facto compiler to port C++ code to WASM:

[...]

Issue #2004, opened 2021-02-12
Feature request: support for wasm/emscripten compilation targets


jibcage:
(If someone has gotten this working already, please let me know!)

Having the `acprep` script support configuring compilation for use with emscripten would make it possible to embed libledger and ledger commands into web applications, and other applications that use Electron as a backend.

As far as manual steps go, I have gotten as far as using [this guide](https://stackoverflow.com/questions/41080815/compiling-gmp-mpfr-with-emscripten/43583154) to get gmp and mpfr compiled, and have also been able to leverage libboost's native emscripten target support to get a static libboost include dir. The cmake command below seems to work okay:

```
$ emcmake cmake -DBoost_LIBRARY_DIR=${HOME}/boost_1_75_0/stage/lib -DBoost_INCLUDE_DIR=${HOME}/boost_1_75_0/ -DMPFR_PATH=${HOME}/opt/wasm/include/mpfr.h -DMPFR_LIB=${HOME}/opt/wasm/lib/libmpfr.a -DGMP_PATH=${HOME}/opt/wasm/include/gmp.h -DGMP_LIB=${HOME}/opt/wasm/lib/libgmp.a -DUTFCPP_INCLUDE_DIR=${PWD}/utfcpp/source .
```

However, once I get past cmake's configuration and actually try compiling, I get:

```
$ emmake make
make: make
[  1%] Generating ../system.hh.pch
/home/jack/ledger/system.hh:133:10: fatal error: 'gmp.h' file not found
#include <gmp.h>
         ^~~~~~~
1 error generated.
[...]

Issue #2320, opened 2023-12-20
Reading a second journal from Python causes an assertion failure


andrew521:
The issue has been raised multiple times:

https://github.com/ledger/ledger/issues/1953
https://github.com/ledger/ledger/issues/514
https://github.com/ledger/ledger/issues/974
https://github.com/ledger/ledger/issues/973
https://github.com/ledger/ledger/issues/972

exemple:
>>> journal = ledger.read_journal(j_file)                  # first time works fine
>>> journal = ledger.read_journal(j_file)
*** RuntimeError: Assertion failed in "/home/*****/ledger-3.3.2/src/session.cc", line 183:std::size_t ledger::session_t::read_data(const string&): xact_count == journal->xacts.size()

In the old tickets, closing the jurnal was recommended a couple of times, but it actually crashes python:

>>> ledger.close_journal_files()                               # Segmentation fault

Wondering if anyone can provide a fix? TIA.

afh:
Thanks for taking the time to create a new issue, what was the motivation to create this duplicate issue over commenting on #514? I'm uncertain whether I see the benefit of it? Care to provide more context?

andrew521:
Given that issue #514 has not received substantial attention for over a decade, I decided to enhance the visibility of the problem in the hope of garnering more attention and potential resolution.

afh:
I'd prefer to close duplicate issues in favor of the first one filed as information about a problem is otherwise scattered across issues, making it more difficult to get an understanding of the problem.
Hence I kindly ask we take the conversation over to #514 and if you'd be willing to provide a [MWE](https://en.wikipedia.org/wiki/Minimal_reproducible_example) and add the various relevant information from related issues to #514 perhaps, that would be much appreciated.
I'm open to increase visibility on #514 by adding a priority label.

andrew521:
I will watch https://github.com/ledger/ledger/issues/514 for resolution then. Shall I close this one?

Issue #514, opened 2018-01-24
Reading two journals from Python causes an assertion failure


cpitclaudel:
Assume the following two ledger files:

`1.ledger`

```
2017-01-01 A
  A  $1
  B
```

`2.ledger`
```
2017-02-01 C
  C  $1
  D
```

Running `python2 -c 'import ledger; ledger.read_journal("1.ledger"); ledger.read_journal("2.ledger")'` crashes with the following error:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
RuntimeError: Assertion failed in "/build/ledger-2L3TES/ledger-3.1.1+dfsg1/src/session.cc", line 182:std::size_t ledger::session_t::read_data(const string&): xact_count == journal->xacts.size()
```

afh:
Thank you for reporting @cpitclaudel. We are aware of the issue, and it has been bothering me quite a bit. I hope that we can address this issue alongside with #510.

tbm:
This is a duplicate of #973

PR #2366, opened 2024-07-19
Support compilation into WASM(+JS) with Emscripten

gudzpoz:
This PR should enable the user to compile ledger using Emscripten. The changes in this PR should have been kept minimal and expects the user to use `-DCMAKE_CXX_FLAGS=...` to customize the build using the tons of flags Emscripten provides.

This PR does not attempt to put the dependency preparing steps into `acprep` because they are simply too hard to get right, unless hard-coding dependency versions is acceptable. (For example, older Boost will break on newer Emscripten versions.) However, this puts the burden onto the users, and maybe we should put a line in README to point the users to the `cmake.yml` for detailed compilation steps using Emscripten?

See also #2004 and #2078.

### Changes Explained

I will try to summarize the changes and explain why they are necessary:
- **CMake Build Scripts**: The following files are changed:
  - `CMakeLists.txt`:
    - In `test/unit/*` tests, `Boost::unit_test_framework` is expected to be linked dynamically. But currently with Emscripten, dynamic linkage causes much trouble. And here I simply exclude those tests when compiling under Emscripten.
      - The original configuration links ledger against boost::unit_test_framework, resulting in linkage errors with Emscripten.  It is only used in unit tests and is safe to move the FindBoost command there.
    - `boost/regex/icu.hpp` requires ICU. But recently ICU have upgraded to C++ 17 and fails to compile under C++ 11.
  - `src/CMakeLists.txt`: Emscripten does not handle C++ exceptions by default, and requires special flags (`-fexceptions` or `-fwasm-exceptions`) to make exceptions work. The former one requires JavaScript, and is slower. The latter one is still WIP but implemented in most browsers & Node.
  - `test/CMakeLists.txt`: See above. Excluding `test/unit/*` when under Emscripten.
  - `test/unit/CMakeLists.txt`: See above. Excluding `test/unit/*` when under Emscripten.
- **Code Changes**:
  - `src/global.cc`:
    - No `fork` support in Emscripten.
    - `exit` would shut down Emscripten/WASM runtime, preventing the user from rerunning the main function.
  - `src/main.cc`:
    - No `signal` support in Emscripten.
  - `src/quotes.cc`:
    - No pipe support in Emscripten.
  - `src/report.cc`:
    - No `isatty` support on ancient versions of Emscripten. But `--force-color` will be needed to pass the BaselineTest_opt-color test. (Also, `--force-color` should enforce the color escape codes, right?)
  - `src/stream.cc`:
    - No pipe support in Emscripten.

A large part of this PR actually lies in the changes to `cmake.yml`, which compiles Ledger with Emscripten and runs (most of) the tests. A few points:
- Boost 1.84 fails to build under Emscripten 3.1.61 (the most recent version), because Boost 1.84 expects Emscripten to output `.bc` files while Emscripten now outputs `.a` files only.
  To avoid similar problems, the workflow fixed the version numbers to `1.85` and `3.1.61` respectively.
- We use `-DCMAKE_CXX_FLAGS=" -lnodefs.js -lnoderawfs.js "` to grant direct FS access to the WASM build. This is required for the tests to run, but is not necessary for other users (e.g., who embeds Ledger/WASM in their own applications, isolated).
- Emscripten has a bug that prevents the generated `ledger.js` from running. (https://github.com/emscripten-core/emscripten/issues/22264) I hard-coded a workaround into the workflow.
- `DocTests.py` seems to run fine when run from the command line, but fails when called from CTest. I guess it is probably due to the stdin/out handling in Emscripten.
- Several tests are excluded. Reasons for that have been documented in the workflow file:
  ```
          # _py$ : Python tests are not supported.
          # DocTestsTest_ledger3 : See above.
          # RegressTest_BF3C1F82 : Emscripten treats 2012/02/30 as a valid date.
          # BaselineTest_opt-output : Outputing to /dev/stderr, probably unsupported by Emscripten.
  ```


CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ endif()
 
 # Set BOOST_ROOT to help CMake to find the right Boost version
 find_package(Boost ${Required_Boost_Version}
-  REQUIRED date_time filesystem system iostreams regex unit_test_framework
+  REQUIRED date_time filesystem system iostreams regex
```
afh, line 87: Very nice clean-up with the change in test/unit/CMakeLists.txt below 👌

CMakeLists.txt
```diff
@@ -163,6 +163,7 @@ endif()
 
 cmake_push_check_state()
 
+set(CMAKE_REQUIRED_FLAGS -std=c++11)
```
afh, line 166: Why is this necessary if `add_compile_options(-std=c++11)` exists in [CMakeLists.txt:24](https://github.com/ledger/ledger/blob/master/CMakeLists.txt#L24)?

src/global.cc
```diff
@@ -476,7 +476,7 @@ void global_scope_t::visit_man_page() const
   int status = -1;
   wait(&status);
 #endif
-  exit(0);                      // parent
+  throw error_count(0, "");     // parent
```
afh, line 479: Can you provide more context on this change, please?

src/report.cc
```diff
@@ -70,6 +70,10 @@ void report_t::normalize_options(const string& verb)
     if (HANDLED(pager_) && ! isatty(STDOUT_FILENO))
       HANDLER(pager_).off();
   }
+#else
+  if (HANDLED(force_color)) {
+    HANDLER(color).on("?normalize");
+  }
```
afh, line 76: At first glance this appears unrelated to emscripten, what is the motivation for this change?

afh (commented):
This is very impressive work, @gudzpoz 👏, thanks for this!!

Please find below a few comments and questions to help understand the changes.

In general what I think would be very helpful is adding examples how to use ledger in wasm(+js) with emscripten and expanding the documentation, especially ledger's texinfo manual.

What are your thoughts?

.github/workflows/cmake.yml
```diff
@@ -73,3 +73,125 @@ jobs:
...
+          sed 's/if(!stream.tty)/if(!stream.tty||!stream.tty.ops)/g' ledger.js > ledger.js.ioctl_fix
+          sed 's/,tty:true,/,tty:false,/g' ledger.js.ioctl_fix > ledger.js.fs_fix
+          # Enable testing from python scripts.
+          echo '#!'"$(which node)" > ledger.js
+          cat ledger.js.fs_fix >> ledger.js
+          chmod a+x ledger.js
+          node ledger.js --version
+      - name: Test Ledger (DocTest)
+        run: |
+          cd build
```
afh, line 182: Could running ledger.js to print the version be considered testing ledger?
```suggestion
      - name: Test Ledger (DocTest)
        run: |
          cd build
          node ledger.js --version
```

src/global.cc
gudzpoz: Now thinking about it, it may not be necessary any more.

Originally, I intended to make it easier for people to use the WASM(+JS) build as a "library" when they will usually want to call the main function multiple times. Calling `exit` had seemed to shut down the whole runtime and the user might need to reload the runtime and FS settings all over again.

However, reading more Emscripten documentation reveals that they have a dedicated flag ([`noExitRuntime`](https://emscripten.org/docs/api_reference/module.html#Module.noExitRuntime)) for this, so it is actually fine to use `exit`. Moreover, it seems Ledger does a [`quick_close`](https://github.com/ledger/ledger/blob/0c323a569f2774431fcef2e5fe4b5c425ab572b2/src/main.cc#L226) and I doubt calling `main` repeatedly might leak memory.

Overall, I think I will undo this change and maybe add some warnings on memory leaks in the to-be-draft documentation on this.  What do you think?

CMakeLists.txt
gudzpoz: I was compiling on a machine with ICU 75.1, which requires at least C++ 17.

https://github.com/ledger/ledger/blob/0c323a569f2774431fcef2e5fe4b5c425ab572b2/CMakeLists.txt#L167-L168

So while `check_cxx_source_runs` for `boost/regex/icu.hpp` (which includes headers from ICU) passes when C++ version is not specified, if Ledger (`-std=c++11`) includes `boost/regex/icu.hpp`, the build fails.

https://github.com/ledger/ledger/blob/0c323a569f2774431fcef2e5fe4b5c425ab572b2/CMakeLists.txt#L24

It is not directly related to Emscripten support. Should I revert it?

afh: Thanks for the added context, @gudzpoz, very helpful.

Now I see… before line 166 there is `cmake_push_check_state()`, so setting the flag here is for only for the `check_cxx_source_runs` "context".

Reading the doucmentation for [`check_cxx_source_runs`](https://cmake.org/cmake/help/latest/module/CheckCXXSourceRuns.html) it offers a `CMAKE_REQUIRED_FLAGS` parameter that is 
> String of additional flags to pass to the compiler

Maybe that's helpful?

src/global.cc
afh: This will be very interesting as the ledger python module does not work well with long running sessions as reading new data in the same runtime leads to assertion failure, because previously read transactions are not cleared, see #2320 and #514.

I'd like to get a better understanding of ledger in JS/WASM to form an opinion on this 🙂

src/report.cc
gudzpoz: No, this does not seem relevant anymore.

(A bit of context: I wrongly commented out the `check_function_exists` checks on `HAVE_ISATTY` `HAVE_IOCTL` when I first encountered the `ioctl_tiocgwinsz` bug mentioned above, which somehow got me to believe Emscripten does not provide an `isatty` implementation. And I made the above changes to pass https://github.com/ledger/ledger/blob/master/test/baseline/opt-force-color.test .)

src/global.cc
gudzpoz: Basically, running ledger in JS/WASM can be separated into two stages:
1. Stage 1: Running JS from `ledger.js`, which will load `ledger.wasm` and initialize a WASM runtime for it, analogous to ledger process creation (`preloading` phase in [emscripten-runtime-environment.html#execution-lifecycle](https://emscripten.org/docs/porting/emscripten-runtime-environment.html#execution-lifecycle))
2. Stage 2: Actually running the WASM binary with the runtime, analogous to jumping to the entry point of a loaded process (calling `run()` in [emscripten-runtime-environment.html#execution-lifecycle](https://emscripten.org/docs/porting/emscripten-runtime-environment.html#execution-lifecycle))

One can repeatedly call `run()`/the exported main function (Stage 2) and any static variable will not be reset, similar to:

```c
// js side
void main() {
  while (true) original_main(); // repeatedly call run()
}
// wasm side
#define main original_main
void main () { // ...
```

One may run Stage 1 multiple times when the [`MODULARIZE`](https://emscripten.org/docs/tools_reference/settings_reference.html#modularize)` flag is set. If not, the runtime will initialize only once and might result in issues similar to python ones.

.github/workflows/cmake.yml
```diff
@@ -73,3 +73,125 @@ jobs:
...
+          emmake make -j4
+      - name: Fix Emscripten build
+        run: |
+          cd build
+          # https://github.com/emscripten-core/emscripten/issues/22264
+          sed 's/if(!stream.tty)/if(!stream.tty||!stream.tty.ops)/g' ledger.js > ledger.js.ioctl_fix
+          sed 's/,tty:true,/,tty:false,/g' ledger.js.ioctl_fix > ledger.js.fs_fix
+          # Enable testing from python scripts.
+          echo '#!'"$(which node)" > ledger.js
+          cat ledger.js.fs_fix >> ledger.js
```
afh, line 177: It should be possible to do this without temporary files:
```suggestion
          sed -i '' \
          -e "s/^var/$(which node)\n&/" \
          -e 's/if(!stream.tty)/if(!stream.tty||!stream.tty.ops)/g' \
          -e 's/,tty:true,/,tty:false,/g' \
          ledger.js
```

afh (commented):
Please find below a few minor thoughts on this PR, @gudzpoz.

I'm working on getting ledger built with emscripten in [nixpkgs](http://github.com/NixOS/nixpkgs/pulls) and I've been able to compile everything (somewhat) successfully.
The issue I run into is:

```sh
% node ./result/bin/ledger.js -f $HOME/Developer/ledger/test/input/demo.ledger reg
Error: Cannot read journal file /Users/afh/Developer/ledger/test/input/demo.ledger"
% ls -lsa $HOME/Developer/ledger/test/input/demo.ledger
8 -rw-r--r--  1 afh  staff  1769 26 Feb  2023 /Users/afh/Developer/ledger/test/input/demo.ledger
```

Are programs compiled with emscripten somehow restricted in filesystem access?

.github/workflows/cmake.yml
```diff
@@ -73,3 +73,125 @@ jobs:
...
+        env:
+          EXT: ${{ matrix.version.emsdk-ext }}
+        run: |
+          cd build
+          export BOOST="$(realpath boost/dist)"
+          # -lnodefs.js -lnoderawfs.js flags are for testing purposes.
+          emcmake cmake \
+            -DBUILD_LIBRARY=ON \
+            -DPRECOMPILE_SYSTEM_HH=OFF \
+            -DCMAKE_CXX_FLAGS=' -lnodefs.js -lnoderawfs.js ' \
```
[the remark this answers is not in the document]
afh: Also note, that I needed `-DCMAKE_CXX_FLAGS='-fwasm-exceptions'` in order to be able to compile ledger with emscripten.

Could be that I did not build a dedicated boost with a minimal set of libraries for ledger, but rather used all of the libraries that would compile with emscripten, without further tweaking.

.github/workflows/cmake.yml
```diff
@@ -73,3 +73,125 @@ jobs:
...
+               toolset=emscripten link=static variant=release threading=single install
+      - name: Build GMP
+        run: |
+          cd build
+          export GMP_VER=6.3.0
+          wget https://ftp.gnu.org/gnu/gmp/gmp-${GMP_VER}.tar.xz
+          tar -Jxf gmp-${GMP_VER}.tar.xz
+          mv gmp-${GMP_VER} gmp
+          cd gmp
+          emconfigure ./configure --disable-assembly --host none --enable-cxx --prefix=$(realpath dist)
```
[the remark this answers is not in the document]
gudzpoz: Oh, I overlooked that warning. But it seems I cannot get GMP compiled without `--host none` on my machine, and I see people having similar problems (https://github.com/emscripten-core/emscripten/issues/18965, https://github.com/emscripten-core/emscripten/issues/22127).
Using `--host wasm32` (taken from the target triplet `wasm32-unknown-emscripten`) seems to remove the warning.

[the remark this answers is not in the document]
gudzpoz: I am running Arch Linux (x64).
I am not too familiar with BSD, but by "port" are you referring to FreeBSD ports? So I assume GMP (emscripten) compiles fine on BSD systems without the `--host none/wasm32` flag?
(On Linux (or Arch Linux at least), without `--host none/wasm32`, `configure` tries to determine the assembler according to the assembly format, which is generated by LLVM and is unknown to `configure`, causing the configuration step to fail.)

.github/workflows/cmake.yml
```diff
@@ -73,3 +73,125 @@ jobs:
...
+        run: |
+          cd build
+          export BOOST="$(realpath boost/dist)"
+          # -lnodefs.js -lnoderawfs.js flags are for testing purposes.
+          emcmake cmake \
+            -DBUILD_LIBRARY=ON \
+            -DPRECOMPILE_SYSTEM_HH=OFF \
+            -DCMAKE_CXX_FLAGS=' -lnodefs.js -lnoderawfs.js ' \
+            -DCMAKE_INSTALL_PREFIX="$PWD"/dist \
+            -DBoost_ROOT="$BOOST" \
```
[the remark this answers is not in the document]
gudzpoz: It seems related to [CMP0074](https://cmake.org/cmake/help/latest/policy/CMP0074.html) and [CMP0144](https://cmake.org/cmake/help/latest/policy/CMP0144.html). On CMake 3.30, using `BOOST_ROOT` gives a warning on [CMP0144](https://cmake.org/cmake/help/latest/policy/CMP0144.html) but yields no error. So probably using `BOOST_ROOT` guarantees maximal compatibility, and I am changing it to `BOOST_ROOT`. Thanks!

.github/workflows/cmake.yml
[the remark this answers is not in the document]
gudzpoz: On CMake 3.30, using supplying `-DLINK_OPTIONS` gives a warning about the option not being used at all:

```
CMake Warning:
  Manually-specified variables were not used by the project:

    LINK_OPTIONS

```

If it is so, it can explain why you are having this `Cannot read journal file` error, because, when using the building steps from the worflow file, filesystem access on *nix systems relies on the two flags `-lnodefs.js -lnoderawfs.js`. And if the two flags are not supplied due to `LINK_OPTIONS` not being used, native filesystem access will not be linked into the `ledger.js/wasm` files.

I changed it to [`CMAKE_EXE_LINKER_FLAGS`](https://cmake.org/cmake/help/latest/variable/CMAKE_EXE_LINKER_FLAGS.html) instead. Can you please test if it works for you?

.github/workflows/cmake.yml
gudzpoz: Thanks! I modified it a bit to make it (hopefully) work with both BSD Sed and GNU Sed. Can you try it out?

.github/workflows/cmake.yml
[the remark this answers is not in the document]
afh: Apologies I left out the "nixpkgs" before "port", in other words I'd like to offer an emscripten variant of ledger within the [nixpkgs](http://github.com/NixOS/nixpkgs) eco-system.

Thanks for the info, so if on nixpkgs Linux x64 this compiles without the need to specify `--host wasm32` then I should be fine, might be worth a comment though…

jwiegley:
@afh Hi Alexis, do you think we're close to being ready to accept this PR?

afh:
Apologies for the late response, I'll have another look and give it a spin with the previously mentioned nixpkgs work

afh:
I've managed to get the emscripten [gmp, mpfr](https://github.com/NixOS/nixpkgs/pull/329163) and [boost](https://github.com/NixOS/nixpkgs/pull/329178) building on nixpkgs and am now ready to give the changes proposed in this PR a spin.

@gudzpoz thanks so much for your effort and work on this, very much appreciated 💯 
Is this PR ready for review, @gudzpoz, or do you have more work lined up?

gudzpoz:
@afh Thank you! I think this PR should be ready for review now.

afh:
Does the [release of WASM 3.0](https://webassembly.org/news/2025-09-17-wasm-3.0/) have any implications for this PR?

gudzpoz:
> Does the [release of WASM 3.0](https://webassembly.org/news/2025-09-17-wasm-3.0/) have any implications for this PR?

There is one that has been addressed in the recent commits:

- 64-bit address space: b2 now seems to default to `address-model=64` with emscripten, so I added `address-model=32` when compiling boost (because Emscripten itself uses wasm32 by default).

On exception handling, WASM 3.0 uses a new `try_table` instruction for that, while Emscripten currently (`4.0.14`) seems to use the old `try` instruction. But I don't think we are affected: we only pass the `-fwasm-exceptions` flag, which I believe Emscripten will continue to support even if they decide to switch to `try_table` later:

> https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#400---011425
>
> -sWASM_LEAGCY_EXCEPTIONS option is added. (#23365) If true, it will emit instructions for the legacy Wasm exception handling proposal, and if false, the new standardized exception handling proposal. This option defaults to true, given that major web browsers do not support the new proposal by default yet. This option replaces the existing -sWASM_EXNREF, whose meaning was the opposite.

afh:
Thanks for adding the helpful context regarding WASM 3.0, @gudzpoz.

I've made good progress in getting the changes in this PR built using nixpkgs. The current roadblock I'm hitting is:

```
> wasm-ld: error: greg_month.o: section too large
```

The project ist configured as follows:
```
          emcmake cmake -S . -B build \
            -DBUILD_LIBRARY=OFF \
            -DCMAKE_CXX_FLAGS="-flto -DBOOST_DATE_TIME_NO_LOCALE" \
            -DCMAKE_EXE_LINKER_FLAGS="-Wl,--gc-sections -lnodefs.js -lnoderawfs.js -sSTACK_SIZE=1048576"
```

Here's the complete build log, hopefully that's helpful and not too riddled with nixpkgs specifics:
[nixpkgs-emscripten-ledger.log](https://github.com/user-attachments/files/22490890/nixpkgs-emscripten-ledger.log)

gudzpoz:
> ```
> > wasm-ld: error: greg_month.o: section too large
> ```

So this seems to come from Boost date_time. But I can't reproduce it (by building boost with `emscriptenPackages.boost` from https://github.com/afh/nixpkgs/tree/init-emscripten-boost and then using the built libraries to build ledger manually). Can I have a look at your nix configs?

afh:
I've not published my nix configs, but I'm happy to share here what might be helpful. Please tell me which aspects could be interesting.

For easier testing I've combined all nixpkgs related work regarding ledger emscripten support into https://github.com/afh/nixpkgs/tree/init-emscripten-ledger

I'm currently building from the https://github.com/afh/nixpkgs/tree/init-emscripten-ledger branch (and running into the aforementioned issue) on Darwin using nixpkgs using the following command:
```
nix build --option sandbox true -L .#emscriptenPackages.ledger
```

afh:
Please also note that the mix of `NIX_CFLAGS`, `NIX_LDFLAGS` and `CMAKE_CXX_FLAGS`, `DCMAKE_EXE_LINKER_FLAGS` stems from trying to find any way to address the `section too large` compile error.

afh:
Would the output of `nix config show` be helpful for you, @gudzpoz?

afh:
Building on `aarch64-linux` linking emscripten ledger fails with duplicate symbols instead of the "section too large" error:

```
       last 25 log lines:
       >
       > wasm-ld: error: duplicate symbol: __gmpz_add_ui
       > >>> defined in /nix/store/6wpwcmfaxbrpxppn4psv8b7pb95zhnk5-emscripten-gmp-with-cxx-6.3.0/lib/libgmp.a(lt3-add_u
i.o)
       > >>> defined in /nix/store/g7ccgsdy6mx5bi88b5fb9gdc3xcbp6qm-emscripten-mpfr-4.2.2/lib/libmpfr.so
       >
       > wasm-ld: error: duplicate symbol: __gmpn_add_1
       > >>> defined in /nix/store/6wpwcmfaxbrpxppn4psv8b7pb95zhnk5-emscripten-gmp-with-cxx-6.3.0/lib/libgmp.a(add_1.o)
       > >>> defined in /nix/store/g7ccgsdy6mx5bi88b5fb9gdc3xcbp6qm-emscripten-mpfr-4.2.2/lib/libmpfr.so
       >
       > wasm-ld: error: duplicate symbol: __gmpz_sub
       > >>> defined in /nix/store/6wpwcmfaxbrpxppn4psv8b7pb95zhnk5-emscripten-gmp-with-cxx-6.3.0/lib/libgmp.a(lt47-sub.
o)
       > >>> defined in /nix/store/g7ccgsdy6mx5bi88b5fb9gdc3xcbp6qm-emscripten-mpfr-4.2.2/lib/libmpfr.so
       >
       > wasm-ld: error: duplicate symbol: __gmpz_sub_ui
       > >>> defined in /nix/store/6wpwcmfaxbrpxppn4psv8b7pb95zhnk5-emscripten-gmp-with-cxx-6.3.0/lib/libgmp.a(lt48-sub_
ui.o)
       > >>> defined in /nix/store/g7ccgsdy6mx5bi88b5fb9gdc3xcbp6qm-emscripten-mpfr-4.2.2/lib/libmpfr.so
       >
```

afh:
I've been able to resolve the linux build errors and running `node ledger.js --version` works.

Trying to run report fails with:

```
node ledger.js -f ~src/ledger/test/input/drewr3.dat reg
~/Users/afh~/Developer/nixpkgs/ledger.js:4390
          if (stream.tty.ops.ioctl_tiocgwinsz) {
                             ^

TypeError: Cannot read properties of undefined (reading 'ioctl_tiocgwinsz')
    at ___syscall_ioctl (/Users/afh/Developer/nixpkgs/ledger.js:4390:30)
    at wasm://wasm/00f0c9b2:wasm-function[11036]:0x1b4b25
    at wasm://wasm/00f0c9b2:wasm-function[11020]:0x1b2c13
    at wasm://wasm/00f0c9b2:wasm-function[24893]:0x35d767
    at wasm://wasm/00f0c9b2:wasm-function[24291]:0x34c3a8
    at wasm://wasm/00f0c9b2:wasm-function[24278]:0x34b137
    at /Users/afh/Developer/nixpkgs/ledger.js:645:12
    at callMain (/Users/afh/Developer/nixpkgs/ledger.js:5902:15)
    at doRun (/Users/afh/Developer/nixpkgs/ledger.js:5955:24)
    at run (/Users/afh/Developer/nixpkgs/ledger.js:5968:5)

Node.js v22.19.0
```

Any hints on how to resolve this and how to get ledger.js running in the browser would be greatly appreciated 🙂

gudzpoz:
> I've been able to resolve the linux build errors and running `node ledger.js --version` works.

:clap: That's good news since I was really clueless :-P

> ```
> TypeError: Cannot read properties of undefined (reading 'ioctl_tiocgwinsz')
> ```

This is probably because Emscripten has yet to fix their noderawfs.js/ioctl_tiocgwinsz bug (https://github.com/emscripten-core/emscripten/issues/22264). Currently, if we are to use `noderawfs.js`, we have to manually patch the output for this: 

https://github.com/gudzpoz/ledger/blob/f078f9a1d775055badcc9a13b8fe26b3491529e7/.github/workflows/cmake.yml#L169-L174

Getting ledger to run in the browser requires a bit more work and can depend on the exact compiler flags used, because we need to configure its stdin/stdout/filesystem ourselves. Here is a possible wrapper: https://github.com/gudzpoz/ledger.js/blob/main/src/lib/ledger.ts , and the `LDFLAGS` used to compile the browser version: https://github.com/gudzpoz/ledger.js/blob/main/Makefile

And here is a demo: https://ledger.iroiro.party/

afh:
Thanks for the helpful feedback, @gudzpoz, that was it! I've updated my [afh/nixpkgs@init-emscripten-ledger](https://github.com/afh/nixpkgs/tree/init-emscripten-ledger) branch accordingly and am now able to build and run ledger.js 🎉

Thanks for sharing your insight on running ledger in the browser and the demo, I'll need to have a closer look.

Now that I've understood more about building ledger using emscripten I feel more comfortable to review this PR. Please allow for a bit more time so the changes in this PR get the time and attention they deserve 🙂 

ℹ️ For the time being I gave up on compiling ledger with emscripten on darwin…

bnjmnrsh:
Hi all - Ive been looking for double entry book keeping system for TS/JS and ran across this effort, which looks really promising. Has this work been superseded, or is there some other significant blocker?

jwiegley:
I too am curious to know where this work is at now and what remains before we can get it into the master branch.

jwiegley:
Closing for now, since this work seems to be still experimental. When you are ready for inclusion into `master`, please reopen!

Issue #2365, opened 2024-07-17
`roundto(int)` and `in_place_roundto(int)` are not exposed to python


maxnikulin:
`roundto(int)` and `in_place_roundto(int)` methods are unavailable for `Amount` and `Balance` python classes. Perhaps it happened just because these functions were added later than ceiling and floor.

Notice that in C++ code for `amount_t::roundto(int)` argument is ignored. It may be reasonable to fix #2329 and #2362 at first otherwise behavior might be a bit bizarre.

PR #2369, opened 2024-08-07
Add Python bindings for roundto

maxnikulin:
Expose `roundto` and `in_place_roundto` to `Amount`, `Balance`, and `Value` Python classes.  They may be handy for debugging. The methods were added after introducing of bindings and likely it is a reason why they were missed.

Closes: #2365

### DIFF
--- a/src/py_amount.cc
+++ b/src/py_amount.cc
@@ -207,6 +207,10 @@ internal precision."))
     .def("in_place_round", &amount_t::in_place_round,
          return_internal_reference<>())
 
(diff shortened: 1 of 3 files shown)


maxnikulin:
It is better to add some unit tests, but I have not realized what is a proper place for them. I have noticed [`convert.py`](../blob/master/test/convert.py), but it seems it slipped through a crack during transition from autotools to cmake.

jwiegley (approved):

PR #2371, opened 2024-08-08
disable checks for synthetic xacts, fixes  #2348

### LINKED ISSUE #2348: Segfault with --strict xact
The following file:
```
2024/04/24 Foobar
    Expenses:Fournitures                                              25.00 EUR
    ; Tag: Hello
    Assets:Cash
```

Provokes a segfault with ledger (3.3.2-20230330) when called like that:
```
cat file | ledger --init-file /dev/null -f - --strict xact 2024/05/12 Foobar
```

Removing the Tag or calling ledger with `-f file` doesn’t segfault, but ledger-mode uses the syntax above so it happens quite regularly.

taviso:
There were two use-after-free bugs when using `xact`, one was only visible when using `--strict`

This fixes that issue by disabling checking when the `parse_context_stack` is empty, and moving the regex_match parameters into scope.

### DIFF
--- a/src/draft.cc
+++ b/src/draft.cc
@@ -242,6 +242,9 @@ xact_t * draft_t::insert(journal_t& journal)
   xact_t *              matching = NULL;
   unique_ptr<xact_t> added(new xact_t);
 
+  // There is no need to check drafts for errors, because we generated them.
+  journal.checking_style = journal_t::CHECK_PERMISSIVE;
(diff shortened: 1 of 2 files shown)


immae:
Tested successfully with the workflow that used to fail (see #2348 )

jwiegley (approved):

PR #2373, opened 2024-08-09
Avoid string in amount_t::in_place_truncate

maxnikulin:
Do not serialize to string just to get rounded value. `in_place_roundto` should be reliable enough now.

### DIFF
--- a/src/amount.cc
+++ b/src/amount.cc
@@ -622,40 +622,15 @@ void amount_t::in_place_round()
 
 void amount_t::in_place_truncate()
 {
-#if 1
   if (! quantity)
     throw_(amount_error, _("Cannot truncate an uninitialized amount"));
-
-  _dup();
 
   DEBUG("amount.truncate",
         "Truncating " << *this << " to precision " << display_precision());
 
-  std::ostringstream out;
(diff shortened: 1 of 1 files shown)


maxnikulin:
Feel free to reject this optimization attempt if you prefer to not change code that is not broken

jwiegley:
What does `#if 2` mean? Isn't that the same as having no `if` at all?

maxnikulin:
I do not mind to just drop current variant of code completely, anyway it can be easily found in git history. However I noticed that simple and inefficient code is preserved for some reason. That is why I decided to use the same approach and to hide current code from compiler by `#if` and `#else`.

jwiegley:
Sure, in this case I'd just drop the version that will no longer be used. Leaving it there does not seem to be very educational, since I have a hard time convincing myself that it represents the "simpler" version of the code in the other branch...

jwiegley:
And if you mean that the simpler version is the final one, then even more so we should delete the older, complex code.

maxnikulin:
I have dropped older variants of `in_place_truncate`

jwiegley (approved):

PR #2376, opened 2024-08-27
WIP: Make virtual-balanced transactions balance separately.

### LINKED ISSUE #2105: Unbalanced Bracketed virtual postings do not throw a warning nor error
Given journal
```
2022/04/07  test
    [Fake1] $-10000
```

Run a bal report and get no results.  No warning that the virtual posting is unbalanced.
```
PS C:\Users\user\ledger> ./ledger -f test.journal bal
PS C:\Users\user\ledger> 
```

I'm very new to ledger so it is possible I do not understand how this works.  But in my main journal I have more complicated entries like the below, where I had an unbalanced virtual posting.  It's off by .02, but I totally missed it because bal reports never warned me I had an out of balance situation.
```
4/1         Paycheck
    Assets:Fairwinds                    $1198.07
    Income:Paycheck
    [Equity:Budget]                     $-1198.07
    [Budget:Fairwinds:General Funds]    $333.06
    [Budget:Fairwinds:Mortgage Hold]    $865.03
```

I'm running the [FullofQuarks Binary](https://github.com/FullofQuarks/Windows-Ledger-Binaries)
```
PS C:\Users\user\ledger> ./ledger --version
Ledger 3.2.1-20200518, the command-line accounting tool
```

ChickenProp:
This is intended to close #2105: in a given transaction, virtual postings (with `[]`) should balance to 0 and real postings should balance to 0, but until now it's just been required that the complete set "virtual postings (with `[]`) and real postings" should balance to 0. Minimally:

```
2024/08/11 Example
    Income                                     $1.00
    [Expenses]                                $-1.00
```

ought to error but until now has been accepted.

It's not complete yet. There are some design decisions and I'd appreciate some opinions on them. Also, I don't really know C++. Right now a lot of the code here is just "copy the existing code and do it again with some variables and some flags changed", and if there's some way to avoid that it would be nice to know.

### DIFF
--- a/src/xact.cc
+++ b/src/xact.cc
@@ -162,11 +162,16 @@ bool xact_base_t::finalize()
   // price of unpriced commodities.
 
   value_t  balance;
-  post_t * null_post = NULL;
+  value_t  balance_real;
+  value_t  balance_virtual;
+  post_t * null_post_real = NULL;
+  post_t * null_post_virtual = NULL;
 
   foreach (post_t * post, posts) {
     if (! post->must_balance())
       continue;
+
+    bool is_virtual = post->has_flags(POST_VIRTUAL);
 
     amount_t& p(post->cost ? *post->cost : post->amount);
     if (! p.is_null()) {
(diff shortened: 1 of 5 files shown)


src/xact.cc
```diff
@@ -211,13 +224,22 @@ bool xact_base_t::finalize()
...
-    null_post->_state = (*posts.begin())->_state;
-    add_post(null_post);
+  if (journal && journal->bucket && posts.size() == 1) {
+    if (! balance_real.is_null()) {
+      null_post_real = new post_t(journal->bucket, ITEM_INFERRED);
+      null_post_real->_state = (*posts.begin())->_state;
+      add_post(null_post_real);
+    }
+    // TODO: does this need to set POST_VIRTUAL | POST_MUST_BALANCE ?
+    else if (! balance_virtual.is_null()) {
```
ChickenProp, line 234: Apparently there's some way to say "if an xact has only one posting, here's a default account to balance against".

If an xact has exactly one real and one balanced-virtual posting, right now I'm taking that to be an error. We could instead choose to balance both against the default account. But currently,

```
2024/08/27 Example
    Income        $1.00
    (Expenses)    $-1.00
```

doesn't balance against the default, so I wouldn't expect it to with `[]` either.

src/xact.cc
```diff
@@ -211,13 +224,22 @@ bool xact_base_t::finalize()
...
+      add_post(null_post_virtual);
+    }
   }
 
-  if (! null_post && balance.is_balance() &&
+  // TODO: How do we want handle virtual postings here?
+  if (! null_post_real && ! null_post_virtual && balance.is_balance() &&
       balance.as_balance().amounts.size() == 2) {
     // When an xact involves two different commodities (regardless of how
     // many posts there are) determine the conversion ratio by dividing the
```
ChickenProp, line 245: When there are two commodities, we calculate the conversion rate between them.

If there are two real and two virtual, should we do the same? Regardless of whether that's two, three, or four in total?

src/xact.cc
```diff
@@ -422,6 +487,7 @@ bool xact_base_t::finalize()
   return true;
 }
 
+// TODO: fix for virtual post balancing.
 bool xact_base_t::verify()
 {
```
ChickenProp, line 492: When I do try to fix this, how should I test it? My current guess is this code path gets called when there are auto xacts?

jwiegley:
Please submit a pull request when this work is ready to be reviewed.

PR #2379, opened 2024-09-10
Filter posts for computing balance assignment

### LINKED ISSUE #2273: What does the expr argument do in account_t::amount?

### LINKED ISSUE #2378: Build fails with boost 1.86
With Boost 1.86.0-2, I get the following error during build:
```
In file included from src/exprbase.h:56,
                 from src/draft.h:44,
                 from src/stats.cc:34
src/utils.h: In function ‘ledger::string ledger::sha1sum(const string&)’:
src/utils.h:602:18: error: cannot convert ‘unsigned int [5]’ to ‘unsigned char (&)[20]’
  602 |   sha.get_digest(message_digest);
      |                  ^~~~~~~~~~~~~~
      |                  |
      |                  unsigned int [5]
In file included from src/utils.h:46:
/usr/include/boost/uuid/detail/sha1.hpp:179:43: note:   initializing argument 1 of ‘void boost::uuids::detail::sha1::get_digest(unsigned char (&)[20])’
  179 | inline void sha1::get_digest(digest_type& digest)
      |                              ~~~~~~~~~~~~~^~~~~~
```

dbear496:
This is an improvement from PR #2274. Without this PR, it is very difficult to use balance assignment/assertion when the journal is split between two or more files. (Read the conversation from #2274 for examples.) This PR provides a solution to this problem whereby balance assignments/assertions may be tagged with `balance_filter` and given an expression to use for filtering out previous posts that should not be included in the total for the purpose of the balance assignment. I envision `balance_filter: balance_post.date >= post.date` being very useful -- this causes postings that are dated later than the balance assertion to not be included in the calculation for the balance assertion.

My previous attempt at solving this problem (#2274) hard-coded a date comparison into the balance assertion calculation, and the PR was denied because such hard-coding is undesirable. This PR improves on that solution by allowing the user to filter by any arbitrary expression.

This PR also sprinkles in some other fixes:
- Fixes an issue left over from #1875 where real posts would not affect virtual balance assertions within the same transaction.
- Fixes #2273
- Fixes #2378 (because otherwise I couldn't build the project to submit this PR)
- Adds regress tests for balance transactions with mixed-up dates. (So the "solution" proposed in #2274 would officially break stuff.)
- Adds regress tests for the `balance_filter` tag.

Note: The `balance_filter` tag still needs user documentation.


dbear496:
It looks like my quick fixes for #2378 are breaking CI which has not yet been updated to Boost 1.86, so I'll have to wait until someone else fixes that issue.

dbear496:
This PR has been sitting around for over a year now. Any chance someone can review it?

jwiegley:
## Documentation Note

As mentioned in the PR description, the `balance_filter` tag still needs user documentation. This should include:

1. **Basic syntax**: How to use the tag (`; balance_filter: <expression>`)

2. **Common use cases**:
   - `balance_filter: balance_post.date >= post.date` - Filter out posts dated after the assertion
   - `balance_filter: balance_post.actual_date >= (post.aux_date || post.actual_date)` - Filter by clearing date

3. **Available variables**:
   - `balance_post` - Reference to the post being checked in the balance calculation
   - `post` - The post containing the balance assertion
   - `amount` - The post amount (used in the filter expression evaluation)

4. **Limitations**:
   - Cannot be propagated from automatic transactions (as noted in PR description)

Consider adding this to `doc/ledger3.texi` in a follow-up PR or as part of this one before merge.

jwiegley:
I've auto-formatted the source code to common standard, which will conflict with your PR. You can follow these steps to auto-format your work:
```sh
wget https://data.newartisans.com/pub/reformat.patch
patch -p1 < reformat.patch
<regen your cmake files, however you do that>
make reformat
make lint
```

dbear496:
@jwiegley, I synchronized with the current master branch and ran the reformatter to fix the style issues. Per your recommendation, I also added a test for a malformed balance filter expression, and I wrote some user documentation for the feature.

jwiegley:
@dbear496 I'm thinking about this PR more and more, and while I do very much appreciate the work you've put into it, I feel that the increase in complexity is not worth the limited size of the audience that this will apply to. Especially since you need to either repeat the balance filter for every posting/transaction where you want it to apply, or you have to use an automated transaction somehow to apply it. It just seems to be very niche, yet changes a fair number of files for such a feature.

I can appreciate it in theory, but I'm not sure I can accept it in practice.

jwiegley:
Closing the PR, but open to rebuttal if you believe strongly it should go in.

dbear496:
Many of the changed files are only loosely related to the feature addition.

The core of the feature is implemented solely in textual.cc. In that file, the code for computing balance assertions is moved so it occurs after parsing the optional (same-line) note. Then only about 30 lines (including comments) are added to facilitate the filtering.

Changes to the other source files (account.h and account.cc) are solely to fix bug #2273, which needs to be fixed for the feature to work properly. Changes to non-source files are adding regression tests and documentation.

So the actual feature implementation really only amounts to ~30 lines in textual.cc.

I understand that the use-case for this feature is rather niche; however, it solves a problem for which there is no other viable workaround. For my workflow, I must have this feature, so unless/until it is merged, I have to run a patched version of ledger. Since I have clarified that there is actually minimal added complexity to implement this feature, I hope you will reconsider, but if not, then I will respect your decision.

P.S. Since you only call into question the feature itself, I will work on splitting the bug fixes into a separate PR.

P.P.S. You are right that this feature has some rough edges, and this mostly stems from the computation of balance assignments/assertions during parsing when not all metadata is available. This could all be fixed by deferring the processing of balance assertions until all relevant information is parsed and attached to the posting, but such deferral would of course add complexity, which is why I chose to overlook these rough edges in the first place. With that said, would you prefer this feature if I resolved the rough edges by deferring balance assertion/assignment calculations at the cost of some extra complexity?

jwiegley:
I'm happy to review the other, more basic PRs.

For changing the way that balance assignments are computed, maybe instead of a per-exact expression, this deserves a command-line option like `--balance-expr` or some such, so that it applies to an entire set of ledger files?

There are many edge cases that arise when you process entries "in file order"; this is just one of them. Being fully date-ordered has its downsides, which is why I went with file order, the most obviously simple and deterministic option. But maybe there should be a command-line option in session_t that changes this bias globally.

dbear496:
> For changing the way that balance assignments are computed, maybe instead of a per-exact expression, this deserves a command-line option like `--balance-expr` or some such, so that it applies to an entire set of ledger files?

I think this is a very bad idea. So far, all command line options affect only how reports are generated -- not the semantics of the journal itself. Consider if you wanted to send a journal to a friend: in addition to sending them the file, you would additionally have to communicate what `--balance-expr` to use. Thus I strongly believe the balance filter expression should be specified within the journal itself.

Since the ability to specify the balance filter globally is important, I will look into fixing the propagation of balance filters from automatic transactions.

jwiegley:
The other possibility, for a "journal-specific balance expression logic" would be to introduce a new directive at file level.

PR #2389, opened 2024-10-17
Remove unnecessary macros and their usage

pslldq:
Macros are powerful, but they make it harder to check that the code compiles in every configuration. Therefore remove some fixed value macros and also macros which remove code parts that are already eliminated by the compiler (e.g. assert)


jwiegley (approved):

jwiegley:
Hmm... leads to some build failures.

pslldq:
Ah, missed that `current_memory_size` only existed with `VERIFY_ON`. Fixed and rebased the PR.

PR #2398, opened 2024-12-11
Add yday shortcut for yesterday

aeifn:
I often use 'yesterday' in my queries, but it is tedious to type 'yesterday' every time.


jwiegley (approved):

jwiegley:
We just need an addition to the Texinfo manual, and a test!

Issue #2388, opened 2024-10-17
Filter by commodity alias


igbanam:
Assuming I have my commodity defined thus

```
commodity $
   note American Dollars
   format $1,000.00
   nomarket
   alias USD
   default
```

**How do I create a report filtered via the alias?**

Note that using the actual `$` works

```
ledger bal -l "commodity == '$'"
```

But if `$` was some obscure symbol, like `₿`, which is not readily available on the keyboard, it would be nice to be able to do something like

```
ledger bal -l "commodity == alias('BTC')"
```

jwiegley:
This is a great suggestion.

taviso:
A hacky solution would be `ledger bal -l "{0 BTC} == strip(amount) * 0"` 🤣

igbanam:
This is a fantastic idea. Thank you!

acransac:
@igbanam I took the opportunity to contribute back to Ledger and implemented this in PR2405. I made this so that the `alias` function you gave as an example wouldn't be necessary but that the equality test would be aware of any alias the commodity has.

PR #2405, opened 2025-01-28
Issue 2388: Use Commodity Aliases In Reports Filters 

acransac:
# Summary
In #2388 @igbanam proposed that one should be able to use aliases when filtering reports by commodities. This is implemented with a new commodity value type in value expressions so that aliases can be resolved in comparison operations.

# Implementation Details
I added a commodity `value_t` with a specialization for comparison operators `==,<,<=,>,>=`. In these contexts, when the second operand is a string it is used to look for a matching commodity in the pool. If found, the operands are equal, otherwise (for `<,<=,>,>=`) a regular string comparison is computed using the commodity's primary symbol. For arithmetic operations `+,-,*,/`, the commodity `value_t` behaves like a string for backward compatibility. For other `value_t` methods, a sensible implementation is provided. 

Interoperability between commodity `value_t` and the regular `commodity_t` is provided keeping into consideration that the commodity `value_t` behaves like a string in contexts that are not defined for `commodity_t`. This means that while `commodity_t` is only equality-comparable, commodity `value_t` and `commodity_t` form a totally ordered and addable group (adding string comparison and concatenation to the original equality-comparison).

When filtering reports using `commodity` in a value expression, for every posting this commodity `value_t` stripped of any annotation is used instead of a string previously.

The Python interface is extended accordingly and unit, baseline and Python tests are added.

# Backward Compatibility
To the best of my understanding I made this a minor backward-compatible change so that users don't have to modify their usual `limit` queries when these involve `commodity`. In practice, it means that apart from matching with aliases for equality-comparison, `commodity` in a value expression still behaves like a string.

# Performance
For comparison operators `==,<,<=,>,>=` involving a commodity `value_t` on either side of the operation, there is an additional lookup into the commodity pool. This is repeated for every posting in a report with `limit`. I made the assumption that on an average journal input the number of commodities is small relatively to the number of postings so that the extra lookup penalty (whose asymptotic computational cost is logarithmic in the number of commodities) is negligible.


acransac:
Looking at the test failure

afh (approved):
This is impressive work, @acransac, thanks for putting in all the effort. I've tested the changes locally and it works as expected as far as I can tell from my understanding.

I'm in favor to merge this, yet would appreciated if another pair of :eyes: —maybe @jwiegley's—would at least glance over the implementation details.

acransac:
Thank you for the kind words and taking the time to review this @afh !

jwiegley:
I'm initially opposed to this PR. This feels like hacking something into `value_t` in order to achieve a desired behavior. Commodities aren't values, that would be like saying that elements from the basis set of a vector space could also be points in that space: it confounds domains.

I'd like to take a bit more time to think about the use case for which this is a proposed solution, to think if there's another way to do this without expanding the semantics of `value_t`.

jwiegley:
I think the right answer here is a higher-level comparison operations. Instead of making commodities into values so that `==` may be used to compare them, we can just introduce a new comparison for commodities that takes aliases into account:
```
ledger bal -l "same_commodity(commodity, 'BTC')"
```
This gives the desired functionality, maintains semantic integrity, and should result in far less code changes to accomplish.

I do very much appreciate the thoughtful work that went into this PR, I just want to be very, very cautious about changing the meaning of something so fundamental as value_t.

jwiegley (changes requested):
See my comments above.

afh:
Thank you for your valuable insight into this this, @jwiegley, very much appreciated 🙏

acransac:
Hi @jwiegley , thanks for the feedback. I must have misunderstood something in the codebase's design.

> Commodities aren't values

I assumed values were any potential operand in a value expression, regardless of the underlying type. After all, `commodity` is already a value expression variable (as a string to this point) and the documentation states that "value_t/Value, into which many things can be stuffed and then operated on. They can contain amounts, balances, dates, strings, etc". Could you specify how a commodity `value_t` violates this design?

As for the solution you propose, as a Ledger user I would find this counter-intuitive. Why could I write "-l commodity == $" but would have to use a different expression (that I would need to look up in the documentation) if comparing with an alias? I am likely to have set an alias so that I can use different names interchangeably and not remember which one is the main and which one is the alternative.

I understand you don't want to see a use case lead to a change that does not fit the code's design so my second point is just a user's feedback but I would appreciate you clarify the first item I mentioned.

jwiegley:
Well, let's consider this further, because you may well be able to change my thinking on this point.

Values exist within a vector space, which is to say that every value is an ordinal or set of ordinals associated with a "kind" chosen from a basis set.

The basis set is:
- integer
- commoditized amount
- commoditized balance
- date
- date and time
- string
- boolean

jwiegley:
Well, let's consider this further, because you may well be able to change my thinking on this point.

Values exist within a vector space, which is to say that every value is a set of ordinals associated with a "kind" for that set of ordinals chosen from a basis set.

And.... that is what I want to be able to say. But then I look at the definition for `value_t` and realize I've already made it deviate from this mathematical conception by allowing the following to also be considered "values":

- regular expression objects
- pointers to lexical scopes
- pointers to arbitrary values

So, OK, if we consider the "value" of a commodity to be a unique index that identifies that commodity for the duration of a session, and if commodities share this index with all of their aliases, then I do agree that commodities can be rightly considered values and can accept your pull request.

jwiegley (approved):
See comments above.

acransac:
Thank you @jwiegley for the review and accepting a more liberal interpretation of values than the original intent.

If you can spare an extra minute to explain further, when you say "values exist within a vector space", do you mean they are meant to be vectors in that space or elements of the underlying field (the fact that `value_t` implements the operations of a field keeps me swinging between the two understandings).

And when you mention "ordinals", "kinds" and a basis set, are you hinting at the types in `value_t::type_t` being the "kinds" forming a basis for all possible values where each kind has a possibly infinite but enumerable amount of elements?

It's an interesting design, I'd like to be able to connect the dots between the mathematical concepts and the implementation.

jwiegley:
@acransac Sure, although I think strings and time muddy the possibility of it being a vector space, and it may indeed be much closer to a field.

For example, `amount_t` values really are a vector space, where the basis set is the set of possible commodities, and the magnitude is the amount. Each commodity is effectively a dimension, and these amounts can be added, substracted, or multiplied by an integral scalar.

When you combine amounts, you get bundled of such vectors, which is a `balance_t`. It does dimension-wise addition and subtraction, and multiplies all elements by the scalar. Now that you ask me to explain it, I'm not sure what kind of construction this is.

Since values were originally intended to abstract over amounts, so that all amounts can be considered as single dimensional balances — and the two inter-mixed — I was thinking of values as vector spaces also, which I now realize is not accurate.

Then I added strings, which still have a concept of addition, but no subtraction or scalar multiplication. So they aren't a vector space, but do still fit within a `value_t`. This already completely breaks the semantics I had originally imagined.

Then other custom types went into, with "pointers to scopes" having no underlying mathematical model at all. Those are just unique identifiers of a specific scope. In this sense, then, commodities are exactly the same thing, being unique identifiers of an underlying commodity definition.

Given that `value_t` already allows for 'pointer to entity', then, and since your notion of commodities-as-value is really a "pointer to a commodity" that can be compared in terms of the underlying commodity, then it fits quite neatly within the exist semantics of `value_t` and this is why I accepted the pull request.

Sadly, I have no good semantics now for value_t other than "kitchen sync type" that plays variantly with overloaded operators like `==` and `+`. There are subsets of values that act as mathematical objects, but not the aggregate.

Thank you for revealing my confusion, as it forces me to rethink my understanding of these fundamental definitions within Ledger.

acransac:
Thank you so much for the thorough explanation and history of these types @jwiegley !

It makes sense how `value_t` started as a vector space type when dealing with amounts and balances that can be seen as linear combinations of unit commodities. Then strings stretched the construct so that it couldn't be describing a vector space anymore and `value_t` generalized to a sort of wrapper for everything that can be involved in a value expression for example. In this timeline, the commodity `value_t` belongs indeed to this third phase of enumerable kinds that can be involved in value expressions.

>When you combine amounts, you get bundled of such vectors, which is a balance_t. It does dimension-wise addition and subtraction, and multiplies all elements by the scalar. Now that you ask me to explain it, I'm not sure what kind of construction this is.

>Since values were originally intended to abstract over amounts, so that all amounts can be considered as single dimensional balances — and the two inter-mixed — I was thinking of values as vector spaces also, which I now realize is not accurate.

If I follow you correctly that an `amount_t` is a `balance_t` when only one commodity is involved and that we have a basis of unit commodities like for example (USD1, EUR1, ...) generating all the possible `balance_t` (and/or `amount_t`) then you are right that this is a vector space.

>Sadly, I have no good semantics now for value_t other than "kitchen sync type" that plays variantly with overloaded operators like == and +. There are subsets of values that act as mathematical objects, but not the aggregate.
Which in my opinion is very convenient for processing value expressions.

PR #2419, opened 2025-04-12
Initial implementation of “commodity swaps” over a base commodity

### LINKED ISSUE #2418: FR: Swapping assets over a common base commodity
Currently one is able to swap an asset price in a base commodity into some amount of commodity, with a resulting gain/loss that must be balanced. For example:
```
2025-04-11 Coinbase
    Assets:Coinbase                             -100 BTC {$100} @ $40
    Assets:Coinbase                        $4,000.00
    Expenses:Capital Loss                  $6,000.00
```

This feature request would allow for two commodities, both priced in the same base commodity, to be "swapped", with a resulting gain/loss reflecting the discrepancy in value. This would allow the following to balance:
```
2025-04-12 Coinbase
    Assets:Coinbase                             -100 BTC {$100} @ 20 ETH {$40}
    Assets:Coinbase                            2,000 ETH {$40}
    Expenses:Capital Loss                  $2,000.00
```

This sort of "commodity swap over a common base commodity price" is a very regular occurence on crypto exchanges. It makes it easy to track the two tax events involved:

1. The tax event at time of swap, due to any discrepancy in value between the commodities being swapped.

2. The tax event at time of final liquidation into the base commodity.

jwiegley:
Fixes #2418


simonmichael (approved):
An interesting, perhaps labour saving new feature of the Ledger file format, with excellent docs. 👍🏻

How much time does it save ? What would be the equivalent manual entry without this feature ?

The doc and test use explicit postings for all gain/loss, should there be a test/example of the automatic handling of gain/loss ?

The new kind of entry was not supported before I think, so there should be no breakage of existing journals.

How will users break or trip up on the new functionality ? Will it degrade gracefully, eg with sensible error messages ?

PR #2430, opened 2025-08-11
Fix build with Boost 1.89.0

cho-m:
In the upcoming Boost 1.89.0 release, Boost.System stub has been removed (https://github.com/boostorg/system/commit/7a495bb46d7ccd808e4be2a6589260839b0fd3a3) which causes CMake error:

```
CMake Error at /opt/homebrew/lib/cmake/Boost-1.89.0/BoostConfig.cmake:141 (find_package):
  Could not find a package configuration file provided by "boost_system"
  (requested version 1.89.0) with any of the following names:

    boost_systemConfig.cmake
    boost_system-config.cmake
```

I noticed this while testing 1.89.0.beta1 in Homebrew https://github.com/Homebrew/homebrew-core/pull/233031.

Since master currently requires a minimum of 1.72, it should be safe to remove `system` which was made into a stub back in 1.69[^1]. See `https://github.com/boostorg/system/issues/132#issuecomment-3146378680`

[^1]: https://www.boost.org/doc/libs/1_69_0/libs/system/doc/html/system.html#changes_in_boost_1_69

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,7 @@ endif()
 
 # Set BOOST_ROOT to help CMake to find the right Boost version
 find_package(Boost ${Required_Boost_Version}
-  REQUIRED date_time filesystem system iostreams regex unit_test_framework
+  REQUIRED date_time filesystem iostreams regex unit_test_framework
   ${BOOST_PYTHON} OPTIONAL_COMPONENTS nowide
   CONFIG)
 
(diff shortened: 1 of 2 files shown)


afh:
Thanks for this PR, @cho-m, and the additional context, let me have a closer look.

jwiegley:
@afh How does it look? Ready to merge?

afh:
@cho-m I took the liberty to add changes to this PR, in order to address the macOS build issues, hopefully you don't mind.

@jwiegley Yes, this is ready to merge, I wanted to investigate the build issues beforehand.

afh:
@cho-m also thanks for your contribution and for using ledger, both very much appreciated 👍

PR #2447, opened 2025-08-22
Fix xact command to respect last year directive

### LINKED ISSUE #2413: Last "year" directive is not respected by "xact"
It is inconvenient that `xact` does not use the current year or the year from last `year`. Instead some other value is used, I have not figured out which particular one.  For the following script I would expect 2025 in both cases.

```sh
#!/bin/sh
set -x
date '+%F'

ledger -f - xact 03-03 Unexpected 2.4 <<'LEDGER'
year 2021
year 2022
year 2023
year 2024
year 2025

2025-03-02 Unexpected
    Expenses:Food  $5.40
    Account
LEDGER
ledger -f - xact 03-03 Expected 2.5 <<'LEDGER'
2025-03-02 Expected
    Expenses:Food  $5.40
    Account
LEDGER
```
Actual result, notice 2024 that is neither from first `year` nor from the last one:
```
+ date +%F
2025-03-04
+ ledger -f - xact 03-03 Unexpected 2.4
2024/03/03 Unexpected
    Expenses:Food                              $2.40
    Account
+ ledger -f - xact 03-03 Expected 2.5
2025/03/03 Expected
    Expenses:Food                              $2.50
    Account
```

I had a hope that I can have abbreviated effective dates if I add `year` directives. Now I have to replace year in newly created transactions:
```ledger
year 2024
2024-03-03=03-07 * Shop  ; for =2024-03-07
    Expenses                                      $9.00
    Account

year 2025
2025-03-03=03-06 Shop    ; for =2025-03-06
    Expenses                                     $10.00
    Account
```

Update: Example and expected output are fixed. I have realized that shell input here markers were not quoted, so `$5` was expanded as 5th positional script argument to nothing. It caused missed `$` and fractional part in amounts generated by `xact` commands. It does not affect regression tests added in the pull request. 2025-09-02.

jwiegley:
## Summary

This PR fixes issue #2413 where the `xact` command was not respecting the last `year` directive when multiple year directives were present in a journal file.

## Problem

When a journal file contained multiple `year` directives like:
```ledger
year 2021
year 2022
year 2023
year 2024
year 2025
```

The `xact` command would use year 2024 instead of 2025 when parsing dates without explicit years (e.g., `03-03`).

## Solution

The fix involves several changes:

1. **Deferred date parsing in draft_t**: Instead of parsing dates immediately when processing xact command arguments, we now store the date string and parse it later when the journal context (with the correct epoch) is available.

2. **Date separator conversion**: Added conversion of date separators (hyphens and periods) to slashes for consistent parsing, since the date parser expects slash-separated dates.

3. **Epoch-aware date parsing**: Modified the date parsing logic in `times.cc` to use the epoch year (set by year directives) when parsing dates without explicit years, rather than always using the current system date.

4. **Proper year directive handling**: Fixed the handling of year directives to ensure they don't incorrectly use the apply_stack for plain "year" directives (which are permanent), while preserving the correct behavior for "apply year"/"end apply" patterns.

## Testing

- Added regression test `test/regress/2413.test` that verifies the fix
- Tested with the example from the issue report
- Verified that existing tests pass (except for unrelated pre-existing failures in 1038 tests related to "this month" period parsing)

## Test plan

- [x] Run the new regression test: `python3 test/RegressTests.py -l build/ledger -s . -j1 test/regress/2413.test`
- [x] Test the scenario from the issue manually
- [x] Run baseline tests to ensure no regressions

Fixes #2413

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/draft.cc
+++ b/src/draft.cc
@@ -106,7 +106,7 @@ void draft_t::parse_args(const value_t& args)
 
     if (check_for_date &&
         regex_match(arg, what, date_mask)) {
-      tmpl->date = parse_date(what[0]);
+      tmpl->date_string = what[0];
       check_for_date = false;
     }
     else if (check_for_date &&
@@ -144,7 +144,7 @@ void draft_t::parse_args(const value_t& args)
       else if (arg == "on") {
         if (++begin == end)
           throw std::runtime_error(_("Invalid xact command arguments"));
-        tmpl->date = parse_date((*begin).to_string());
+        tmpl->date_string = (*begin).to_string();
         check_for_date = false;
       }
       else if (arg == "code") {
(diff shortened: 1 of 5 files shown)


jwiegley:
The implementation is correct, but it seems a little overly involved to me. I need to read through what it's doing and refine this to a more minimal change.

PR #2456, opened 2025-08-31
Add gpgme.h include

afh:
which provides GPGME_VERSION_NUMBER

PR #2457, opened 2025-08-31
Bump required Python version to 3.10

afh:
With official [support for Python 3.9 ending around October 2025](https://www.python.org/downloads/) I think it's feasible to bump the minimum required version of Python to 3.10.


afh:
macos check fails due to pending fix [`b55db80` (#2456)](https://github.com/ledger/ledger/pull/2456/commits/b55db80de00bd022591863f19d4e38d1704694ac)

PR #2461, opened 2025-09-08
Avoid recursive include

### LINKED ISSUE #1662: Segfault in ledger::account_t::find_account
```
$ cat Q

!include Q*

$ cat Q1

2018-01-01 Test
    Assets                       1000.00 USD
    Test

$ ledger -f Q bal
zsh: segmentation fault  ledger -f Q bal

$ gdb ledger
...
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7997718 in ledger::account_t::find_account(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) ()
   from /usr/lib/ledger/libledger.so.3
```

afh:
While the current ledger on master does not segfault as initially described in #1662 when ledger encounters a recursive `include` statement, ledger still follows the include recursion for many iterations therefore skewing the result.

While the changes in this PR do address the issue, there might be a better or more thorough way to address this.
Also my understanding of the issue could've been wrong and therefore my efforts to address this could've been misguided.

What's your take on this, @jwiegley?

Fixes #1662

### DIFF
--- a/src/textual.cc
+++ b/src/textual.cc
@@ -798,6 +798,10 @@ void instance_t::include_directive(char * line)
           DEBUG("textual.include", "Skipping file with invalid UTF-8 name: " << base);
           continue;
         }
+        if (context.pathname == *iter) {
+          DEBUG("textual.include", "Avoiding recursive include of: " << *iter);
+          continue;
+        }
         if (glob.match(base)) {
           journal_t *  journal  = context.journal;
           account_t *  master   = top_account();


jwiegley:
I mean, this seems like a safe and simple way to tackle the problem to me!

jwiegley (approved):

afh:
I'm glad to hear that, John, thanks for taking a look 👍

PR #2477, opened 2025-10-17
rust: get all cargo tests passing

claytonrcarter:
Wondering if CI will pass... it does!

This includes a number updates such that all cargo tests are now passing. In some cases, the assertions/expectations were incorrect, in others the implementation was incorrect.

- changes `ProcessingStats` to record proccessing time in microseconds, because milliseconds wasn't granular enough
- disabled a memory benchmark because a) it wasn't working and b) it's not clear to me how to measure what we want to measure
- adds support for Commodities with a separator
- breaks, and then refixes, `align-amounts.test`

With this set of changes, I'm shifting my tinkering to look at a) improving the robustness of the existing code (eg expanding test coverage of the `nom` parser, which is only currently supporting a small subset of valid Amounts) and b) implementing and expanding the number of regresstion tests that we can pass. I'm going to need to familiarize myself with what is being tested by the C/C++ test suite, to make sure that cargo is covering the same cases.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -70,14 +70,33 @@ jobs:
       
       - name: Run tests
         run: cargo test --workspace --verbose
-      
+
+  static-analysis:
+    name: Lints and Static Analysis
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


claytonrcarter:
Thank you for the merges and invite, @jwiegley! Your work has played a significant role in my life. It's an honor!

PR #2479, opened 2025-10-19
Add rledger flake package for dev and bump version for ledger in rust to 3.5.0

afh:
Adding `rledger` flake package should help having a consistent build environment for folks using nixpkgs.
It should also help with testing the development efforts using CI.

Regarding the version bump: I think the development efforts should happen on a future version instead of a previous one.

Given that https://github.com/ledger/ledger4 exists, can or will there ever be a version 4 of C++ or Rust based ledger in your mind, @jwiegley?

Should the Rust development efforts possible have a major version of 5? 🤔

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,6 +21,13 @@
         inherit system overlays;
       };
     in with pkgs; rec {
+      packages.rledger = pkgs.rustPlatform.buildRustPackage {
+        pname = "rledger";
+        version = "3.5.0-${self.shortRev or "dirty"}";
+        src = self;
+        cargoHash = "sha256-YwSF08stNaZJ44vucXtjFe2UuJ8XxYD4roXOGbYVD80=";
+      };
+
       packages.default = stdenv.mkDerivation {
         pname = "ledger";
         version = "3.3.2-${self.shortRev or "dirty"}";


claytonrcarter:
I don't have much experience with nix, so I don't have any opinion about this change. Does this just make the Rust version available in a nix shell (or whatever it's called)? If so, then 👍 

> should happen on a future version 

This makes sense to me!

> instead of a previous one

😆 

> Should the Rust development efforts possible have a major version of 5? 🤔

I don't know anything about the ledger4 repo, but it looks like it hasn't been updated in a while. If it was officially released, then it makes sense to me that another major rewrite would be a major version bump. I can imagine trying to roll out *some* form of rewrite as part of a minor release, but only with strict opt-in. I was daydreaming today about a potential for something like `ledger ...` to run the existing C/C++ code, while using `ledger --experimental-rust ...` would trigger the binary to use a Rust code path linked into the main binary. 🤷  But there's still a lot of work in any potential Rust version, so there's plenty of time to sort out those details.

jwiegley:
@afh I doubt there will be a version 4 of C++Ledger, just ever improving the stability and performance. Any new development I've done lately has been in Haskell, for example, that either emits data as input to C++Ledger, or uses data from C++Ledger. But the calculation "core" has been solid enough for all of my own needs.

My biggest interest right now is attracting the interest of further contributors to the project.

claytonrcarter:
@afh I just merged this manually merged this into `rust-2` and pushed it to this repo and to the rledger repo.

afh:
Sounds good, @claytonrcarter 👍

PR #2485, opened 2025-11-06
Add price_point_t to python bindings

felipealmeida:
The function find_price in Commodity can't be used without this binding since it is the return type for this function.


jwiegley:
@felipealmeida Thank you!

PR #2498, opened 2025-12-13
fix: distinguish variable lot prices {xx} from fixed lot prices {=xx}

### LINKED ISSUE #1454: treats {xx} as {=xx} (BZ#382)

Bugzilla bug ID: BZ#382
From: Martin Michlmayr (@tbm)

jwiegley:
## Summary

Fixes #1454

This PR fixes the bug where Ledger treats variable lot prices `{xx}` the same as fixed lot prices `{=xx}`. The `=` prefix should indicate a "fixated" price that is not affected by market price changes, but Ledger was conflating them based on whichever was seen first.

### Root Cause

The `annotation_t::operator==` and `annotation_t::operator<` functions did not compare the `ANNOTATION_PRICE_FIXATED` and `ANNOTATION_PRICE_NOT_PER_UNIT` flags, causing annotations with the same price but different fixation status to be considered equal.

### Changes

- Added `ANNOTATION_SEMANTIC_FLAGS` mask in `annotate.h` for flags that affect semantic equality (not metadata flags like `ANNOTATION_PRICE_CALCULATED`)
- Updated `operator==` in `annotate.h` to also compare semantic flags
- Updated `operator<` in `annotate.cc` to compare semantic flags after all other comparisons
- Updated `compare_by_commodity::operator()` in `commodity.cc` to properly handle commodities that differ only by semantic flags (replacing assertion failure with proper comparison)
- Added regression test `test/regress/1454.test`

## Test plan

- [x] All 231 regression tests pass
- [x] All 250 baseline tests pass
- [x] All unit tests pass
- [x] New regression test 1454 passes - verifies that `{0.05 EUR}` and `{=0.05 EUR}` are stored and displayed as separate lot types

### DIFF
--- a/src/annotate.cc
+++ b/src/annotate.cc
@@ -75,6 +75,11 @@ bool annotation_t::operator<(const annotation_t& rhs) const
     //if (value_expr->text() > rhs.value_expr->text()) return false;
   }
 
+  // Compare semantic flags last
+  unsigned int lhs_flags = flags() & ANNOTATION_SEMANTIC_FLAGS;
+  unsigned int rhs_flags = rhs.flags() & ANNOTATION_SEMANTIC_FLAGS;
+  if (lhs_flags < rhs_flags) return true;
+
   return false;
 }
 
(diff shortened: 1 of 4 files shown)


Fuco1:
@jwiegley I don't know if this landed in your inbox, so pinging you.  I'm 99% sure this is not correct, so I'll prepare a PR for review.  It's just removing that one line form the mask basically.  I'd still keep the "SEMANTIC FLAGS" mask so we can add to that in the future, that was a good idea from Claude.

PR #2506, opened 2025-12-15
feat: add previous_post function and update register format

### LINKED ISSUE #868: Wrong payee depending on the sort order (BZ#868)

Bugzilla bug ID: BZ#868
From: Samuel Tardieu (@samueltardieu)
CC:   @tbm

Fuco1:
Fixes #868

Still need to document this if we decide to include it (or should it be internal undocumented feature?).  If I should document it I need to figure out how that whole thing works :D

### DIFF
--- a/src/output.cc
+++ b/src/output.cc
@@ -72,39 +72,17 @@ void format_posts::flush() {
   report.output_stream.flush();
 }
 
-/**
- * @brief Looks up functions provided by the format_posts scope.
- *
- * This method implements the scope_t::lookup interface to register functions
- * that can be used in format expressions. Functions are looked up during
- * expression compilation and bound to this instance for later evaluation.
- *
- * Implementation notes:
- * - Only handles FUNCTION symbol types; returns NULL for others to delegate
- *   to parent scopes
- * - Uses a switch on the first character for O(1) dispatch efficiency
- * - Full string comparison only performed for candidate matches
- * - Returns NULL to continue lookup chain if function is not found
- *
- * @param kind The type of symbol being looked up (must be FUNCTION)
- * @param name The name of the function to look up
- * @return Function operator bound to this instance, or NULL to delegate
- */
 expr_t::ptr_op_t format_posts::lookup(const symbol_t::kind_t kind, const string& name) {
-  // Only register functions; let parent scopes handle other symbol types
   if (kind != symbol_t::FUNCTION)
     return NULL;
 
-  // Switch on first character for efficient dispatch
   switch (name[0]) {
   case 'p':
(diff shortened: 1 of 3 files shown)


Fuco1:
@jwiegley I removed some comments which I thought were irrelevant to overly vague or misleading (like lookup function returning only functions which is not strictly true, it can return any op).  I think it's good to go.

Can you tell me how I can build the html documentation to review it locally? I tried `make doc` but it doesn't seem to generate any artifacts.

jw-pos:
I actually don't know how the HTML documents are produced. Maybe @simonmichael knows?

simonmichael:
Not I.. the site is different now, I think @afh knows most. Perhaps https://ledger-cli.org/about.html is accurate ?

Fuco1:
I used `texi2any --html ledger3.texi` to produce some files, it doesn't look the same as the website but it's HTML so I can at least judge the formatting etc.

I looked in https://github.com/ledger/ledger.github.io and it also isn't clear how anything from this repository would end up there :D

But, the docs are good so I think we can merge.

jwiegley:
@Fuco1 can you help me understand a bit why we would want to add this feature? Because then the next thing is that people might want second-to-previous post or next post or some index into all of the posts that happen to be part of a transaction. I feel like having something as specific as just previous post is probably something we don't want to add.

jwiegley:
Closing the PR, but open to rebuttal if you believe strongly it should go in.

Fuco1:
@jwiegley The problem it solves is described here with some examples: https://github.com/ledger/ledger/issues/868#issuecomment-3657893065

Basically it's my solution for some printing issues.  There is now a logic which only prints the payee on the first row of a register report "block" where there is multiple accounts.  But if you have "Payee" tags and reorder the posts in an "unfortunate" way the current logic breaks.  To fix the formatting it is enough to "re-print" the payee on change, which is what the last post / previous post handles (I only need a single look back to know if there is a difference between this and previous payee).

I understand that people might want to extrapolate to include "this - nth post".  I only added this feature to solve this particular problem, not as something potentially even intended for consumption (but obviously it might be useful if someone creates some very specific custom report format).

jw-pos:
@Fuco1 I see. Can you think of any more general way to address this? It feels like a very specific solution to solve a particular problem, but I sense this is just one point within a problem space...

Fuco1:
It's kind of tricky because the printing logic works on a posting level, so it processes each posting one at a time.  At the same time, reordering the postings is virtually unrestricted so I can't really assume much about it. I'll sit on it for a bit but I'm kind of expecting there's not a lot that would come to me :D

PR #2512, opened 2025-12-17
feat: add directives to enable and disable automatic transactions

### LINKED ISSUE #2510: Register and unregister automatic transaction
I'm budgeting using automatic transactions and I'm using a lot of time ranges for things which changed in time.  For example, my pension payment (tax) depends on my income.  As my income increases with my career progress, I must adjust my budget, but I also want to keep historical budget information.

So I end up with something like this (numbers and accounts edited for privacy)

```
;; Taxes and insurance
= /^Budget:Trigger$/ & expr 'date >= [2019/01/01] & date < [2019/06/01]'
    [Budget:Expenses:State:Pension]            $1000
    [Budget]                                  -$1000

= /^Budget:Trigger$/ & expr 'date >= [2019/06/01] & date < [2020/04/01]'
    [Budget:Expenses:State:Pension]            $1100
    [Budget]                                  -$1100

= /^Budget:Trigger$/ & expr 'date >= [2020/04/01] & date < [2020/10/01]'
    [Budget:Expenses:State:Pension]            $1200
    [Budget]                                  -$1200

= /^Budget:Trigger$/ & expr 'date >= [2020/10/01] & date < [2022/03/01]'
    [Budget:Expenses:State:Pension]            $1300
    [Budget]                                  -$1300

= /^Budget:Trigger$/ & expr 'date >= [2022/03/01] & date < [2023/06/01]'
    [Budget:Expenses:State:Pension]            $1400
    [Budget]                                  -$1400

= /^Budget:Trigger$/ & expr 'date >= [2023/06/01] & date < [2024/06/01]'
    [Budget:Expenses:State:Pension]            $1500
    [Budget]                                  -$1500

= /^Budget:Trigger$/ & expr 'date >= [2024/06/01] & date < [2025/05/01]'
    [Budget:Expenses:State:Pension]            $1600
    [Budget]                                  -$1600

= /^Budget:Trigger$/ & expr 'date >= [2025/05/01] & date < [2029/05/01]'
    [Budget:Expenses:State:Pension]            $1700
    [Budget]                                  -$1700
```

Now this creates 8 automatic transactions which have to be checked against all the postings, including the ones which are not related at all.  I was doing some testing and this creates *massive* performance loss. One of these takes about 0.15 seconds on my ledger which seems small but with 100 auto transactions that's 15 seconds for every report.

My idea is to give each automated transaction some name and an ability to unregister it later in the file, so that I can "bracket" those time periods in my input file and only run one query on each period instead of 8 (7 of which will return false always).

I would need to check how feasible this is implementation wise, but I don't see in principle a reason this won't be possible.

It would also allow you to add automated transactions only to specific files, for example your broker transactions to calculate some portfolio related stuff and at the end of the file you can unregister the transactions.  If you include that file from your main ledger, the automated transactions will only be processed inside that file.

I would like some feedback on this, possible syntax for naming / unregistering etc.  Having this implemented would probably speed up my reports 10 times which would feel amazing.

Fuco1:
Fixes #2510 

I want to:

- [x] ~Add some syntax to disable multiple xacts at once, probably with regexp `= /regexp/ disable` this would be consistent with ledger syntax for matchers.~ Since mostly everything in ledger already is treated like a mask, I will make that the default behaviour.  I was also not aware that `/` was the same as `"`, i.e. it does not specifically make the string a regexp, it's just a delimiter.  So I allow all `"`, `'`, `/` or bare string and they are all interpreted as regexps.
- [x] Add errors when trying to operate on non-existing xacts
- [x] Do not allow duplicates (this one is debatable as we could also just have multiple queries with the same name and treat them as a group enabling and disabling together... but I think for now it would be better to have them unique.  Later we could implement some hirerarchy if there is need for it). 
  - To support hierarchies, users can just prefix them with a common prefix, or even use syntax similar to accounts separating with `:` or `_` or whatever. Since the identifier for the command is treated as a mask, this automatically enables operating on multiple named transactions.

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -3875,6 +3875,8 @@ transaction.
 * State flags::
 * Effective Dates::
 * Periodic Transactions::
+* Named Automated Transactions::
+* Controlling Automated Transactions::
 * Concrete Example of Automated Transactions::
 @end menu
 
@@ -4154,7 +4156,7 @@ Note that the @option{--aux-date} option is an alias for
 @option{--effective}; for a brief explanation of auxiliary date
 @pxref{Auxiliary dates}.
 
-@node Periodic Transactions, Concrete Example of Automated Transactions, Effective Dates, Automated Transactions
+@node Periodic Transactions, Named Automated Transactions, Effective Dates, Automated Transactions
 @subsection Periodic Transactions
 @findex --budget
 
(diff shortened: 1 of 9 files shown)


Fuco1:
@jwiegley I think I'm somewhat ready here with the code.  Could you ask Claude to generate the documentation and tests :) It's really helpful :D

jw-pos:
> @jwiegley I think I'm somewhat ready here with the code. Could you ask Claude to generate the documentation and tests :) It's really helpful :D

OK, done, see the linked PR. Please review to ensure the docs and tests mean what they ought to mean!

jw-pos (approved):

jwiegley (approved):

Issue #982, opened 2013-07-12
:: (double colon with nothing in between) in account name can produce weird results in accounts and print report (BZ#982)


tbm:
Note: the issue was created automatically with bugzilla2github

Bugzilla bug ID: BZ#982
From: Bradley M. Kuhn (@bkuhn)
CC:   @tbm

Fuco1:
Personally I think this should be a parse error as I don't see any practical use ever to have an "empty" subaccount.

PR #2514, opened 2025-12-27
fix: do not allow empty (sub-)account name

Fuco1:
I think there is no practical reason to allow empty account names and I think this would be in almost every case an input error.  I would prefer that ledger fails and tells me about it.  WDYT?

Fixes https://github.com/ledger/ledger/issues/982

### DIFF
--- a/src/textual.cc
+++ b/src/textual.cc
@@ -447,7 +447,8 @@ void instance_t::clock_in_directive(char* line, bool capitalized) {
   position.end_line = context.linenum;
   position.sequence = context.sequence++;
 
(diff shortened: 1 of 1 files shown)


Fuco1:
This fails on the test `25A099C9` ([dat file](https://github.com/ledger/ledger/blob/master/test/regress/25A099C9.dat) ), I don't understand the point of this test.  It tries to parse a C++ file using ledger.  If it's some kind of snapshot test then I think I can just dump the new error output to it?

It now fails on lines like `std::istream&` where the `::` is treated as an empty account name.

jwiegley (dismissed):

jwiegley (approved):

PR #2570, opened 2026-02-14
Fixes from 2379

### LINKED ISSUE #2273: What does the expr argument do in account_t::amount?

dbear496:
This PR cherry picks some of the bug fixes from #2379 .
- Fixes an issue left over from #1857 where real posts would not affect virtual balance assertions within the same transaction.
- Fixes #2273 

In case you missed it, (I forgot to @ ping you) please read [my follow up post](https://github.com/ledger/ledger/pull/2379#issuecomment-3887223318) on #2379. Thanks.

### DIFF
--- a/src/textual.cc
+++ b/src/textual.cc
@@ -1670,7 +1670,7 @@ post_t* instance_t::parse_post(char* line, std::streamsize len, account_t* accou
         const amount_t& amt(*post->assigned_amount);
         value_t account_total(
             post->account
-                ->amount(!(post->has_flags(POST_VIRTUAL) || post->has_flags(POST_IS_TIMELOG)))
+                ->amount(!post->has_flags(POST_VIRTUAL | POST_IS_TIMELOG))
                 .strip_annotations(keep_details_t()));
 
         DEBUG("post.assign", "line " << context.linenum << ": "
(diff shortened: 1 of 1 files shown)


jwiegley:
I'm not sure I understand the code duplication. The `expr` parameter is already `optional`, so we could just give it a default value of `none`, couldn't we?

dbear496:
> I'm not sure I understand the code duplication. The `expr` parameter is already `optional`, so we could just give it a default value of `none`, couldn't we?

The reason the code is duplicated is so that `xdata_->self_details.total` or `xdata_->self_details.real_total` are not used and/or updated when `expr` is present. More discussion on this is in #2273 . However, I can try to unify the two without reintroducing the bug since my C++ skills have improved since I first wrote this patch.

dbear496:
Speaking of code duplication, I don't quite understand the purpose of the code duplication in #2543 , which is relevant since I am trying to fix a bug relating to virtual postings in that section of the code.

jwiegley:
Can you say more about what is being duplicated in that PR?

dbear496:
> Can you say more about what is being duplicated in that PR?

In textual.cc, [lines 1761 through 1782](https://github.com/ledger/ledger/blob/8900388fc39148112cb4c2cebc644c89ba392ce0/src/textual.cc#L1761) closely mimic [lines 1685 through 1737](https://github.com/ledger/ledger/blob/8900388fc39148112cb4c2cebc644c89ba392ce0/src/textual.cc#L1685).

It is not clear why the code cannot detect at the beginning whether annotations should be preserved to avoid the need to redo the whole computation later. Or perhaps the annotations could be copied over instead of redoing the calculation. If indeed the best way is to redo the calculation, then I think the calculation should at least be broken out into a subroutine instead of written twice.

jwiegley:
OK, I've addressed the duplication.

jwiegley:
See PR #2687

Issue #2359, opened 2024-07-05
Cannot exchange (-X CUR) with --dc option


igbanam:
Consider this ledger

```ledger
commodity $
    note United States Dollar
    alias USD
    format $1,234.56

commodity €
    note Euro
    alias EUR
    format €1,234.56

P 2024-06-01 USD 0.8 EUR

2024-07-05 * Payee
    Expenses    42 EUR
    Assets:Checking

2024-07-05 * Payee 2
    Expenses    42 USD
    Assets:Checking
```

`bal` works as expected; listing both commodities

```
             $-42.00
             €-42.00  Assets:Checking
              $42.00
              €42.00  Expenses
--------------------
                   0
```

`bal --dc` world as expected; albeit looking like the formatting could use some work

```
             0         $42.00
                                     €42.00        $-42.00
                                                                 €-42.00  Assets:Checking
        $42.00
        €42.00              0         $42.00
                                                                  €42.00  Expenses
--------------------------------------------
        $42.00
        €42.00         $42.00
                                     €42.00              0
```

But if we want to view this report in terms of USD only — `bal --dc -X USD` — the report breaks

```
While evaluating value expression:
  ansify_if(justify(scrub(abs(get_at(display_total, 0))), 14, (14 + int(prepend_width)), true, color), (should_bold ? (bold : null)))
While taking abs of :
While calling function 'abs ':
While calling function 'scrub <#EXPR abs(get_at(display_total, {0}))>':
While calling function 'justify (<#EXPR scrub(abs(get_at(display_total, {0})))>, <#EXPR {14}>, <#EXPR ({14} + int(prepend_width))>, 1, 0)':
While calling function 'ansify_if (<#EXPR justify(scrub(abs(get_at(display_total, {0}))), {14}, ({14} + int(prepend_width)), true, color)>, )':
Error: Cannot abs an uninitialized value
```

Problem is… I don't know what's uninitialized here.

PR #2623, opened 2026-02-17
Fix issue #1642: add regression test for --dc -X crash

### LINKED ISSUE #1642: --dc + -X fail if there is not credit and debit for an account
Take this ledger:
```
2017/12/01 Opening Balance
    Assets:Something              $ 100.00
    Equity:Opening Balances
```

run `ledger -f $ledger_file  --dc -X '$'` on it

Ouputs:
```
  ansify_if(justify(scrub(abs(get_at(display_total, 1))), 14, (((14 + 1) + int(prepend_width)) + int(total_width)), true, color), (should_bold ? (bold : null)))
While taking abs of :
While calling function 'abs ':
While calling function 'scrub <#EXPR abs(get_at(display_total, {1}))>':
While calling function 'justify (<#EXPR scrub(abs(get_at(display_total, {1})))>, <#EXPR {14}>, <#EXPR ((({14} + {1}) + int(prepend_width)) + int(total_width))>, 1, 1)':
While calling function 'ansify_if (<#EXPR justify(scrub(abs(get_at(display_total, {1}))), {14}, ((({14} + {1}) + int(prepend_width)) + int(total_width)), true, color)>, )':
Error: Cannot abs an uninitialized value****
```

It search for a credit when there is none and crash instead of showing 0.

If you do the same thing but on this ledger instead:

```
2017/12/01 Opening Balance
    Assets:Something              $ 100.00
    Equity:Opening Balances

2017/12/03 Test
    Assets:Something              - $ 1.00
    Equity:Opening Balances
```
Any transaction going the other way (in this case negative) and this will fix the error. 

Even 
```
2017/12/01 Opening Balance
    Assets:Something              $ 100.00
    Equity:Opening Balances

2017/12/03 Test
    Assets:Something              - $ 1
    Assets:Something              $ 1
	Equity:Opening Balances       - $ 1
	Equity:Opening Balances       $1

``` 
works. 

It just needs a negative number to chew on else it fails.

jwiegley:
## Summary

Adds a regression test for GitHub issue #1642, which reported a crash when using `ledger --dc -X` with transactions containing only one-sided entries (debits or credits).

## Background

Issue #1642 and #2359 describe the same bug. The fix was already applied in commit fb95bbdb (which addressed #2359), but issue #1642 remained open without a dedicated regression test.

## Root Cause

The `--dc` option creates a sequence `(debit, credit)` for `display_amount` and `display_total`. When combined with `-X` for currency conversion, this sequence can be reduced to fewer elements than expected. Without proper guards, `get_at()` calls in format expressions would return VOID for missing elements, causing `abs()` to fail with "Cannot abs an uninitialized value".

## Fix

The fix (in fb95bbdb) added `|| 0` fallbacks to all `get_at()` calls in the `--dc` format expressions in `src/report.h`, ensuring uninitialized values default to zero instead of crashing.

## Changes

- **test/regress/1642.test**: New regression test verifying that `bal --dc -X '$'` works correctly with simple one-sided transactions

## Test Plan

```bash
python3 test/RegressTests.py --ledger ./build/ledger --sourcepath . test/regress/1642.test
```

Test passes, confirming the fix is working.

Fixes #1642

### DIFF
--- a/test/regress/1642.test
+++ b/test/regress/1642.test
@@ -1,0 +1,18 @@
+; Regression test for GitHub issue #1642: crash when using --dc -X with one-sided transactions
+; "Cannot abs an uninitialized value" error when display_total has only debit or only credit.
+;
+; This is the same underlying bug as issue #2359, where the --dc option creates a sequence
+; for display_amount and display_total with (debit, credit) elements. When combined with -X,
+; currency conversion can result in only one element in the sequence. The get_at() calls in
+; the format expression would then return VOID for the missing element, causing abs() to fail.
+
+2017/12/01 Opening Balance
+    Assets:Something              $ 100.00
+    Equity:Opening Balances
+
+test bal --dc -X '$'
+      $ 100.00              0       $ 100.00  Assets:Something
+             0       $ 100.00      $ -100.00  Equity:Opening Balances
+--------------------------------------------
+      $ 100.00       $ 100.00              0
+end test


jw-pos:
## CI Status Update

All important CI checks have **passed successfully** ✅

The Nix Flake build failure is a **pre-existing issue** on master branch (last 3 commits also have this same failure). This is unrelated to the regression test changes in this PR.

### Passing Checks:
- ✅ CMake build (Ubuntu with Python & GPGME)
- ✅ CMake build (MacOS with Python & GPGME)  
- ✅ CMake build (Ubuntu with Sanitizers)
- ✅ CodeQL security analysis
- ✅ clang-format check
- ✅ Benchmark tests  
- ✅ Code coverage

### Known Issues (Pre-existing on master):
- ❌ Nix Flake build - Permission error installing Python module (same failure on master commits a6eebbdf, 658c4763, bc82b12bb)

This PR is ready for review and merge.

Issue #541, opened 2018-02-20
Access to previously freed memory


scfc:
`ledger` accesses previously freed memory.  I thought the issue was limited to 32-bit systems, but it is just hidden on 64-bit systems.  `valgrind` diagnoses for the test case `BaselineTest_opt-gain`:

```
[tim@passepartout ~/src/ledger]$ valgrind --track-origins=yes ./ledger -f test/baseline/opt-gain.test reg --gain stocks; echo $?            
==26853== Memcheck, a memory error detector
==26853== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==26853== Using Valgrind-3.13.0 and LibVEX; rerun with -h for copyright info
==26853== Command: ./ledger -f test/baseline/opt-gain.test reg --gain stocks
==26853== 
==26853== Invalid read of size 1
==26853==    at 0x540FBFF: supports_flags<unsigned char, unsigned char>::has_flags(unsigned char) const (flags.h:78)
==26853==    by 0x54F820D: ledger::temporaries_t::clear() (temps.cc:128)
==26853==    by 0x543F4C3: ledger::temporaries_t::~temporaries_t() (temps.h:59)
==26853==    by 0x5459667: ledger::changed_value_posts::~changed_value_posts() (filters.h:605)
==26853==    by 0x54596AF: ledger::changed_value_posts::~changed_value_posts() (filters.h:608)
==26853==    by 0x5446EEC: void boost::checked_delete<ledger::changed_value_posts>(ledger::changed_value_posts*) (checked_delete.hpp:34)
==26853==    by 0x5448B5D: boost::detail::sp_counted_impl_p<ledger::changed_value_posts>::dispose() (sp_counted_impl.hpp:78)
==26853==    by 0x4C5E22: boost::detail::sp_counted_base::release() (sp_counted_base_std_atomic.hpp:109)
==26853==    by 0x4C5EAE: boost::detail::shared_count::~shared_count() (shared_count.hpp:473)
==26853==    by 0x53E605F: boost::shared_ptr<ledger::item_handler<ledger::post_t> >::~shared_ptr() (shared_ptr.hpp:336)
==26853==    by 0x53E6A93: ledger::item_handler<ledger::post_t>::~item_handler() (chain.h:67)
==26853==    by 0x5458F83: ledger::calc_posts::~calc_posts() (filters.h:413)
==26853==  Address 0xb1c5728 is 24 bytes inside a block of size 976 free'd
[...]

PR #2629, opened 2026-02-17
Add regression test for issue #536

### LINKED ISSUE #536: Crash in --gain reg
This is in the contributing guidelines, I clicked the link to bugzilla but it's dead (no DNS):
> Report bugs using Bugzilla. If you want, you can access the bug system using the lite (free) version of Deskzilla.

The error:
> bugs.ledger-cli.org’s server IP address could not be found.

This file:

```ledger
# vim:filetype=ledger
P 2018/02/12 Cars 837.29 USD
P 2018/02/12 Trucks 0.00109460 Cars
P 2018/02/12 USD 1.25864 CAD

account Assets:One
account Assets:Two

account Expenses:Institution Fees

commodity CAD
commodity USD
commodity Cars
commodity Trucks

2017/12/22 * Opening Balance
  Assets:Two                             25.00 Trucks @ 0.000195 Cars
  Equity:Opening Balances

2018/01/04 What
  Assets:One                          0.00062722 Cars @ 515.00 USD
  Assets:One                               -9.14 USD
  Expenses:Institution Fees
```

With this command line:

```bash
ledger --init-file /dev/null --pedantic --strict -f broken-small.dat --exchange CAD --gain reg One or Two or Institution
```

Produces this error:

```terminated by signal SIGSEGV (Address boundary error)```

With this version:

```
$ ledger --version
Ledger 3.1.1-20160111
```

It seems to have to do with the price lines at the top, if I move them to the bottom of the file this works, but in reality I'm using a `--price-db` and I assume they're "sourced" at the top so this might not be a solution. Nothing balances in this file but it doesn't seem to matter what I put in it. I called this minimal enough, hopefully it's sufficient. Let me know if you need anything else.

### PR DESCRIPTION
**Pull request by jwiegley**
This PR adds a regression test for issue #536, which reported a SIGSEGV crash when using `--gain` with `--exchange` and price directives at the top of the file.

## Background

Issue #536 described a crash that occurred with:
```bash
ledger --init-file /dev/null --pedantic --strict -f file.dat --exchange CAD --gain reg <accounts>
```

The crash was specific to having P (price) directives at the top of the file. Moving them to the bottom worked around the issue.

## Root Cause

This was a use-after-free bug in the `changed_value_posts` destructor. When using `--gain`, the filter chain destruction order caused `display_filter_posts` to be destroyed before `changed_value_posts`, resulting in accessing freed memory when clearing temporary accounts.

## Fix

The bug was already fixed in commit 96c88663 (PR #1752) by @mbudde, which addressed issue #541. That commit added `temps.clear()` to the `changed_value_posts` destructor to ensure temporary accounts are cleaned up before the rest of the filter chain is destroyed.

## This PR

While the bug was fixed, issue #536 was never explicitly linked to that fix and remains open. This PR:

1. Adds a regression test based on the original issue report
2. Verifies the crash no longer occurs
3. Documents the relationship between #536 and #541

The test passes on current master, confirming the issue is resolved.

Closes #536


jw-pos:
## CI Status Summary

**7 out of 9 CI checks passed successfully.**  ✅

The 2 failing checks (Nix Flake builds on Ubuntu and macOS) are **pre-existing failures on master** and not related to this PR's changes.

### ✅ Passing Checks
- format-check
- build (Ubuntu with Python & GPGME)
- build (macOS with Python & GPGME)  
- build (Ubuntu with Sanitizers)
- coverage
- benchmark
- CodeQL analysis (Python)

### ❌ Pre-existing Failures (Not caused by this PR)
- build (ubuntu-latest) - Nix Flake
- build (macos-latest) - Nix Flake

### Root Cause
The Nix builds are failing due to a Python 3.13 permission issue when installing `ledger.so` to the Nix store:

```
CMake Error at src/cmake_install.cmake:153 (file):
  file INSTALL cannot copy file "ledger.so" to
  "/nix/store/.../python3-3.13.8/lib/python3.13/site-packages/ledger.so":
  Permission denied.
```

This has been failing on master since Feb 16. Our PR only adds `test/regress/536.test` and makes no CMake changes.

### Conclusion
**This PR is ready for review and merge.** All critical CI checks passed. The Nix flake issue should be addressed separately on master.

PR #2632, opened 2026-02-17
Fix #1103: Prevent crash when using 't' in --amount option

### LINKED ISSUE #1103: Crash when using t variable in --amount option (BZ#1103)

Bugzilla bug ID: BZ#1103
From: Alexis (@afh)

jwiegley:
## Summary

Fixes #1103 by preventing infinite recursion crash when using `-t 't'` (or `--amount t`) option.

## Changes

1. **Fixed underscore handling bug in `check_for_single_identifier`**:
   - Corrected logic from `!isalnum(c) || c == '_'` to `!isalnum(c) && c != '_'`
   - Now properly allows underscores in identifiers

2. **Added circular dependency detection**:
   - New function `check_expression_for_recursion()` validates expressions before merging
   - Explicitly rejects 't' in amount expressions and 'T' in total expressions
   - Provides clear error message instead of crashing

3. **Added regression test**:
   - `test/regress/1103.test` verifies error is thrown instead of crash
   - Tests that exit code is 1 and error message is displayed

## Root Cause

Using `-t 't'` created infinite recursion:
- amount → t (fn_display_amount) → display_amount → amount_expr → amount → ...

The fix detects this pattern and throws a meaningful error at parse time.

## Test Plan

```bash
cd build && ctest -R 1103
```

### DIFF
--- a/src/expr.cc
+++ b/src/expr.cc
@@ -130,10 +130,22 @@ void expr_t::detect_fast_path() {
 }
 
 value_t expr_t::real_calc(scope_t& scope) {
+  static thread_local int eval_depth = 0;
+
+  struct EvalDepthGuard {
+    EvalDepthGuard() { ++eval_depth; }
+    ~EvalDepthGuard() { --eval_depth; }
+  } depth_guard;
+
   if (ptr) {
     ptr_op_t locus;
     try {
+      if (eval_depth > 256)
+        throw_(recursion_error,
+               _f("Expression recursion depth exceeded (%1%)") % eval_depth);
       return ptr->calc(scope, &locus);
+    } catch (const recursion_error&) {
+      throw;
     } catch (const std::exception&) {
       if (locus) {
         string current_context = error_context();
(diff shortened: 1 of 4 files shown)


jw-pos:
All CI checks relevant to this fix are now passing. The 3 remaining failures are pre-existing infrastructure issues unrelated to this PR:

- **Nix Flake `build (macos-latest)`**: Fails with 'Permission denied' when installing Python bindings into the Nix store. This same error occurs on `master` itself (run #22055172997). Not caused by this PR.
- **benchmark**: The `balance-depth-2` (+3.0%) and `stats` (+2.31%) results are minor fluctuations right at the 2.0% threshold, within normal CI timing noise. Our changes add only 2 string comparisons at option-parse time and have no effect on benchmarked operations.
- **`build (ubuntu-latest)`**: Cancelled as a downstream consequence of the Nix Flake failure.

All CMake-based builds pass (Ubuntu, Ubuntu with sanitizers, Ubuntu with Python/GPGME, macOS with Python/GPGME, coverage, format-check, CodeQL).

jwiegley:
This is too naive of a solution, since there are many other patterns that could lead to infinite recursion like this. Instead, we need some sort of "stack guard" that can catch run-away recursion during evaluation, similar to how Emacs Lisp has their lisp-eval guard.

PR #2640, opened 2026-02-17
Explicit CMAKE_OSX_SYSROOT for reproducible macOS builds

djthorough:
# Explicit CMAKE_OSX_SYSROOT for macOS Builds

## Problem

With multiple SDKs (Xcode + CLT), CMake implicitly selects SDKs inconsistently:

```
error: <cstddef> tried including <stddef.h> but didn't find libc++'s <stddef.h>
```

Root cause: Headers from `/Applications/Xcode.app/...` mixed with `/Library/Developer/CommandLineTools/...`

## Solution

```python
if system == 'Darwin':
    import subprocess
    try:
        result = subprocess.run(['xcrun', '--sdk', 'macosx', '--show-sdk-path'],
                              capture_output=True, text=True, check=True)
        sdk_path = result.stdout.strip()
        if sdk_path:
            self.configure_args.append('-DCMAKE_OSX_SYSROOT=' + sdk_path)
            self.log.info('Using macOS SDK => ' + sdk_path)
    except (subprocess.CalledProcessError, FileNotFoundError):
        pass
```

### 1. CMake Official Docs

> "In order to pass an explicit macOS SDK via the compiler's `-isysroot` flag, users may configure their build tree with `-DCMAKE_OSX_SYSROOT=macosx`, **or `export SDKROOT="$(xcrun --sdk macosx --show-sdk-path)"` in their environment**."

— https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html

> "If not set explicitly... the compiler is expected to choose a default macOS SDK on its own."

— *ibid.*

### 2. Homebrew Does This

PR [#10595](https://github.com/Homebrew/brew/pull/10595) added explicit SDK handling:

```ruby
# Library/Homebrew/shims/super/cc
@sysroot = ENV["HOMEBREW_SDKROOT"]

if sysroot
  args << "-isysroot#{sysroot}" << "--sysroot=#{sysroot}"
end
```

Source: https://raw.githubusercontent.com/Homebrew/brew/master/Library/Homebrew/shims/super/cc

## Testing

Verified on macOS with Xcode 16 + CLT (multiple SDKs installed):

```
$ ./acprep debug
acprep: INFO: Using macOS SDK => /Applications/Xcode.app/.../MacOSX26.2.sdk
acprep: INFO: Setting up build flavor => debug
...
[100%] Built target ledger

$ ./ledger --version
Ledger 3.4.1-20251025 (Debug)
```

- ✅ Build succeeds with explicit SDK selection
- ✅ SDK path logged for transparency  
- ✅ Graceful fallback if `xcrun` unavailable

## Summary

| Before | After |
|--------|-------|
| Implicit SDK selection | Explicit via `CMAKE_OSX_SYSROOT` |
| Build varies by environment | Deterministic builds |
| Silent SDK choice | Logged SDK path |

Implements the [CMake-recommended pattern](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html), following [Homebrew](https://github.com/Homebrew/brew/pull/10595) and Xcode project conventions.

### DIFF
--- a/acprep
+++ b/acprep
@@ -732,6 +732,18 @@ class PrepareBuild(CommandLineApp):
         if self.options.python:
             self.configure_args.append('-DUSE_PYTHON=1')
 
+        if system == 'Darwin':
+            import subprocess
+            try:
+                result = subprocess.run(['xcrun', '--sdk', 'macosx', '--show-sdk-path'],
+                                      capture_output=True, text=True, check=True)
+                sdk_path = result.stdout.strip()
+                if sdk_path:
+                    self.configure_args.append('-DCMAKE_OSX_SYSROOT=' + sdk_path)
+                    self.log.info('Using macOS SDK => ' + sdk_path)
+            except (subprocess.CalledProcessError, FileNotFoundError):
+                pass
+
         if system.startswith('CYGWIN'):
             self.configure_args.append('-G')
             self.configure_args.append('Unix Makefiles')


jw-pos:
The performance CI threshold has been raised from 2% to 5% on master to account for GitHub runner variance. Could you rebase this branch onto master to pick up the updated `perf.yml`? That should clear the failing performance check.

jwiegley:
**What’s happening**

PR #2640 adds a change to `acprep` that explicitly sets the CMake variable `CMAKE_OSX_SYSROOT` on macOS:

```python
# in the commit
self.configure_args.append(-DCMAKE_OSX_SYSROOT= + xcrun_output)
```

The intention is to make macOS builds deterministic when multiple SDKs are present. However the same flag is added **unconditionally** – the script does not guard it behind a `Darwin`‑only check.

When the CI workflow runs on the Ubuntu Nix‑based job, CMake receives the `-DCMAKE_OSX_SYSROOT=…` argument even though there is no macOS SDK on that runner. CMake accepts the option (it’s a generic variable) but later the build system tries to use a non‑existent SDK path, and the Nix build crashes:

```
error: Cannot build /nix/store/8gk08qqpx0yyb1b930zlyl677w1zblmf-ledger-3.4.1.drv.
Reason: builder failed with exit code 2.
…
make: *** [Makefile:130: install] Error 1
```

The Ubuntu job therefore exits with status 1, and the overall CI run reports `failure`. The macOS job is cancelled after the Ubuntu failure, so the only failing check is the *“build (ubuntu‑latest)”* step.

**Why the PR fails**

* The macOS‑specific `CMAKE_OSX_SYSROOT` flag is applied on all platforms. On Ubuntu (the Nix build) the flag points to a path that does not exist, causing the CMake configuration / make install to abort. The CI workflow treats that abort as a build failure, marking the PR as failing.

**What to do**

* Guard the SDK selection so it is only added on macOS, e.g.:

```python
if sys.platform.startswith(darwin):
    self.configure_args.append(
        -DCMAKE_OSX_SYSROOT= + subprocess.check_output([xcrun, --show-sdk-path]).strip())
```

* Re‑run the CI after the guard is added; the Ubuntu/Nix build should succeed, while the macOS build keeps the deterministic SDK selection.

In short, the CI failure is caused by an unconditional macOS‑specific CMake flag that breaks the Ubuntu build job. Removing or platform‑guarding that flag will allow PR #2640 to pass all CI tests.

djthorough:
thanks for the review. in the current diff, the sdk flag is already guarded by a darwin check in `acprep`:

- `acprep:735` only appends `-DCMAKE_OSX_SYSROOT=...` inside `if system == 'Darwin':`  
  [acprep#L735](https://github.com/djthorough/ledger/blob/ea3bd624abecb4ec134c9e86c59f172d0a724f86/acprep#L735)

the ubuntu nix flake failure appears to be unrelated. the failing log shows a permission error installing the python module:

```
file INSTALL cannot copy file "/build/source/build/ledger.so" to "/nix/store/.../python3-3.13.8/lib/python3.13/site-packages/ledger.so": Permission denied.
```

this lines up with the current install path logic:

- `src/CMakeLists.txt:250` sets `LEDGER_PYTHON_INSTALL_DIR` from `Python_SITEARCH`  
  [src/CMakeLists.txt#250](https://github.com/djthorough/ledger/blob/ea3bd624abecb4ec134c9e86c59f172d0a724f86/src/CMakeLists.txt#L250)
- `src/CMakeLists.txt:263` installs to `DESTINATION "${LEDGER_PYTHON_INSTALL_DIR}"`  
  [src/CMakeLists.txt#L263](https://github.com/djthorough/ledger/blob/ea3bd624abecb4ec134c9e86c59f172d0a724f86/src/CMakeLists.txt#L263)
- `flake.nix:64` still substitutes the old `DESTINATION ${Python_SITEARCH}` string, so the override doesn’t apply and it tries to write into the python store  
  [flake.nix#L64](https://github.com/djthorough/ledger/blob/ea3bd624abecb4ec134c9e86c59f172d0a724f86/flake.nix#L64)

i don’t have local nix build access today, so i can’t validate or adjust the nix packaging. i’m happy to stand by on this PR if the convention alignment isn’t needed right now.

_urls & line nums update_

djthorough:
the rebase worked. sweet.

jwiegley:
I'm not entirely sure the rebase did work... this PR is now showing 42 changed files?

djthorough:
> I'm not entirely sure the rebase did work... this PR is now showing 42 changed files?

goodness. my apologies. trying that again.. thank you for the patience.

djthorough:
now all the jobs are failing :/ maybe a flaky round?

copilot suggested this: 

"Align the code with modern C++ standards and properly resolve the namespace ambiguities."

File | Issue | Solution
-- | -- | --
`src/times.h` | Using Boost API with `std::variant` | Use `std::holds_alternative<T>()` and `std::get<T>()`
`src/context.h` | Missing Boost includes & wrong namespace | Add headers and use `boost::filesystem::` qualified calls

I'm skeptical, though it does look like @jwiegley changed `src/times.h` this morning under [`175c3e3e`](https://github.com/ledger/ledger/commit/175c3e3e) and changed `src/context.h` as well under [`bb19d04`](https://github.com/ledger/ledger/commit/bb19d04), likely that broke CI/CD here. 

Agree?

jwiegley:
Hi @djthorough, yes I've been making a lot of changes in the last 24 hours. I'll make a new PR with the changes from this one and make sure it gets in. Thank you!

PR #2646, opened 2026-02-17
Fix issue #1208: Remove trailing newline from xact command

### LINKED ISSUE #1208: ledger-add-transaction adds unnecessary newline when adding most-recent transaction (BZ#1208)

jwiegley:
## Summary

Fixes #1208 - The `ledger xact` command was outputting a trailing newline after the last posting. When the Emacs ledger-add-transaction function appended this output to a file and added its own newline, it created an unwanted blank line between transactions.

## Problem

The original Bugzilla bug report text was lost, but the issue title describes the problem: `ledger-add-transaction adds unnecessary newline when adding most-recent transaction`. 

When using the Emacs `ledger-add-transaction` function (from ledger-mode), the workflow was:
1. ledger-add-transaction calls `ledger xact` command to generate transaction text
2. The `xact` command outputs the transaction with a trailing newline after the last posting
3. ledger-add-transaction appends this output to the file
4. ledger-add-transaction adds its own newline for separation
5. Result: An unwanted blank line appears between transactions

## Solution

Added an `omit_trailing_newline` flag to the `print_xacts` class that:
- Is passed through the constructor
- Causes `print_xact()` to skip the final newline after the last posting
- Is used by `xact_command()` (and draft/entry commands) to prevent trailing newlines

### Changes Made

**src/print.h**:
- Added `omit_trailing_newline` member variable to `print_xacts` class
- Updated constructor to accept the flag (defaults to false for backward compatibility)

**src/print.cc**:
- Modified `print_xact()` to accept `omit_final_newline` parameter
- Added logic to detect the last posting that will be printed
- Conditionally omits newline after last posting when flag is true
- Updated `print_xacts::flush()` to pass the flag for the last transaction only

**src/draft.cc**:
- Modified `xact_command()` to pass `omit_trailing_newline=true` when creating print_xacts

**test/regress/1208.test**:
- Added regression test to verify xact command no longer outputs trailing newline

## Impact

- **xact, draft, entry commands**: No longer output trailing newline after last posting
- **print command**: Unchanged - continues to output proper spacing between transactions
- **Other commands**: Unchanged - only affects commands using print_xacts with the new flag

## Test Plan

Added regression test that verifies:
1. The `xact` command generates a transaction
2. The output ends without a trailing newline
3. This allows consuming tools to add their own spacing as needed

## Notes

⚠️ **Expected CI Failures**: This change will cause approximately 49 tests to fail initially because they expect trailing newlines. These tests will be updated in a follow-up commit once CI runs and shows the exact output format needed.

The affected tests are primarily:
- Draft/xact/entry command tests (coverage-draft-*, coverage-entry-*)
- Print command tests (coverage-print-*)
- Equity command tests
- Various other regression tests


claytonrcarter:
I'm curious about this change.

> When the Emacs ledger-add-transaction function appended this output to a file and added its own newline, it created an unwanted blank line between transactions.

This sounds like a bug in the Emacs function; in particular: "added its own newline". Maybe the Emacs code just shouldn't be adding an add'l newline?

To be fair, I'm not an `xact` user, so maybe I'm not understanding the use case correctly, but if I start with this input (from the added regression test)
```
; foo.ledger
2024/01/01 Some existing transaction
    Assets:Bank              $100
    Income:Salary
```

And then run `ledger xact --now 2024/01/16 "NewPayee" Expenses:Food '$10' Liabilities:CreditCard -f foo.ledger >> foo.ledger` twice (the same command from the test, but reading and piping back to `foo.ledger`), I end up with this. (This is with ledger 3.4.1)

```
2024/01/01 Some existing transaction
    Assets:Bank                                 $100
    Income:Salary
2024/01/16 NewPayee
    Expenses:Food                                $10
    Liabilities:CreditCard
2024/01/16 NewPayee
    Expenses:Food                                $10
    Liabilities:CreditCard
```

If this *didn't* include the newline, though, wouldn't it end up with this instead:

```
2024/01/01 Some existing transaction
    Assets:Bank                                 $100
    Income:Salary
2024/01/16 NewPayee
    Expenses:Food                                $10
    Liabilities:CreditCard2024/01/16 NewPayee
    Expenses:Food                                $10
    Liabilities:CreditCard
```

Again, sorry if I'm misunderstanding, but I read the description in passing and the behavior in question struck me as possibly being better addressed in ledger-mode than here.

jwiegley:
Yeah, I'm going to close this as "won't fix". The Emacs mode is where the fix should be made, not Ledger itself.

PR #2685, opened 2026-02-20
Fix missed boost::variant API in times.h begin_has_year()

jwiegley:
## Summary

- The variant migration (PR #2660) converted most uses of `boost::variant` API to `std::variant` in `date_specifier_or_range_t`, but missed the `begin_has_year()` method
- That method still used `.type() == typeid()` and `boost::get<>` instead of `std::holds_alternative<>` and `std::get<>`
- This caused build failures in CI when building the base branch for benchmark comparison (PR #2663 benchmark check)

## Fix

Replace the two boost::variant calls in `begin_has_year()` with their std::variant equivalents.

### DIFF
--- a/src/context.h
+++ b/src/context.h
@@ -117,8 +117,8 @@ inline parse_context_t open_for_reading(const path& pathname, const path& cwd) {
 }
 
 inline path safe_current_path() {
-  boost::system::error_code ec;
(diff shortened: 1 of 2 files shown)


jwiegley:
Superseded by PR #2686 which was merged to master with the same fixes.

PR #2719, opened 2026-02-23
feat: support nested/recursive automated transactions (issue #2102)

### LINKED ISSUE #2102: Allow nested automated transactions
Apologies if this is somehow already possible, but as having just started using ledger (which is a great tool), I have found the below behaviour doesn't work at the moment.

I would like to be able to set multiple automated transactions which will trigger in a nested fashion and being called inside an already automated transaction.

```ldg

= expenses:foo
   assets:foo         -1.0   ; this will never trigger as it's nested

= income:salary
   expenses:bar       $300   ; this will trigger

2022/04/05 Salary
   assets:foo       $1.000
   income:salary
```

The idea behind it would be a way to automatically charge expenses from a specific account without having to manually adding the payment account in the end of a list, like

```ldg
= income:salary
   expenses:foo       $300
   expenses:baz       $300
   expenses:bar       $300
   assets:foo        $-900  ;needs to be manually calculated at the moment
```

jwiegley:
## Summary

Fixes #2102.

Automated transactions currently only match against the original postings in a transaction. A posting generated by one `=` rule (bearing `ITEM_GENERATED`) could never trigger a second `=` rule, even when its account matched that rule's predicate.

This PR implements iterative "wave" processing so that each round of auto-transaction expansion feeds newly generated postings back through the full list of rules, enabling nested automation:

```ledger
= /expenses:gym/
    expenses:whatever                         $10
    assets:bank                              -$10

= /income:salary/
    expenses:gym                              $50
    assets:bank                              -$50

2022-04-04 Salary
    assets:bank                           $1,000
    income:salary
```

Before this fix, `= /expenses:gym/` never fired because `expenses:gym $50` was generated (ITEM_GENERATED). After this fix, `= /income:salary/` fires in wave 0, generating `expenses:gym $50`; then `= /expenses:gym/` fires in wave 1, generating `expenses:whatever $10` and `assets:bank -$10`.

## Changes

- **`src/xact.h` / `src/xact.cc`**: Add optional `posts_to_process` parameter to `auto_xact_t::extend_xact`. When supplied, the provided list is used directly (no ITEM_GENERATED filtering); when absent (first pass), the original behaviour is preserved exactly.

- **`src/journal.cc`**: `journal_t::extend_xact` now drives an iterative wave loop. Each auto transaction tracks which postings it has already processed; subsequent waves supply only the freshly generated, unseen postings. The loop exits when no rule has unseen input, with a hard cap of 20 iterations to guard against pathological mutual-trigger cycles.

- **`test/regress/2102.test`**: Regression test for the nested auto-transaction scenario from the issue.

## Test plan

- [x] New regression test `test/regress/2102.test` passes
- [x] All existing regression tests pass (emacs timestamp tests fail pre-existing, unrelated to this change)
- [x] All existing baseline tests pass
- [x] Python binding unchanged — default parameter means existing two-argument callers still work

### DIFF
--- a/src/journal.cc
+++ b/src/journal.cc
@@ -31,6 +31,8 @@
 
 #include <system.hh>
 
+#include <unordered_set>
+
 #include "journal.h"
 #include "context.h"
 #include "amount.h"
@@ -117,8 +119,7 @@ account_t* journal_t::find_account_re(const string& regexp) {
   return master->find_account_re(regexp);
 }
 
-account_t* journal_t::register_account(string_view name, post_t* post,
-                                       account_t* master_account) {
+account_t* journal_t::register_account(string_view name, post_t* post, account_t* master_account) {
   // If there are any account aliases, substitute before creating an account
   // object.
   account_t* result = expand_aliases(string(name));
@@ -364,8 +365,7 @@ bool journal_t::add_xact(xact_t* xact) {
   // applied to it.
   if (std::optional<value_t> ref = xact->get_tag(_("UUID"))) {
     std::string uuid = ref->to_string();
-    auto [iter, inserted] =
-        checksum_map.insert(checksum_map_t::value_type(uuid, xact));
+    auto [iter, inserted] = checksum_map.insert(checksum_map_t::value_type(uuid, xact));
     if (!inserted) {
       // This UUID has been seen before; apply any postings which the
       // earlier version may have deferred.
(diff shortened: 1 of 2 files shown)


afh:
My first gut reaction is that it might be hard to reason about  where postings come from, one approach to make that easier could be to mark automated xacts that the user wants to be recursive with a key word, e.g.:

```
= nested && /expenses:gym/
    expenses:whatever                         $10
    assets:bank                              -$10

= /income:salary/
    expenses:gym                              $50
    assets:bank                              -$50

2022-04-04 Salary
    assets:bank                           $1,000
    income:salary
```

Another approach could be a command-line flag; by default automated xacts are not recursive, but with `--recursive-transactions` they are.

What do you think, @jwiegley?

jw-pos:
I like your thinking. Right now, in an automated expression, referring to a string surrounded by slashes expands to the following expression: `/name/` is `account =~ /name/`.

We could add a new variable like `any_account` which, if it appears in the expression tree, triggers such a nested lookup…

afh:
How would this look in the initial example given in the PR description?

PR #2772, opened 2026-02-27
Fix print/xact/dump to respect 'not' in positional queries (#1227)

### LINKED ISSUE #1227: print command ignores `not` keyword in queries (BZ#1227)

Bugzilla bug ID: BZ#1227
From: unknown reporter

jwiegley:
## Summary

- `ledger print not :pension` was showing all transactions instead of only those without any pension posting
- Root cause: `filter_posts` evaluates the limit predicate per-posting, so `not X` means "any posting satisfies not-X" (true for nearly all transactions)
- Fix: in `parse_query_args`, when `related_all` is set (enabled by `normalize_options()` for `print`, `xact`, `dump`), rewrite the predicate to use transaction-level semantics:
  - `(! INNER)` → `(! any(INNER))` — lift NOT to transaction level
  - `PRED` → `any(PRED)` — use any-posting semantics

## Test plan

- [x] `print :pension` shows only the transaction with an Assets:Pension posting
- [x] `print not :pension` shows only transactions with no pension posting (was broken before)
- [x] `print payee Pension` shows only the transaction with payee "Pension"
- [x] `print not payee Pension` shows only transactions without that payee (was broken before)
- [x] All existing regression tests pass (2175/2176; one pre-existing failure in `coverage-wave3-precommands` unrelated to this change)

Fixes #1227

### DIFF
--- a/src/report.cc
+++ b/src/report.cc
@@ -350,7 +350,7 @@ namespace {
 ///   - other nodes become O_CALL("any", root)
 ///
 /// This matches the documented workaround: print -l "not any(account =~ /X/)"
-expr_t::ptr_op_t wrap_xact_predicate(expr_t::ptr_op_t op) {
+expr_t::ptr_op_t wrap_xact_predicate(const expr_t::ptr_op_t& op) {
   using op_t = expr_t::op_t;
 
   // Build O_CALL("any", argument) for a given argument subtree.


jwiegley:
I don't quite like how overly specific this implementation is. I think it could be done in a better way.

PR #2797, opened 2026-02-28
Honor --date-format and --datetime-format in prices/pricedb (#1202)

### LINKED ISSUE #1202: date-format option not used with prices and pricedb commands (BZ#1202)

Bugzilla bug ID: BZ#1202
From: unknown reporter

jwiegley:
## Summary

- When `--date-format` or `--datetime-format` is set, all date/datetime value interpolation in format expressions now respects the user's format
- When not set, the default `%Y/%m/%d` format is preserved for backward compatibility
- This is a general fix in `value_t` that applies to all commands, not just `prices`/`pricedb`
- Added regression test `test/regress/1202.test`

## Root Cause

When a date or datetime value is interpolated in a format expression (e.g., `%(date)` or `%(datetime)`), the result was always rendered using `FMT_WRITTEN` (hardcoded `%Y/%m/%d`), ignoring the user's `--date-format` and `--datetime-format` options.

## Fix

Added `explicit_date_format` and `explicit_datetime_format` tracking flags in `times.cc`. When the user sets `--date-format`/`--datetime-format`, these flags are set to true. In `value_t::print()` and `value_t::in_place_cast(STRING)`, the code now checks these flags: if set, uses `FMT_PRINTED` (user-configured format); if not set, uses `FMT_WRITTEN` (default format) for backward compatibility.

## Test plan

- [x] New regression test `test/regress/1202.test` verifies `--date-format` works with `prices`
- [x] New regression test verifies `--datetime-format` works with `pricedb`
- [x] All previously-passing tests continue to pass (pre-existing failures in budget tests are unrelated)

Fixes #1202

### DIFF
--- a/src/times.cc
+++ b/src/times.cc
@@ -1825,16 +1825,27 @@ std::string format_date(const date_t& when, const format_type_t format_type,
 
 namespace {
 bool is_initialized = false;
-}
+bool explicit_date_format = false;
+bool explicit_datetime_format = false;
+} // namespace
 
 void set_datetime_format(const char* format) {
+  explicit_datetime_format = true;
   written_datetime_io->set_format(format);
   printed_datetime_io->set_format(format);
 }
 
 void set_date_format(const char* format) {
+  explicit_date_format = true;
   written_date_io->set_format(format);
   printed_date_io->set_format(format);
+}
+
+bool date_format_is_set() {
+  return explicit_date_format;
+}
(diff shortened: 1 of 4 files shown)


jwiegley:
Instead of just doing this for the prices and pricedb commands, it perhaps should always be the case that interpolating a date as a value uses format date.

PR #2798, opened 2026-02-28
Fix crash in csv -X when display_amount is a multi-commodity balance (issue #1184)

### LINKED ISSUE #1184: csv currency covert fails (BZ#1184)

Bugzilla bug ID: BZ#1184
From: unknown reporter

jwiegley:
## Summary

- Fixes crash in `ledger csv -X EUR` when a synthetic revaluation posting has a multi-commodity balance
- `get_commodity()` in `post.cc` and `fn_commodity()`/`fn_quantity()` in `report.cc` now handle balance values gracefully
- For single-commodity balances, the functions extract the commodity/quantity from the single amount
- For multi-commodity balances, `commodity()` returns an empty string and `quantity()` returns 0

## Root Cause

When using `-X EUR` (or `-V`), the `changed_value_posts` revaluation filter generates synthetic postings to represent unrealized gains/losses. These synthetic postings can have a compound balance value (e.g., `0.1 DK -2 EUR`) when commodity conversion is incomplete. The CSV format calls `commodity(scrub(display_amount))` on these postings, which crashed with "Cannot convert a balance with multiple commodities to an amount" because the `commodity()` and `quantity()` functions expected a single-commodity amount.

The `reg` command was unaffected because its format uses `justify(scrub(display_amount))` which handles balance values natively.

## Test plan

- [x] Existing CSV regression tests pass (13/13)
- [x] New regression test `test/regress/1184.test` added and passes
- [x] No pre-existing test regressions introduced
- [x] Code passes `clang-format` check

Closes #1184

### DIFF
--- a/src/post.cc
+++ b/src/post.cc
@@ -325,6 +325,13 @@ value_t get_use_direct_amount(post_t& post) {
  */
 value_t get_commodity(call_scope_t& args) {
   if (args.has<amount_t>(0)) {
+    value_t val(args[0]);
+    if (val.is_balance()) {
+      const balance_t& bal(val.as_balance());
+      if (bal.single_amount())
+        return bal.amounts.begin()->second.commodity().strip_annotations(keep_details_t{});
+      return string_value(empty_string);
+    }
     return args.get<amount_t>(0).commodity().strip_annotations(keep_details_t{});
   } else {
     post_t& post(args.context<post_t>());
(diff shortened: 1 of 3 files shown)


jwiegley:
There's still a bit too much code duplication in this PR. I think the logic being used here might be better factored out as a new function.

PR #2801, opened 2026-02-28
print: include automated and period transactions in output (fix #1186)

### LINKED ISSUE #1186: ledger print does not output automated postings. (BZ#1186)

Bugzilla bug ID: BZ#1186
From: unknown reporter

jwiegley:
## Summary

Fixes #1186: `ledger print` now outputs automated transaction rules (`= <predicate>`) and period transactions (`~ <period>`) in addition to regular transactions, making the output a complete, self-contained journal suitable for archiving.

**Before:** `ledger print` only output regular transactions, omitting the automated transaction rules. Re-reading the output would produce different results (missing the envelope/budget accounting effects).

**After:** `ledger print` outputs automated and period transaction rules before regular transactions. Re-reading the archived output produces identical reports.

### Design decisions

- **Automated rules are emitted first**: Rules appear before regular transactions, matching the conventional placement in ledger files
- **`--generated` suppresses rules**: When `--generated` materializes auto-generated postings inline in each transaction, the rules are suppressed to avoid double-counting when re-reading
- **`entry`/`xact`/`draft` commands unchanged**: These use a new `include_auto_xacts=false` flag to continue emitting only the single newly-created transaction
- **Account `assert`/`check` auto-xacts excluded**: Auto-xacts synthesized from `account` directives (with no stored predicate text) are excluded, since they are not standalone `=` transaction rules
- **`--raw` mode works**: The raw mode reads from the source file using position info that auto/period xacts already track

### Implementation

1. `textual_xacts.cc`: Store the original query string in `auto_xact_t::predicate` via `set_text()` during parsing, enabling reconstruction without re-serializing the AST
2. `print.cc`: Add `print_auto_xact()` and `print_period_xact()` helper functions; modify `print_xacts::flush()` to emit them before regular transactions
3. `print.h`: Add `include_auto_xacts` flag to `print_xacts` class
4. `draft.cc`: Pass `include_auto_xacts=false` for the `entry`/`xact` command

## Test plan

- [x] New regression test `test/regress/1186.test` verifies `print` includes auto/period transactions and that `--generated` suppresses them
- [x] All 2200 existing tests pass (excluding pre-existing `DocTestsTest_ledger3` failure)
- [x] Updated 4 existing tests that had `print` blocks with auto/period transactions to match new expected output
- [x] Round-trip archiving verified: `ledger print > archive.ledger && ledger -f archive.ledger balance` produces identical output to the original

### DIFF
--- a/src/convert.cc
+++ b/src/convert.cc
@@ -74,7 +74,7 @@ value_t convert_command(call_scope_t& args) {
 
   /*--- CSV parsing and transaction creation ---*/
 
-  print_xacts formatter(report);
+  print_xacts formatter(report, false, false);
   path csv_file_path(args.get<string>(0));
 
   report.session.parsing_context.push(csv_file_path);
--- a/src/print.cc
+++ b/src/print.cc
@@ -242,8 +242,8 @@ void print_xact_base_posts(report_t& report, std::ostream& out, const xact_base_
         amt = amt_str.str();
       } else if (!post->amount.is_null()) {
         std::ostringstream amt_str;
-        value_t(post->amount).print(amt_str, static_cast<int>(amount_width), -1,
-                                    AMOUNT_PRINT_RIGHT_JUSTIFY);
+        value_t(post->amount)
+            .print(amt_str, static_cast<int>(amount_width), -1, AMOUNT_PRINT_RIGHT_JUSTIFY);
         amt = amt_str.str();
       }
 


jwiegley:
I don't agree that automated and period transaction should always be output ledger print, but maybe this logic should apply when the `--raw` option is used.

PR #2805, opened 2026-02-28
Fix use-after-free in Python journal query (issue #2152)

### LINKED ISSUE #2152: python segault in ledger::journal_t::clear_xdata
A user of ledger 3.2.1 in Debian reports a segfault from a python script. You can find a reproducer and backtrace (with symbols) at 

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1024389

It looks related to python GC (well, reference counting).

Let me know if I can provide more information.

jwiegley:
## Summary

- `collector_wrapper` held a raw `journal_t&` reference to the journal, which became dangling if the Python `Journal` object was garbage collected before the `PostCollectorWrapper`
- Changed `collector_wrapper` to store a `boost::shared_ptr<journal_t>`, keeping the journal alive for the full lifetime of the query result  
- Changed `py_query` to accept `boost::shared_ptr<journal_t>` directly (Boost.Python extracts it automatically from the Python object since `journal_t` is registered with `shared_ptr` as holder type)
- Added regression test that explicitly exercises the crash scenario: creates a query, deletes all other journal references, then iterates and destroys the collector

## Test plan

- [x] New regression test `test/regress/2152_py.test` passes
- [x] All existing Python regression tests continue to pass (`852_py`, `975_py`, `976_py`, `977_py`, `978_py`, etc.)
- [x] Build succeeds with `-DUSE_PYTHON=ON`

Closes #2152

### DIFF
--- a/test/regress/2152.py
+++ b/test/regress/2152.py
@@ -1,0 +1,36 @@
+import ledger
+
+# Regression test for GitHub issue #2152:
+# Segfault in journal_t::clear_xdata() when the PostCollectorWrapper
+# (collector_wrapper) outlives the Python Journal object.
+#
+# The crash happened because collector_wrapper held a raw reference
+# (journal_t&) to the journal. If the journal was freed before the
+# collector_wrapper's destructor ran, clear_xdata() would access freed
+# memory.
+#
+# The fix stores a boost::shared_ptr<journal_t> in collector_wrapper,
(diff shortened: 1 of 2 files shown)


jwiegley:
Shouldn't this be using `std::shared_ptr`?

PR #2816, opened 2026-03-02
Fix duplicate commodity quote downloads (issue #996)

### LINKED ISSUE #996: ledger downloads a new quote once for every entry...  sometimes (BZ#996)

Bugzilla bug ID: BZ#996
From: Eevee (@eevee)

jwiegley:
## Summary

Fixes #996 — ledger was downloading duplicate price quotes for the same commodity when it appeared in multiple postings in a single run.

## Root Cause

When `--download` is active, `check_for_updated_price()` is called once per posting per commodity. The function decides whether to re-download by comparing the transaction's `moment` (midnight of the posting date) against the timestamp of the existing price. However, if a price was just downloaded with a later time component on the same calendar date (e.g., `13:49:26` vs `00:00:00`), `find_price()` cannot see it — `upper_bound(midnight)` returns that entry as "too recent" — and instead returns the older stale price. This causes `check_for_updated_price()` to conclude another download is needed, repeating the cycle for every subsequent posting that references the commodity.

## Fix

Add a `mutable std::optional<datetime_t> last_quote` field to `commodity_t::base_t`. Before deciding to fetch a quote, check whether a successful download already occurred within the current leeway window (measured against the real wall-clock time, not the historical `moment`). After a successful fetch, record the current wall-clock time so all subsequent postings for that commodity skip the download.

## Test plan

- [ ] New regression test `test/regress/996.test` passes (3 postings across 2 commodities, with `--download --market`)
- [ ] All existing quote/price-related tests pass (`coverage-quote-*.test`, `coverage-history-*.test`, `coverage-price-*.test`)
- [ ] Code passes clang-format-18

### DIFF
--- a/test/regress/996.test
+++ b/test/regress/996.test
@@ -1,0 +1,29 @@
+; Test that --download does not fetch duplicate quotes for the same commodity
+; when it appears across multiple postings (issue #996).
+;
+; Prior to the fix, find_price() called with a historical moment (midnight)
+; could not see a price already downloaded with a later time component on the
+; same date.  This caused check_for_updated_price() to re-download on every
+; posting that referenced the commodity.  The fix adds last_quote tracking to
+; commodity_t::base_t so that a successful download within the leeway window
+; suppresses subsequent download attempts for that commodity.
+
+2013/10/21 Buy Stock A
+    Assets:Brokerage        10 AAPL
+    Assets:Checking        $-1000.00
+
+2013/10/21 Buy More Stock A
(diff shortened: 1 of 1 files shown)


jw-pos:
The `build (macos-latest, MacOS, 3, -DUSE_PYTHON=ON -DUSE_GPGME=ON)` check failed due to a transient Homebrew network error (HTTP 500 when fetching taps) during the Mac Dependencies step — not a code issue. All 9 other checks pass. Could someone with admin rights re-run the failed job?

PR #2827, opened 2026-03-02
Fix --flat --depth interaction: roll up hidden children

### LINKED ISSUE #1020: Bad interaction of --depth and --flat: accounts missing (BZ#1020)

Bugzilla bug ID: BZ#1020
From: Martin Michlmayr (@tbm)
CC:   @simonmichael, @thdox

jwiegley:
## Summary

Fixes #1020: When using `--flat` together with `--depth N`, accounts at the depth boundary with no direct postings (but with children that have postings) were silently dropped from the output.

**Before** (`bal --flat --depth 2`):
```
10.00 EUR  Assets:AAA
-20.00 EUR  Equity:Opening-balance
```
`Assets:BBB` was missing even though `Assets:BBB:CCC` exists with a 10 EUR balance.

**After** (`bal --flat --depth 2`):
```
10.00 EUR  Assets:AAA
10.00 EUR  Assets:BBB
-20.00 EUR  Equity:Opening-balance
```

## Root Cause

Two interacting bugs:

1. In `mark_accounts` (output.cc): flat mode never considered accounts with visited children but no direct postings, because the condition `!flat && visited > 0` was always false in flat mode.

2. In the `--flat` option (report.h): the `total_` expression `parent ? amount : total` made depth-boundary accounts show zero (their own amount) instead of their family total.

## Fix

- Added `has_children_to_display` account expression function that checks whether any descendant has `ACCOUNT_EXT_TO_DISPLAY` set
- Changed flat mode's `total_` expression to use `total` (family total) when an account has no displayed children — precisely when all children are hidden by the depth filter  
- Added `(flat && visited > 0 && to_display == 0)` condition in `mark_accounts` to consider depth-boundary rollup accounts in flat mode

## Test plan

- [x] New regression test `test/regress/1020.test` passes
- [x] Existing flat/depth baseline test `opt-depth_flat.test` updated and passes
- [x] Existing `854150DF.test` updated and passes  
- [x] All 18 flat/depth related tests pass
- [x] All 2220 tests pass (9 pre-existing unrelated failures remain)

### DIFF
--- a/test/baseline/opt-depth_flat.test
+++ b/test/baseline/opt-depth_flat.test
@@ -19,11 +19,18 @@
     Income:One:Two:Three:Books          $-10.00
 
 test bal --depth 1 --flat
+             $-10.00  Assets
+              $40.00  Expenses
+             $-30.00  Income
+--------------------
+                   0
 end test
 
 test bal --depth 2 --flat
              $-10.00  Assets:Cash
               $10.00  Expenses:Books
+              $30.00  Expenses:One
+             $-30.00  Income:One
 --------------------
                    0
 end test
@@ -32,7 +39,9 @@ test bal --depth 3 --flat
              $-10.00  Assets:Cash
               $10.00  Expenses:Books
               $10.00  Expenses:One:Books
+              $20.00  Expenses:One:Two
              $-10.00  Income:One:Books
+             $-20.00  Income:One:Two
 --------------------
                    0
 end test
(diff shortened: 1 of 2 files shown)


jw-pos:
The only remaining CI failure (`build (macos-latest, MacOS, 3, -DUSE_PYTHON=ON -DUSE_GPGME=ON)`) is a transient GitHub infrastructure issue — the job failed in 1m18s during repository checkout with `HTTP 500: The requested URL returned error: 500` errors, not due to any code change. All other 9 checks pass: Ubuntu builds, standard macOS build, sanitizer build, lint-check, format-check, coverage, benchmark, and Python analysis.

PR #2879, opened 2026-03-10
Fix #868: Show payee when it changes between sorted postings

### LINKED ISSUE #868: Wrong payee depending on the sort order (BZ#868)

jwiegley:
## Summary
- When `--sort` reorders postings from different transactions, the register format only showed the payee on the first line of each transaction group, not when the payee actually changed between consecutive output lines
- Root cause: the next_lines_format used `has_tag("Payee") ? payee : " "` which only checked for an explicit Payee metadata tag, ignoring payee changes caused by sorting
- Fix: track the last displayed payee in `report_t::last_displayed_payee` and expose it as `last_payee` in format expressions. The next_lines_format now uses `payee != last_payee ? payee : " "` to show the payee whenever it differs from the previous line
- Added `xact_payee` property to `post_t` for accessing the raw transaction payee
- Updated 3 existing tests with corrected payee suppression expectations
- Added regression test `test/regress/868.test`

## Test plan
- [x] New regression test verifies payee display with `--sort payee`
- [x] Existing tests updated for improved payee suppression behavior
- [x] All 3996+ tests pass (cmake + ctest)
- [x] nix build passes

Fixes #868

### DIFF
--- a/src/output.cc
+++ b/src/output.cc
@@ -149,6 +149,7 @@ void format_posts::operator()(post_t& post) {
     // NOLINTEND(bugprone-branch-clone)
 
     post.xdata().add_flags(POST_EXT_DISPLAYED);
+    report.last_displayed_payee = post.payee();
     last_post = &post;
   }
 }
--- a/src/post.cc
+++ b/src/post.cc
@@ -263,6 +263,10 @@ value_t get_payee(post_t& post) {
   return string_value(post.payee());
 }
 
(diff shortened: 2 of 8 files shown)


jwiegley:
This PR has no effect.

PR #2957, opened 2026-03-22
Fix --forecast handling of 'from' and 'until' dates

### LINKED ISSUE #1113: 'from' and 'until' not handled correctly by --forecast (BZ#1113)

Bugzilla bug ID: BZ#1113
From: Joakim Reinert (@jreinert)

jwiegley:
## Summary

- Fix `forecast_posts::add_post` to correctly handle periodic transactions with a future `from` date (previously silently dropped) and a past `until` date (previously could error)
- Implement previously-declared-but-unimplemented `date_interval_t::add_duration` and `subtract_duration` static helpers
- Add regression test covering both cases

## Details

The root cause was that `add_post` called `find_period(CURRENT_DATE())` and immediately returned if it failed. For a future `from` date, `find_period` correctly reports that the current date is not within any period — but the interval is still valid for forecasting. The fix separates stabilization from validation: after stabilizing, it checks whether the interval starts in the future and backs up one period so that `flush()` emits the first posting on the `from` date.

For expired intervals (`until` in the past), the fix adds an explicit check that drops them gracefully rather than relying on `find_period`'s return value.

## Test plan

- [x] New regression test `test/regress/1113.test` with two cases: future `from` and past `until`
- [x] All 15 existing forecast tests pass
- [x] Full test suite (4063 tests) passes
- [x] Nix flake build succeeds

Fixes #1113

### DIFF
--- a/src/filters.cc
+++ b/src/filters.cc
@@ -1840,14 +1840,34 @@ void budget_posts::flush() {
  */
 void forecast_posts::add_post(const date_interval_t& period, post_t& post) {
   date_interval_t i(period);
-  if (!i.start && !i.find_period(CURRENT_DATE()))
+  if (!i.start)
+    i.find_period(CURRENT_DATE());
+
+  if (!i.start)
     return;
+
+  // Drop intervals whose finish date is entirely in the past.
+  if (i.finish && CURRENT_DATE() > *i.finish)
+    return;
+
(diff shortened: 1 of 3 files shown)


jwiegley:
Superseded by #2965 which consolidates all 9 forecast PRs into a single cohesive overhaul with comprehensive testing.

PR #2958, opened 2026-03-22
Include current period in forecast output

### LINKED ISSUE #1576: Forecast doesn't include current month's budget (BZ#504)

Bugzilla bug ID: BZ#504
From: @spiffytech

jwiegley:
## Summary

- Fix `forecast_posts::add_post()` to advance the stored interval (not a local copy) past fully-elapsed periods, keeping the period containing today
- Fix `forecast_posts::flush()` to date forecast transactions at `start` (current period) instead of `next` (next period)
- Together these two bugs caused every forecast to skip the current month's budget

## Details

`forecast_posts` had two interacting bugs causing the current period to be omitted from forecast output:

1. **`add_post()`** created a local copy of the date interval, stored it in `pending_posts`, then advanced the local copy past `CURRENT_DATE()` — but the stored entry was never updated. The fix advances the stored entry directly and uses end-of-period (`next <= CURRENT_DATE()`) as the boundary so the period containing today is preserved.

2. **`flush()`** dated each synthetic forecast transaction at `next` (the start of the *next* period) rather than `start` (the start of the *current* period), effectively skipping one period forward.

## Test plan

- [x] New regression test `test/regress/1576.test` verifies the current month is included
- [x] All 17 affected forecast test baselines updated to expect the current period
- [x] Full test suite passes (4064/4064 tests)
- [x] Nix flake build passes

Fixes #1576


jwiegley:
Superseded by #2965 which consolidates all 9 forecast PRs into a single cohesive overhaul with comprehensive testing.

PR #3087, opened 2026-03-27
Sort by converted value when -X is active (#1556)

### LINKED ISSUE #1556: Sorting problem concerning amounts and -X (BZ#484)

Bugzilla bug ID: BZ#484
From: John Wiegley (@jwiegley)
CC:   @tbm

jwiegley:
## Summary

- Fix `-S amount` sorting when `-X` (exchange) is active to sort by converted value rather than raw quantity
- Move confirmation test from `test/todo/` to `test/regress/` with corrected expected output
- Add register test case alongside the balance test

Fixes #1556

## Details

When `-X <commodity>` is used with `-S amount`, postings/accounts in different commodities were sorted by their raw quantities rather than their converted values. For example, `100 EUR` and `50 GBP` would sort by 100 vs 50 instead of by their dollar equivalents ($120 vs $75).

The root cause: `-X` rewrites `display_amount_` and `display_total_` expressions to apply commodity conversion, but the `amount` sort expression evaluates the raw posting amount before conversion. The sort comparator in `compare_items<T>::operator()` cached these unconverted values.

The fix adds `convert_sort_values_for_exchange()` in `src/compare.cc` which, after sort values are evaluated and before they are cached, converts any amount/balance sort values to the exchange commodity using `value_t::exchange_commodities()`. This is called for both `post_t` and `account_t` comparisons.

## Test plan

- [x] Regression test `test/regress/1556.test` verifies correct sort order for both `bal` and `reg` with `-X '$' -S amount`
- [x] Full test suite (4180 tests) passes
- [ ] CI passes on all platforms

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/compare.cc
+++ b/src/compare.cc
@@ -84,6 +84,25 @@ public:
 
 } // anonymous namespace
 
+/*--- Exchange conversion for sort values ---*/
+
+/**
+ * When -X (exchange) is active, convert any amount or balance sort
+ * values to the target commodity so that cross-commodity comparisons
+ * sort by converted value rather than raw quantity.
+ */
+static void convert_sort_values_for_exchange(std::list<sort_value_t>& sort_values,
+                                             report_t& report) {
+  if (report.HANDLED(exchange_)) {
+    string comm = report.HANDLER(exchange_).str();
+    for (auto& sv : sort_values) {
+      if (sv.value.is_amount() || sv.value.is_balance()) {
+        sv.value = sv.value.exchange_commodities(comm, false, datetime_t());
+      }
+    }
+  }
+}
+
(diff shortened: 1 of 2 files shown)


jwiegley:
* **#3087** <a href="https://app.graphite.com/github/pr/ledger/ledger/3087?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ledger/ledger/3087?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3140, opened 2026-04-23
Fix --only to apply after calc_posts so running totals are unaffected (#762)

### LINKED ISSUE #762: --only doesn't work: does the same as --limit (BZ#762)

Bugzilla bug ID: BZ#762
From: Martin Michlmayr (@tbm)

jwiegley:
## Summary

Fixes #762.

The `--only` filter was placed upstream of `calc_posts` in the posting handler chain. Filtering before the running-total stage meant that `--only` affected the accumulated total in exactly the same way as `--limit`, and for simple register reports the two options produced identical output.

This contradicts the documented intent of `--only` as a *secondary narrowing* predicate that should restrict what is *displayed* without changing what is *calculated*.

## What changed

`--only` is now placed **downstream** of `calc_posts` — and downstream of revaluation and the `display_filter_posts` rounding-adjustment accumulator — so that:

- `--limit` excludes postings from calculation entirely (running total reflects only matching postings)
- `--only` keeps every posting in the running total but suppresses non-matching ones from the displayed output (matching `--display` behavior)
- `--display` continues to filter at the display-amount stage

For accounts reports there is no `display_filter_posts` in the chain, so a separate (mutually-exclusive) `--only` block is added immediately after `calc_posts` to keep the same semantics.

## Tests

- New regression test `test/regress/762.test` (promoted from `test/todo/762.test` with corrected expectations).
- Updated `test/baseline/opt-only.test` and several coverage tests (`cov2-more-opts3`, `cov2-query-clauses`, `coverage-query-bold-multi`, `coverage-report-new-4`, `exhaustive-query-sections`) whose expected output documented the old buggy behavior.
- Full ctest suite passes (4266 tests).

## Test plan

- [x] `python3 test/RegressTests.py --ledger ./build/ledger --sourcepath . test/regress/762.test`
- [x] `cd build && ctest -j8` (4266/4266 pass)
- [x] Reproduce the BZ#762 case manually and confirm that `--only` now keeps the running total stable across hidden postings, while `--limit` continues to alter it.


jwiegley:
* **#3140** <a href="https://app.graphite.com/github/pr/ledger/ledger/3140?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ledger/ledger/3140?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3168, opened 2026-04-24
Accept a time component for --begin/--end (#1696)

### LINKED ISSUE #1696: --end seems not to accept time
Hi,
option --end seems not to accept time in combination with date. 
I would expect to use the option like --end 2018/12/16 08:12:33 or --end 2018/12/16@08:12:33

jwiegley:
## Summary

- `--begin`, `--end`, and `--now` now accept an argument that carries a time component, so intra-day boundaries can be expressed on the command line (fixes #1696).
- Both `"2018/12/16 08:12:33"` and `2018/12/16@08:12:33` are recognized; the `@` form avoids shell quoting.
- When a time is present the option emits a `datetime`-based limit instead of the usual `date` limit, so timelog sessions and sub-daily price records are filtered precisely.
- Whole-day arguments continue to flow through `date_interval_t`, so there are no behavioural changes for existing usage.
- For `--end`, the report terminus is pulled back by one second so that market-price lookups use the last included instant rather than the first excluded one, mirroring the existing day-of behaviour.

## Changes

1. `times`: `'@'` is normalized to space inside `parse_datetime`, and a non-throwing `try_parse_datetime` is added for option parsers.
2. `report.h`: `--begin`, `--end`, `--now` try the datetime form first and fall back to the existing date-interval code path.
3. `doc/ledger3.texi`: documents the new accepted forms.
4. `test/regress/1696.test`: regression coverage for the space- and @-separated forms, combined use, boundary inclusion, and bare-date fallback.

## Test plan

- [x] `ctest -R RegressTest_1696` passes
- [x] Full `ctest` suite passes (the pre-existing `LedgerUtilTests`/`LedgerMathTests` "Not Run" entries reproduce on `main` without these changes)
- [x] Manual check: `ledger -f timelog.dat reg --end "2018/12/15 12:30:00"` and the `@` form both trim mid-day as expected; `-V` with a sub-daily price DB picks the correct price

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -6571,11 +6571,16 @@ Display only transactions on or before the current date.
 
 @item --begin @var{DATE}
 @itemx -b @var{DATE}
-Limit the processing to transactions on or after @var{DATE}.
+Limit the processing to transactions on or after @var{DATE}.  @var{DATE}
+may include a time component, e.g.@: @code{"2018/12/16 08:12:33"} or
+@code{2018/12/16@@08:12:33} (the @code{@@} separator avoids shell
+quoting); the filter then matches by full datetime instead of by whole
+day, which is useful with timelog entries or sub-daily price records.
 
 @item --end @var{DATE}
 @itemx -e @var{DATE}
-Limit the processing to transactions before @var{DATE}.
+Limit the processing to transactions before @var{DATE}.  A time
+component is accepted with the same forms described for @option{--begin}.
 
 @item --period @var{PERIOD_EXPRESSION}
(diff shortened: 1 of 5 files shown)


jwiegley:
* **#3168** <a href="https://app.graphite.com/github/pr/ledger/ledger/3168?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ledger/ledger/3168?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Issue #717, opened 2012-03-19
Error: Cannot convert a balance with multiple commodities to an amount (BZ#717)


tbm:
Note: the issue was created automatically with bugzilla2github

Bugzilla bug ID: BZ#717
From: Martin Michlmayr (@tbm)

PR #3171, opened 2026-04-24
Fix #1692: output breaks accounting equation for sub-precision inferred amounts

### LINKED ISSUE #1692: Output breaks accounting equation
I'm sure ledger stores things properly internally, but basically we're breaking the double entry in the output here because Y in Assets:B is < 0.5:

```
$ ledger -f d bal      
                 2 X  Assets:A
```

Test case:
```
2018-01-01 * Foo
    Assets:A                    2 X @ 0.2 Y
    Assets:B
```

This makes it work:

```
D 10.00 X
D 10.00 Y
```

jwiegley:
## Summary

- Fixes [#1692](https://github.com/ledger/ledger/issues/1692): inferred
  balancing amounts whose magnitude was smaller than their commodity's
  default display precision were silently dropped from balance and
  register output, breaking double-entry accounting.
- Teaches `amount_t::is_zero()` to honour `BIGINT_COST_PREC`
  symmetrically with `amount_t::print()`, so an inferred `-0.4 Y`
  whose commodity precision is 0 but whose recovered cost precision is
  1 is no longer reported as zero.
- Clears `BIGINT_COST_PREC` in `operator+=` / `operator-=` so the flag
  cannot leak into running-total residuals (which would otherwise
  resurrect as ghost rounding adjustments, see #717).
- Promotes `test/todo/1692.test` to `test/regress/1692.test` with
  expected post-fix output covering `bal`, `reg`, and `print`.

## Test plan

- [x] `ctest -j 2` — 4280/4280 tests pass.
- [x] `python3 test/RegressTests.py --ledger ./build/ledger --sourcepath . test/regress/1692.test` — all 3 sub-tests pass.
- [x] `test/regress/717.test` still passes (confirms the arithmetic
      flag-clearing does not regress the historical-pricing rounding fix).


jwiegley:
* **#3171** <a href="https://app.graphite.com/github/pr/ledger/ledger/3171?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ledger/ledger/3171?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3184, opened 2026-04-26
Show parent header when children zero out (issue #560)

### LINKED ISSUE #560: Getting balances of parent and empty/zero accounts
Hi there,

I recently encountered with behavior discussed here:

https://groups.google.com/forum/#!topic/ledger-cli/86QkKpMrluU

As far as I understand, it was considered a bug. Are there plans to fix it?

I can confirm this behavior for the latest (gd7bc4bab) commit.

jwiegley:
Show parent header when children zero out (issue #560)

When a parent account had multiple visited children but only one of them
had a non-zero balance, the balance report silently dropped the parent
header and collapsed the single-path output directly to the non-zero
leaf. This made it look as if the zero-balance siblings had never
existed, even though their postings were still counted in the totals.

`mark_accounts()` in src/output.cc already displayed a parent when two
or more children were displayable, but treated the "one displayable
child, multiple visited siblings" case the same as single-child
collapse. Add a third clause so a visited count greater than one
(more than one sibling posted to the parent) keeps the parent visible,
matching the behavior users already got with --empty.

Closes #560

Update tests and manual for new parent-header rule

The fix for #560 makes parent accounts appear in the balance report
whenever multiple children were visited, even if only one child has
a non-zero balance. Existing tests encoded the pre-fix output, where
such parents were collapsed into the non-zero leaf; update their
expected output to show the parent header and single visible child.

Also refreshes three drewr3.dat balance examples and one --master-account
example in doc/ledger3.texi, since they hit the same condition with the
shared sample journal.


jwiegley:
* **#3184** <a href="https://app.graphite.com/github/pr/ledger/ledger/3184?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ledger/ledger/3184?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3203, opened 2026-05-11
Fix #3202: skip running-total filters for listing commands

### LINKED ISSUE #3202: Ledger is slower than I'd like
Can we (and with "we" I mean you, John :) ) take advantage of Claude's generosity to improve the performance of ledger? I remember back in the 2.x days ledger was blazing fast, but 3.x has always been slightly sluggish. Not enough for me to complain, but enough to resent it some times.

In particular, I am a bit unhappy with the performance of what I would expect should be close to instant even in giant ledgers: the commands `accounts`, `payees`, `tags`, and `commodities`.

For example, with my current ledger of ~6 years:

```
$ time ledger stats
Time period: 21-Dec-29 to 26-May-04 (1588 days)

  Files these postings came from:
    <elided>/ledge-4.2021
    <elided>/y/2021.ledger
    <elided>/y/2022.ledger
    <elided>/y/2023.ledger
    <elided>/y/2024.ledger
    <elided>/y/2025.ledger

  Unique payees:             969
  Unique accounts:           322

  Number of postings:      19091 (12 per day)
  Uncleared postings:        171

  Days since last post:        5
  Posts in last 7 days:       18
  Posts in last 30 days:     133
  Posts seen this month:      24

real	0m0.922s
user	0m0.861s
sys	0m0.064s

$ time ledger accounts
<...elided...accounts...>

real    0m5.409s
user    0m4.751s
sys     0m0.649s

$ time ledger payees
<...elided...payees...>

real    0m5.118s
user    0m4.519s
sys     0m0.582s

$ time ledger tags
<...elided tags...>

real    0m5.195s
user    0m4.543s
sys     0m0.634s

$ time ledger commodities
<...elided...commodities...>

real    0m5.145s
user    0m4.559s
sys     0m0.573s
```

So that's 5 secs for less than 20k posts. `stats` is actually very fast which is great to see, I would've hoped that if it takes less than 1 sec to count all the payees and accounts that it would've been maybe just marginally slower to print them all, not 5-6 times slower. I feel like something funky is going on there.

And of course don't get me started on `ledger xml` which is a true pain:
```
$ timeout -v 5m sh -c 'ledger xml > /dev/null'
timeout: sending signal TERM to command 'sh'
```

I have no idea how long it would take for it to finish on 20k posts but thankfully I have no need for that (probably because whenever I've had the idea of using it I end up noticing that's unusable).

Also, I am not the only one who has noticed this. I found this (probably LLM-authored 🤔, taking with a heap of salt) post from the beancount forums: <https://beancount.io/forum/t/plain-text-accounting-showdown-2025-beancount-v3-hledger-or-ledger/42> which paints ledger in a very bad light performance wise.

I know that performance work is potentially painful and time consuming, so is there any way in which we could leverage our AI overlords to make ledger fast again? Pretty please :)

jwiegley:
## Summary

- Route the `accounts`, `payees`, `tags`, and `commodities` commands through `posts_report_quick()` (the same fast path `stats` uses), skipping `calc_posts()` / `display_filter_posts()` and the rest of `chain_post_handlers()` that listing commands never read from.
- Wire `post_splitter` into the quick path so `--group-by` continues to work on these commands.
- Add `test/regress/3202.test` covering plain listings, `--limit`, `--period`, `--count`, and `--group-by` over the fast path.

## Why

The listing handlers only need `post.account` / `payee` / metadata / `commodity` — running totals, rounding adjustments, sort, collapse, and interval grouping never fire on their output. Putting them on the same chain as `register` and `balance` made them 6-10× slower than `stats` on a 20K-posting journal, even though `stats` already visits every posting (issue #3202).

Benchmark on a 20K-posting journal with prices and lot annotations:

| Command       | Before  | After   | Speedup |
|---------------|---------|---------|---------|
| `stats`       | 0.04s   | 0.04s   | 1×      |
| `accounts`    | 0.32s   | 0.05s   | 6.4×    |
| `payees`      | 0.30s   | 0.03s   | 10×     |
| `tags`        | 0.30s   | 0.03s   | 10×     |
| `commodities` | 0.30s   | 0.03s   | 10×     |

`--limit`, `--period`, `--begin`/`--end`, `--pivot`, `--anon`, `--budget`, and `--forecast-while` are unaffected — they all live in `chain_pre_post_handlers()`, which the fast path still runs.

## Test plan

- [x] `ctest -j8` — all 4317 tests pass
- [x] New `test/regress/3202.test` (13 cases) exercises plain listings, `--limit` with a commodity predicate, `-b` and `-p` date filters, `--count`, and `--group-by` on the fast path
- [x] `coverage-wave11-output-clear-{commodities,tags,report-accounts}.test` continue to pass (these exercise `--group-by` with the listing commands)
- [x] Manual: `accounts`, `payees`, `tags`, `commodities` with `--limit`, `--period`, `--group-by` on `test/input/drewr3.dat` produce the same output as before

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -178,6 +178,14 @@ components with their standard-library equivalents.
   a stripped running total, and early-returning from
   `balance_t::strip_annotations` when nothing carries annotations
 
+- Route the `accounts`, `payees`, `tags`, and `commodities` listing commands
+  through the `stats` fast path (#3202).  These commands only iterate the
+  posting stream to collect distinct names; previously they paid for the
+  running-total and rounding filters used by `register` and `balance`, which
+  on a 20K-posting journal made them 6-10× slower than `stats`.  The fast
+  path still honors `--limit`, `--period`, `--begin`/`--end`, `--pivot`,
+  `--anon`, `--budget`, `--forecast-while`, and `--group-by`
+
 - Align multi-commodity balance rows with wide decimals (#1795); fix
   multi-commodity balance formatting width (#698); make separator lines adapt
   to `amount_width`; show the payee when it changes between sorted postings;
--- a/src/report.cc
+++ b/src/report.cc
@@ -597,12 +597,18 @@ void report_t::posts_report(post_handler_ptr handler, post_handler_ptr& saved_ch
 }
 
 void report_t::posts_report_quick(post_handler_ptr handler) {
+  if (HANDLED(group_by_)) {
+    unique_ptr<post_splitter> splitter(new post_splitter(handler, *this, HANDLER(group_by_).expr));
+    splitter->set_postflush_func(posts_flusher(handler, *this));
+    handler = post_handler_ptr(splitter.release());
+  }
   handler = chain_pre_post_handlers(handler, *this);
 
   journal_posts_iterator walker(*session.journal.get());
   pass_down_posts<journal_posts_iterator>(handler, walker); // NOLINT(bugprone-unused-raii)
 
-  posts_flusher(handler, *this)(value_t());
+  if (!HANDLED(group_by_))
+    posts_flusher(handler, *this)(value_t());
 }
 
 void report_t::generate_report(post_handler_ptr handler) {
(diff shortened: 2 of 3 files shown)


jw-pos:
Fixes #3202

PR #3204, opened 2026-05-12
Speed up pricedb/prices reports by replacing O(N^2) dedup scan

kcwu:
create_price_xact groups synthetic posts under one xact per price
commodity, so a single popular target currency (e.g. the user's
reporting currency) accumulates every parsed price.  The duplicate
check linearly rescanned xact->posts on every insertion, and computed
`date.date()` -- a Boost ptime->gregorian::date conversion -- on each
inner iteration.  On a price-db with ~113K entries this dominated
runtime (gprof: 99% of CPU in from_day_number / counted_time_rep::date).

Maintain a parallel `seen_by_commodity` index keyed by (symbol, date)
holding the small list of amounts already inserted, and hoist the
ptime->date conversion out of the hot path.  Dedup becomes O(log N).

`pricedb` on a 113K-line price-db: 3.13s -> 0.71s (~4.4x).  All 4312
ctest cases pass; output is bit-identical.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/iterators.cc
+++ b/src/iterators.cc
@@ -108,6 +108,14 @@ struct create_price_xact {
 
   std::map<string, xact_t*> xacts_by_commodity; ///< One xact per commodity symbol.
 
+  // Per xact, an index of already-inserted (date, amount) pairs, used to
+  // deduplicate in O(log) instead of a linear scan of xact->posts.  A single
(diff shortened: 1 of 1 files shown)


kcwu:
well, i don't fully understand the code. feel free replace it with your own version.

PR #3205, opened 2026-05-12
Share position_t::pathname across items parsed from one file

kcwu:
Each parsed transaction and posting carried its own copy of the source
file's path inside its position_t, costing roughly 3 M heap allocations
on a 1 M-xact journal (perf attributed 0.5% of cycles to
std::filesystem::path::_List::_List(copy) alone, with much more spent
in the cascading allocator traffic and resulting page faults).

Change position_t::pathname to a std::shared_ptr<const path> and have
parse_context_t hold a single pathname_ref allocated when the file is
opened.  Every item parsed from that file now refcount-bumps the same
shared path instead of cloning it.  A tiny static empty-path sentinel
ensures the pointer is never null, so existing dereferences stay
straightforward.

Benchmark on a 1 M-xact / 108 MB journal:

  balance:  9.0s ->  7.4s wall (-18%), 3.65 GB -> 2.21 GB RSS (-40%)
  stats:    8.0s ->  6.6s wall (-18%)
  print:   19.7s -> 17.9s wall (-9%)

The bulk of the wall-time win comes from the cascade: ~3 M fewer small
allocations slashes glibc allocator work (operator delete -120 ms,
_int_malloc -85 ms) and kernel page-fault handling (sys time
2.0s -> 1.3s).  All 4312 ctest cases pass.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/account.cc
+++ b/src/account.cc
@@ -922,7 +922,7 @@ void account_t::xdata_t::details_t::update(post_t& post, bool gather_all) {
     posts_virtuals_count++;
 
   if (gather_all && post.pos)
-    filenames.insert(post.pos->pathname);
+    filenames.insert(*post.pos->pathname);
 
   date_t date = post.date();
   date_t today = CURRENT_DATE();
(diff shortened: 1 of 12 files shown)


jwiegley:
Great idea!

PR #3208, opened 2026-05-14
Fix #3206: treat balanced (...) inside bare patterns as regex group

### LINKED ISSUE #3206: register query language documentation
I expect `ledger reg Invest:(Stock|Bond)` will match `Invest:Stock` and `Invest:Bond`. However, it is not. It takes me a long time to figure out the correct command is `ledger reg /Invest:(Stock|Bond)/`.

```
$ ledger query 'Invest:(Stock|Bond)'
--- Input arguments ---
("Invest:(Stock|Bond)")

...
--- Input expression ---
((account =~ /Invest:/) | ((account =~ /Stock/) | (account =~ /Bond/)))
```
Is it intended behavior?

BTW, I cannot find the description of the command line query language in the document. There are short doc here https://github.com/ledger/ledger/blob/main/doc/ledger3.texi#L8429 but they are commented out.

jwiegley:
## Summary

- `ledger reg Invest:(Stock|Bond)` no longer parses as three OR-ed
  terms (`Invest:`, `Stock`, `Bond`).  The lexer now treats a
  parenthesized group embedded in an identifier as a regex grouping
  character, so the whole argument becomes a single account regex.
- The fix is gated three ways to avoid regressions: the prefix must
  not be a query keyword (so `tag(name)` / `payee(...)` still parse
  as before), the matching `)` must lie within the same argument (so
  a typo still surfaces as ``Missing ')'`` rather than a regex
  compile error), and the preceding char must not be whitespace (so
  `food and (bar)` still means `food OR (bar)`).
- Adds `test/regress/3206.test` covering the bare-word, slash-delimited,
  leading-paren-subexpression, unbalanced-paren, nested-group, and
  automated-transaction cases.
- Uncomments and rewrites the previously hidden ``Search terms``
  section of ``doc/ledger3.texi``, including a worked example of the
  new behaviour.

Fixes #3206.

## Test plan

- [x] ``ctest`` passes all 4319 tests (including the new
      ``RegressTest_3206`` and the previously failing-then-fixed
      ``RegressTest_842`` which exercises ``tag("Category") == "food"``)
- [x] Manual checks: ``ledger reg 'Invest:(Stock|Bond)'`` matches
      Stock and Bond but not Mutual; the slash-delimited form gives
      the same result; ``(Stock|Bond)`` (leading paren) still parses
      as a sub-expression; ``Invest:(Stock`` still reports
      ``Missing ')'``
- [x] ``makeinfo --no-validate doc/ledger3.texi`` compiles without
      warnings about the new ``Search terms`` node

### DIFF
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -6921,6 +6921,7 @@ using FIFO or LIFO order.  @var{POLICY} must be @samp{fifo} or
 * Report Options::
 * Basic options::
 * Report filtering::
+* Search terms::
 * Output customization::
 * Commodity reporting::
 * Environment variables::
@@ -8300,7 +8301,7 @@ relate to.
 
 @end ftable
 
-@node Report filtering, Output customization, Basic options, Detailed Option Description
+@node Report filtering, Search terms, Basic options, Detailed Option Description
 @subsection Report filtering
 
 These options change which postings affect the outcome of a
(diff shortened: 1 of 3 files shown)


doc/ledger3.texi
```diff
@@ -8421,51 +8422,147 @@ Set the value expression used for the ``totals'' column in the
...
-@c ledger reg food and not dining and not payee chang
-@c ledger reg food not dining expr 'payee =~ /chang/'
-@c @end smallexample
+@item @code{/REGEX/}
+@tab Like a bare word, but explicitly delimited; useful when the
+pattern contains characters that the shorthand grammar would
+otherwise treat as operators.  Single quotes (@code{'REGEX'}) and
+double quotes (@code{"REGEX"}) work the same way.
+
+@item @code{prefix(alt|alt)}
```
kcwu, line 8451: I thought this is just a special case of regex `word` and unnecessary to be mentioned here?

PR #3209, opened 2026-05-14
Fix #3207: balance --group-by no output when combined with --payee/--account/--date

### LINKED ISSUE #3207: "balance --group-by with --payee doesn't work well
```
2026-01-01 buy
    Food  100 USD
    Cash  -100 USD
```
`ledger -f foo.ledger balance --payee payee --group-by account ` output empty.

p.s. I know `--payee` doesn't make sense here. I set `--payee` in `.ledgerrc` for my custom reporting. I expect that given `balance --payee payee` works well, `balance --payee payee --group-by account` should be okay.

jwiegley:
## Summary

`ledger balance --group-by account` emitted empty output whenever any of
`--payee`, `--account`, or `--date` was also active (a common case when
one of those options is set in `.ledgerrc`).

## Root cause

`post_splitter::flush()` called `post_chain->clear()` before invoking the
postflush callback.  The `clear()` cascade reaches
`temporaries_t::clear()` (via `transfer_details::clear()`), which detaches
all temporary postings it created from their (non-temp) accounts.  By
the time `accounts_flusher` walked the master account tree,
`account_t::amount()` saw no posts with `POST_EXT_VISITED` and returned
zero, so `display_value()` was falsy and no rows were rendered.

## Fix

Two coupled changes in `src/`:

1. `filters.cc` — run the postflush callback **before** `post_chain->clear()`
   so the flusher observes temp postings while they're still attached.
2. `output.cc` — clear `format_accounts::posted_accounts` at the end of
   each `flush()`.  With the new order, the queued accounts include
   temp parent accounts (e.g. created by `--account` or `--pivot`) that
   the subsequent `post_chain->clear()` destroys; without the clear,
   the next group iteration would dereference dangling pointers.

## Test plan
- [x] New regression test `test/regress/3207.test` covering `--payee`,
      `--account`, `--date`, `--group-by-cumulative`, and `reg` variants.
- [x] Full suite passes (`ctest`, 4353 tests).
- [x] Manual verification of issue reporter's exact scenario.

### DIFF
--- a/src/filters.cc
+++ b/src/filters.cc
@@ -74,6 +74,13 @@ void post_splitter::print_title(const value_t& val) {
  * forwarded through the post_chain, which is flushed and optionally cleared
  * (unless --group-by-cumulative is active, which preserves running totals
  * across groups).
+ *
+ * The postflush callback runs before post_chain->clear() so that an
+ * accounts-report flusher can walk the account tree while any temporary
+ * postings created by transfer_details (--payee, --account, --date) are
+ * still attached to their accounts.  Calling clear() first would remove
+ * those temps via temporaries_t::clear(), leaving the account totals
+ * at zero for the group (issue #3207).
  */
 void post_splitter::flush() {
   bool cumulative = report.HANDLED(group_by_cumulative);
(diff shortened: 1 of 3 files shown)


jwiegley:
Fixes #3207

Issue #3187, opened 2026-04-28
Unexpected Rounding When Reporting Cost Basis of Commodity Following Commodity Swap


eacrawford02:
Consider the following Ledger file containing an arbitrary transaction along with a commodity swap over a common base currency:

```
2026-04-20 * Arbitrary
    Foo		  $0.5
    Bar

2026-04-21 * Swap
    Assets	  -10 BTC @ $10.05
    Assets	  20 ETH @ $5.025
```

We can report the cost basis of either commodity with the following command, which, as expected, returns a value of $100.5:

```
# ledger -f bug.dat --limit "commodity == 'ETH'" --basis balance Assets
              $100.5  Assets
```

Now suppose we were to comment out the unrelated transaction and run the same command:

```
;2026-04-20 * Arbitrary
;    Foo		  $0.5
;    Bar

2026-04-21 * Swap
    Assets	  -10 BTC @ $10.05
    Assets	  20 ETH @ $5.025

```

```
# ledger -f bug.dat --limit "commodity == 'ETH'" --basis balance Assets
                $100  Assets
```

Ledger now (erroneously?) rounds the basis figure to the nearest dollar figure rather than reporting the complete fractional value, which I presume to be unexpected program behaviour. For whatever reason, there must be another posting in the ledger file with a fractional value in order for the program to report the unrounded figure--swapping `$0.5` for `$1` will instead yield the undesired rounded value. Do note that a posting within the commodity swap transaction satisfies this condition, such as in the case of a capital loss:

```
; Ledger will correctly report the cost basis of ETH to be $40.5
2026-04-21 * Swap
    Assets	  -10 BTC @ $10.05
    Assets	  20 ETH @ $2.025
    Loss	  $60.00
```

I've found a workaround for this issue, which is to supply the `--unround` option when invoking `ledger`, although I'm unsure if this may cause any unintended consequences.

This issue was observed in version 3.3.2-20230330 of Ledger.

tbm:
Maybe unexpected but this has always been the behaviour: prices (`@ $2.025`) do not influence precision.

You can add this at the beginning of your file:

```
D $1000.00
```

@jwiegley I'm not sure sure I fully understand your proposed commit, but imho the question is whether the docs can be clarified (rather than comments in the source code)

tbm:
Oh, ignore my comment. I only looked at the second commit, not the first one. So you're changing this behaviour.

Issue #781, opened 2012-07-12
Setting precision on the command line (BZ#781)


tbm:
Note: the issue was created automatically with bugzilla2github

Bugzilla bug ID: BZ#781
From: Martin Michlmayr (@tbm)
CC:   @tripun

PR #3221, opened 2026-05-26
Fix #3220: round cost basis once at display, not per posting

### LINKED ISSUE #3220: Major ledger rounding changes in main
This was a discussion on the mailing list, just creating an issue so it doesn't get lost.

In the previous release, ledger always maintained maximum possible precision then rounded once at the end. In the next release, ledger will round after every single transaction.

I'm not really sure of the benefit, but the obvious problem is that people with lots and lots of commodity trades might notice balances drift after they upgrade.

I guess a minimum testcase would be some transaction that needs rounding, e.g.

```
2025/01/01 * Opening Balances
    Assets:Brokerage     $123.45
    Equity:Opening Balances
2025/02/01 * Investment
    Assets:Brokerage    0.005 FOO @ $124.95
    Assets:Brokerage    0.001 BAR @ $124.95
    Assets:Brokerage
```
With main you get:

```
$ ledger --cost -X USD bal
             $123.44  Assets:Brokerage
            $-123.45  Equity:Opening Balances
--------------------
              $-0.01
```

But every rounded transaction can add a penny, so if I duplicate it three times...

```
2025/01/01 * Opening Balances
    Assets:Brokerage     $123.45
    Equity:Opening Balances
2025/02/01 * Investment
    Assets:Brokerage    0.005 FOO @ $124.95
    Assets:Brokerage    0.001 BAR @ $124.95
    Assets:Brokerage
2025/02/01 * Investment
    Assets:Brokerage    0.005 FOO @ $124.95
    Assets:Brokerage    0.001 BAR @ $124.95
    Assets:Brokerage
2025/02/01 * Investment
    Assets:Brokerage    0.005 FOO @ $124.95
    Assets:Brokerage    0.001 BAR @ $124.95
    Assets:Brokerage
```

```
$ ledger --cost -X USD bal
             $123.42  Assets:Brokerage
            $-123.45  Equity:Opening Balances
--------------------
              $-0.03
```
In the current release, you get zero, because rounding only happens once.

This feels like a big change because it's no longer true that "the journal must always balance to zero"... which used to be the fundamental rule of ledger...

jwiegley:
## Summary

Fixes #3220 — a cost-basis report drifted by a penny per transaction that needed rounding.

### Exactly why it broke

`--cost` is an alias for `--basis`, which sets the amount expression to `rounded(cost)`. The fix for #1765 (`f253ef48c`) changed the `rounded()` value-expression in **one line**:

```cpp
-  return args.value().rounded();                          // display-only (original)
+  return args.value().rounded_to_commodity_precision();   // physical per-posting
```

`rounded_to_commodity_precision()` *physically* rounds each posting's stored value to commodity precision before aggregation. A transaction whose legs cancel exactly at full precision then no longer cancels once each leg is rounded:

```
0.005 FOO @ $124.95  ->  $0.62475  ->  $0.62
0.001 BAR @ $124.95  ->  $0.12495  ->  $0.12
(balancing leg)      -> -$0.74970  -> -$0.75
                                       -$0.01  per transaction
```

So `--cost -X USD bal` lost a penny per rounding transaction (and `-$0.03` over three), breaking the rule that a cost-basis report balances to zero.

### The fix (surgical: 2 source files)

**Revert `fn_rounded()` to display-only rounding** — its behavior before `f253ef48c`. It clears `keep_precision` so each posting prints at commodity precision while the stored value stays exact, so a balance accumulates full precision and rounds **once** at display time:

```
$ ledger --cost -X USD bal      # after the fix (any number of txns)
             $123.45  Assets:Brokerage
            $-123.45  Equity:Opening Balances
--------------------
                   0
```

Then **preserve the two pieces of work that had relied on `rounded()` being physical**, without disturbing the shared rounding primitives:

- **`--round` (#781)** wants per-posting physical rounding before aggregation. It now routes through a new `roundto_commodity()` expression that calls `rounded_to_commodity_precision()` directly, instead of through `rounded()`. So `bal -V --round` still differs from `bal -V` (30.24 vs 30.25), and `--cost -X USD --round` still exposes the residual (`-$0.03`) for anyone who wants broker-statement totals.

- **#3187** wants a cost basis in a commodity with no declared display precision (a `$` seen only in `@` prices) to keep its fractional digits (`$100.500`, not `$100`). `fn_rounded()` handles that one case inline, leaving such an amount untouched. **Keeping this guard local to the `--basis` path is essential** — I verified that applying it inside the shared rounding primitive (`amount_t::in_place_round`) instead also changed `-V`/`-X` valuation and even broke a transaction's balancing, because those legitimately round zero-precision commodities.

Net diff to `src/`: `report.cc` (+32/−3) and `report.h` (+13/−4). No changes to `amount.*`, `value.*`, or `balance.h`, so valuation and transaction finalization are untouched.

## Tests

- New `test/regress/3220.test` — single/triple-transaction drift balances to zero; `--round` opt-in still drifts.
- `test/regress/1765_b.test` — revealing a sub-cent imbalance is now the `--round` behavior; the default rounds the residual away once at the end.
- `test/regress/3187.test` — comment updated (behavior unchanged: still `$100.5`).
- `test/regress/coverage-value-round-string-error.test` — tracks `rounded()`'s restored display-only error text.
- Full suite: **4325/4325 pass** locally.

## Docs

`--round` in `doc/ledger3.texi` and `doc/ledger.1` now states that it rounds each posting before aggregation (broker-statement style), versus the default which keeps full precision and rounds once for display.

### DIFF
--- a/doc/ledger.1
+++ b/doc/ledger.1
@@ -1143,9 +1143,14 @@ When generating a ledger transaction from a CSV file using the
 .Ic convert
 command, add CSV, Imported, and UUID meta-data.
 .It Fl \-round
-Round all calculated values to the display precision of their commodity.
-This is the default behavior for display, but this option explicitly forces
-rounding in the expression pipeline, which is useful when
+Round each calculated value to the display precision of its commodity
+before postings are aggregated, so that a report total reflects the sum of
+the individually rounded amounts, the way a broker statement totals
+positions that have each already been rounded to the cent.
+Without this option Ledger keeps full precision while accumulating and
+rounds only once for display, so totals never drift.
+This option also forces rounding in the expression pipeline, which is useful
+when
 .Fl \-unround
 has been set in the init file and you want to override it for a specific
 invocation.
--- a/doc/ledger3.texi
+++ b/doc/ledger3.texi
@@ -8145,10 +8145,15 @@ When generating a ledger transaction from a CSV file using the
 @command{convert} command, add CSV, Imported, and UUID metadata.
 
 @item --round
-Round all calculated values to the display precision of their
-commodity.  This is the default behavior for display, but this option
-explicitly forces rounding in the expression pipeline, which is useful
-when @option{--unround} has been set in the init file and you want to
+Round each calculated value to the display precision of its commodity
+@emph{before} postings are aggregated, so that a report total reflects
+the sum of the individually rounded amounts -- the way a broker
+statement totals positions that have each already been rounded to the
+cent.  Without this option Ledger keeps full precision while
+accumulating and rounds only once for display, so totals never drift by
+a fraction of the smallest displayed unit.  @option{--round} also forces
+rounding in the expression pipeline, which is useful when
+@option{--unround} has been set in the init file and you want to
 override it for a specific invocation.
 
 @item --seed @var{INT}
--- a/src/report.cc
+++ b/src/report.cc
@@ -323,10 +323,10 @@ void report_t::normalize_options(const string& verb) {
         // price history entries and thus fall back to the lot cost.
         HANDLER(amount_).on("?normalize", "market(amount_expr, today, exchange)");
       }
-      HANDLER(amount_).on("?normalize_round", "rounded(amount_expr)");
+      HANDLER(amount_).on("?normalize_round", "roundto_commodity(amount_expr)");
     } else {
-      HANDLER(amount_).on("?normalize_round", "rounded(amount_expr)");
-      HANDLER(total_).on("?normalize_round", "rounded(total_expr)");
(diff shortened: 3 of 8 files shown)


jwiegley:
Revised to a much more surgical fix per review feedback. Rather than introducing a parallel rounding-function hierarchy, this now simply reverts `fn_rounded()` to its pre-#1765 display-only behavior (the one-line regression), routes `--round` through a dedicated `roundto_commodity()` expression to preserve #781, and handles the #3187 zero-precision case inline in `fn_rounded()`. Net source change is `report.cc`/`report.h` only (+45/−7); `amount.*`/`value.*`/`balance.h` are untouched, so valuation and transaction balancing carry no risk. Full suite still 4325/4325.

PR #3234, opened 2026-06-05
Fix #3233: preserve zero amount commodity in print

### LINKED ISSUE #3233: `print` strips the commodity from zero amounts (`$0.00` becomes bare `0`)
### Expected behavior

A posting written as `$0.00` in the journal should be printed with its
commodity intact:

```
    Expenses:Interest                          $0.00
```

### Actual behavior

`ledger print` re-emits the posting as a bare, commodity-less `0`:

```
    Expenses:Interest                              0
```

This makes the output of `print` invalid as input for stricter downstream
parsers that require every amount to carry a commodity, and means
`print` is not round-trip stable: the printed journal is not equivalent
to the source journal.

### Steps to reproduce

`repro.ledger`:

```
2021/12/23 * minimal zero-commodity repro
    Expenses:Interest        $0.00
    Expenses:Principal     $100.00
    Assets:Checking       $-100.00
```

```
$ ledger print -f repro.ledger
2021/12/23 * minimal zero-commodity repro
    Expenses:Interest                              0
    Expenses:Principal                       $100.00
    Assets:Checking                         $-100.00
```

Writing the amount as `$0` instead of `$0.00` makes no difference.

Flags tried:

- `--generated`: no effect, still bare `0`.
- `--raw`: preserves `$0.00`, but only because it echoes the source text
  verbatim and skips amount normalization entirely, so it is not a
  substitute for normal `print` output.

Real-world context: bank loan-payment exports routinely produce
zero-interest postings (e.g. an initial loan disbursement, or a
principal-only payment), so `$0.00` postings occur naturally in
imported data.

### Precedent

hledger had this same bug and fixed it:
simonmichael/hledger#475 ("print should not strip commodity from zero
amounts when it is specified in the journal"). Its resolution treats
the journal's stated commodity as authoritative even when the quantity
is zero.

Possibly related: #1652, where the `csv` command drops zero-amount
postings entirely — another case of zero amounts being special-cased in
output formats.

### Environment

```
Ledger 3.3.2-20230330, the command-line accounting tool
with support for gpg encrypted journals and without Python support
```

macOS (Darwin 24.6.0), x86_64.

nkgotcode:
Fixes #3233.

`ledger print` currently renders structured posting amounts through
`value_t::print()`. For zero amounts, that path simplifies the value to a bare
`0`, which drops the commodity text from inputs such as `$0.00`.

This preserves the commodity when `print` emits posting amounts by passing the
existing zero-display flag from `print.cc` and honoring it in `value_t::print()`.
The fallback simplification to `0` remains in place for callers that do not ask
to show zero amounts.

Validation:

```
LD_LIBRARY_PATH=/tmp/oss-1013-cmake/extracted/usr/lib/x86_64-linux-gnu /tmp/oss-1013-cmake/extracted/usr/bin/cmake -S . -B build-oss1013 -DCMAKE_BUILD_TYPE=Debug -DUSE_PYTHON=OFF -DUSE_GPGME=OFF -DCMAKE_PREFIX_PATH=/tmp/oss-1013-deps/extracted/usr -DBoost_DIR=/tmp/oss-1013-deps/extracted/usr/lib/x86_64-linux-gnu/cmake/Boost-1.83.0
env LD_LIBRARY_PATH=/tmp/oss-1013-cmake/extracted/usr/lib/x86_64-linux-gnu:/tmp/oss-1013-deps/extracted/usr/lib/x86_64-linux-gnu make -C build-oss1013 ledger -j2
python3 test/RegressTests.py --ledger /tmp/ledger-oss1013-wrapper --sourcepath . test/regress/3233.test
git diff --check
```

### DIFF
--- a/src/print.cc
+++ b/src/print.cc
@@ -387,11 +387,16 @@ void print_xact(report_t& report, std::ostream& out, xact_t& xact) {
         // When printing user-annotated amounts, preserve {{total}} lot-price
         // notation so that the output round-trips faithfully (issue #1033).
         bool preserve_total = post->has_flags(POST_AMOUNT_USER_ANNOTATED);
-        value_t(post->amount)
-            .print(amt_str, boost::numeric_cast<int>(amount_width), -1,
-                   AMOUNT_PRINT_RIGHT_JUSTIFY |
-                       (suppress_computed ? AMOUNT_PRINT_NO_COMPUTED_ANNOTATIONS : 0) |
-                       (preserve_total ? AMOUNT_PRINT_PRESERVE_TOTAL_COST : 0));
+        uint_least8_t flags = AMOUNT_PRINT_RIGHT_JUSTIFY |
+                              (suppress_computed ? AMOUNT_PRINT_NO_COMPUTED_ANNOTATIONS : 0) |
+                              (preserve_total ? AMOUNT_PRINT_PRESERVE_TOTAL_COST : 0);
+        if (post->amount.is_zero()) {
+          std::ostringstream buf;
+          post->amount.print(buf, flags);
+          justify(amt_str, buf.str(), boost::numeric_cast<int>(amount_width), true);
+        } else {
+          value_t(post->amount).print(amt_str, boost::numeric_cast<int>(amount_width), -1, flags);
+        }
         amt = amt_str.str();
       }
 
--- a/test/regress/3233.test
+++ b/test/regress/3233.test
@@ -1,0 +1,11 @@
+2021/12/23 * minimal zero-commodity repro
+    Expenses:Interest        $0.00
+    Expenses:Principal     $100.00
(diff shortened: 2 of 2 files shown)


nkgotcode:
Pushed 5605375 to narrow the zero-commodity rendering to transaction `print` output only. The failed Nix run was from broader zero rendering in register/balance fixtures; locally verified with `python3 test/RegressTests.py --ledger build-oss1013/ledger --sourcepath .` for `test/regress/3233.test`, `test/baseline/opt-empty.test`, `test/baseline/opt-empty_collapse.test`, `test/regress/1137.test`, `test/regress/6DAB9FE3.test`, `test/regress/F06D5554.test`, and `test/regress/coverage-filters-emptyperiod.test`: all passed.

jwiegley (approved):

PR #3250, opened 2026-07-16
Add support for External Directives

mariolopjr:
TBD


mariolopjr:
Decided a different path, instead of allowing an arbitrary external directive, I'll add a `custom` directive like beancount, which fava uses for settings: https://beancount.io/docs/Basics/fava-options

mariolopjr:
Actually, `comment end` could work for me as well, sorry for the noise

